PR #6, opened 2023-08-08
Initial Replication Mock.

szmyd:


JacksonYao287:
截屏2023-08-09 13 22 27

the build error seems happens because of lz4

src/mocks/repl_set.hpp
```diff
@@ -0,0 +1,61 @@
...
+    /// @brief Send the final responce to the rpc client.
+    /// @param outgoing_buf - response buf to client
+    /// @param rpc_data - context provided by the rpc server
+    virtual void send_data_service_response(sisl::io_blob_list_t const& outgoing_buf,
+                                            boost::intrusive_ptr< sisl::GenericRpcData >& rpc_data) = 0;
+
+    /// @brief Append an application/user specific message to the journal.
+    ///
+    /// @param buffer - Opaque buffer to be interpreted by the user
+    virtual void append_entry(nuraft::buffer const& b) = 0;
```
zichanglai, line 55: what is the difference between this interface and write() and who will use this interface.

lucasgonze:
Could you remove me from the repo owners list? I am no longer working for
eBay.

I would do this myself but can't get to the repo without signing in to the
eBay organization, which I can't do.

szmyd:
> Could you remove me from the repo owners list? I am no longer working for eBay. I would do this myself but can't get to the repo without signing in to the eBay organization, which I can't do.
Lucas, I did this last week or two; you may still be "watching" the repo which I can not control. Let me know if you continue to experience otherwise.

src/mocks/repl_set.hpp
szmyd: This is a generic interface to send opaque entries to all members. This can be used for shard creation without exposing the knowledge of shards themselves to HomeReplication. `write` is for data transfer outside of RAFT, and is followed up with `transfer_pba_ownership` to commit tie the operation back into RAFT consensus. I think we are okay with HomeReplication knowing about this level of detail, but want to make some things opaque.

zichanglai: got it.  so append_entry() is the interface for nuraft log replication whether the log content is pba or create shard request etc,  inside write() it will invoke append_entry() too after data is transfered among members via data channel.

PR #28, opened 2023-08-25
Convert API to use Folly::Expected.

szmyd:


src/mocks/mock_blob_manager.cpp
```diff
@@ -1,11 +1,13 @@
...
+#define MAKE_BLOB_PROMISE(T) auto [p, f] = folly::makePromiseContract< folly::Expected< T, BlobError > >()
+
 namespace homeobject {
 using namespace std::chrono_literals;
 constexpr auto disk_latency = 15ms;
 
-folly::SemiFuture< std::variant< blob_id, BlobError > > MockHomeObject::put(shard_id shard, Blob&& blob) {
-    auto [p, f] = folly::makePromiseContract< std::variant< blob_id, BlobError > >();
+BlobManager::AsyncResult< blob_id > MockHomeObject::put(shard_id shard, Blob&& blob) {
+    MAKE_BLOB_PROMISE(blob_id);
```
hkadayam, line 10: Just curious, if we can avoid this macro with following, since we already have the alias

```folly::makePromiseContract< Result< blob_id > >()```

szmyd: We have to qualify Result. I've posted that change.

PR #30, opened 2023-08-31
SDSTOR-11619 Implement HeapChunkSelector

JacksonYao287:


JacksonYao287:
implemented in homestore

src/lib/homestore/CMakeLists.txt
```diff
@@ -6,6 +6,7 @@ target_sources("${PROJECT_NAME}_homestore" PRIVATE
 		homeobject.cpp
 		blob_manager.cpp
 		shard_manager.cpp
+    heap_chunk_selector.cpp
```
szmyd, line 9: formatting

src/lib/homestore/heap_chunk_selector.h
```diff
@@ -0,0 +1,46 @@
...
+#include <vector>
+#include <mutex>
+#include <utility>
+#include <memory>
+
+#include <folly/concurrency/ConcurrentHashMap.h>
+
+namespace homeobject {
+
+using csharedChunk = homestore::cshared<homestore::Chunk>;
```
szmyd, line 17: I don't understand this `cshared` naming convention? I see it's being used in HomeStore but why `c`? Seems names like `chunk_ptr` would be better? Then we don't have to change the code everywhere if we switch from shared to unique or raw...etc

src/lib/homestore/heap_chunk_selector.h
```diff
@@ -0,0 +1,46 @@
...
+
+namespace homeobject {
+
+using csharedChunk = homestore::cshared<homestore::Chunk>;  
+
+class HeapChunkSelector : public homestore::ChunkSelector {
+public:
+    HeapChunkSelector() = default;
+    HeapChunkSelector(const HeapChunkSelector&) = delete;
+    HeapChunkSelector(HeapChunkSelector&&) noexcept = delete;
```
szmyd, line 23: Why do we explicitly delete this? We have a default CTor and DTor; let the compiler decide if we can support this by default based on member fields?

src/lib/homestore/heap_chunk_selector.h
```diff
@@ -0,0 +1,46 @@
...
+    class VChunkComparator {
+    public:
+        bool operator()(VChunk& lhs, VChunk& rhs) {
+            return lhs.available_blks() < rhs.available_blks();
+        }
+    };
+
+    using VChunkHeap = std::priority_queue<VChunk, std::vector<VChunk>, VChunkComparator>;
+
+    void add_chunk(csharedChunk&) override;
```
szmyd, line 39: I realize this is from HomeStore's interface but should be either:
`void add_chunk(csharedChunk const&) override; // < Hari's Preference`
or
`void add_chunk(csharedChunk) override; // < Brian's Preference`
But we intend to take ownership of the chunk here yeah, not mutate the caller's own stack.

src/lib/homestore/heap_chunk_selector.h
```diff
@@ -0,0 +1,46 @@
...
+    void foreach_chunks(std::function< void(csharedChunk&) >&& cb) override;
```
szmyd, line 40: This too; are we intending the closure to mutate the Chunk Pointer itself? Or just the Chunk....

src/lib/homestore/heap_chunk_selector.cpp
```diff
@@ -0,0 +1,38 @@
+#include "heap_chunk_selector.h"
+
+namespace homeobject {
+
+void HeapChunkSelector::add_chunk(csharedChunk& chunk) {
+    uint32_t pdevId = VChunk(chunk).get_pdev_id();
+    auto it = _pdev_heap_map.find(pdevId);
+    if (it == _pdev_heap_map.end()) it = _pdev_heap_map.try_emplace(pdevId, std::make_shared<std::pair<std::mutex, VChunkHeap>>()).first;
```
szmyd, line 8: Can we use the pattern:
```
if (auto [it, _] = _pdev_heap_map.try_emplace(...); _pdev_heap_map.end() != it) {
    auto l = std::scoped_lock(it->second->first);
    it->second->second.emplace(chunk);
} else RELEASE_ASSERT(false, "Something Bad!");
```
then you don't have to call `find` as well.

src/lib/homestore/heap_chunk_selector.cpp
```diff
@@ -0,0 +1,38 @@
...
+    std::unique_lock<std::mutex> l(it->second->first);
+    it->second->second.emplace(chunk);
+}
+
+csharedChunk HeapChunkSelector::select_chunk(homestore::blk_count_t count, const homestore::blk_alloc_hints& hint) {
+    auto it = _pdev_heap_map.find(hint.dev_id_hint);
+    if (it == _pdev_heap_map.end()) return nullptr;
+    std::unique_lock<std::mutex> l(it->second->first);
```
szmyd, line 16: `scoped_lock` or `lock_guard` if not intending to move it.

szmyd (dismissed):
Some initial comments/requests.

src/lib/homestore/CMakeLists.txt
yamingk: Does it make sense to apply some unified format plugin and apply to all? So that when another one is making changes to same file, it won't yield code changes for all the lines.

src/lib/homestore/heap_chunk_selector.cpp
JacksonYao287: sure, will change

src/lib/homestore/CMakeLists.txt
JacksonYao287: sorry for this. it is formatted in my vscode before pushing the PR, i am not sure why it is not formatted here. will check it again

src/lib/homestore/heap_chunk_selector.cpp
```diff
@@ -0,0 +1,38 @@
...
+    if(it->second->second.empty()) return nullptr;
+    auto& vchunk = it->second->second.top();
+    it->second->second.pop();
+    return vchunk.get_internal_chunk();
+}
+
+void HeapChunkSelector::foreach_chunks(std::function< void(csharedChunk&) >&& cb) {
+    VChunkHeap tempHeap;
+    for(auto& p : _pdev_heap_map) {
+        std::unique_lock<std::mutex> l(p.second->first);
```
zichanglai, line 26: I think it will be more clear to move L24 to here to use a local variable.

src/lib/homestore/heap_chunk_selector.cpp
JacksonYao287: good, this pattern seems better, thanks for the suggestion

src/lib/homestore/heap_chunk_selector.h
JacksonYao287: actually, i have the similar idea. but , this is from homestore, maybe we can create a pr in homestore to make this change

src/lib/homestore/heap_chunk_selector.h
JacksonYao287: yes , we do not want to mutate chunk in the caller`s stack , will change this in the homestore pr

src/lib/homestore/heap_chunk_selector.cpp
JacksonYao287: if that , tempHeap will be declared multiple times in the loop. the current will reuse the only tempHeap , which will be constructed once

zichanglai: ok, make sense to me, my origin point is if we use local variable for per loop, if something wrong happens in one loop, the next loop is not affected.

src/lib/homestore/heap_chunk_selector.cpp
```diff
@@ -0,0 +1,38 @@
...
+        auto& heap = p.second->second;
+        while(!heap.empty()) {
+            auto& vchunk = heap.top();
```
zichanglai, line 29: after heap.pop() call, the vchunk may become  a stale object if use the reference?

src/lib/homestore/heap_chunk_selector.h
szmyd: So; I misunderstood the `c` effecting the value type, not being simply an alias for `const shared<>`; I'm okay with the signature, but the alias i'd prefer to just see:
```
void add_chunk(const sharedChunk&) override;
```
but it's a nitpick now that i understand this alias.

src/lib/homestore/heap_chunk_selector.cpp
JacksonYao287: i do get your concern here. 

actually ,`top()` will return const_reference, so the  `auto&` here will be deduced to `const VChunk&`, it is a lvalue references to const 

well, [in this page](https://en.cppreference.com/w/cpp/language/reference) ,  it says "lvalue references to const can extend the lifetimes of temporary objects too, but they are not modifiable through them". so the object will not be stale.

i write a program to testify this rule:

src/lib/homestore/CMakeLists.txt
xiaoxichen: It seems this file use TAB instead of space.  I think we should expand TAB with some space (ts:4 or ts:2, based on teams preference).  It seems like @JacksonYao287 's IDE configured with ts=2

JacksonYao287: i tried several times, it is looks good in my local IDE, not sure why it is like this in github show page

JacksonYao287:
compiling error is caused by `operator==` and `get_chunk_id()` in VChunk.h , which will be added to homestore soon

src/lib/homestore/heap_chunk_selector.cpp
```diff
@@ -0,0 +1,79 @@
...
+    auto it = m_pdev_heap_map.find(pdevID);
+    if (it == m_pdev_heap_map.end()) {
+        LOGINFO("No pdev found for pdev {}", pdevID);
+        return nullptr;
+    }
+    
+    it->second->first.lock();
+    if(it->second->second.empty()) {
+        LOGINFO("No more available chunks found for pdev {}", pdevID);
+        it->second->first.unlock();
```
JacksonYao287, line 52: thanks @xiaoxichen  for pointing out this

JacksonYao287:
compiling will fail until [this PR](https://github.com/eBay/HomeStore/pull/170) in merged in homestore

src/lib/homestore/heap_chunk_selector.cpp
```diff
@@ -0,0 +1,110 @@
...
+#include <execution>
+#include <algorithm>
+#include <sisl/logging/logging.h>
+
+namespace homeobject {
+
+//https://github.com/facebook/folly/blob/61c11d77eb9a8bdc60f673017fccfbe900125cb6/folly/AtomicHashMap.h#L42
+//AtomichashMap has a max size limit of ~18x initial size.
+//we support max of 64k chunks for now, so 64*1024/18 + 1 = 3641, we use 4000 for sufficiently initial size 
+#define DEFAULT_ATOMICHASHMAP_INITIAL_SIZE 4000
```
xiaoxichen, line 12: we know for every HS we will go to 64K ,maybe just using 64K?

xiaoxichen:
```
static constexpr HeapChunkSelector::chunk_atomicmap_init_size() {
     return 2<<16;}
static constexpr HeapChunkSelector::pdev_atomicmap_init_size() {
     return 2<<4;}
```

src/lib/homestore/heap_chunk_selector.cpp
```diff
@@ -0,0 +1,110 @@
...
+m_chunks(ChunkMap(chunk_initial_num)) {}
+
+
+//add_chunk may be called in Homeobject(when sealing a shard) or in Homestore(when adding chunks to vdev) concurently
+void HeapChunkSelector::add_chunk(csharedChunk& chunk) {
+    VChunk vchunk(chunk);
+    auto&& pdevID = vchunk.get_pdev_id();
+    auto&& [it, _] = m_pdev_heap_map.emplace(pdevID, std::make_shared<std::pair<std::mutex, VChunkHeap>>());
+    const auto& available_blk_num = vchunk.available_blks();
+    auto&& [ignore, ok] = m_pdev_avalable_blk_map.emplace(pdevID, available_blk_num);
```
xiaoxichen, line 32: maybe s/ok/inserted for better readability

xiaoxichen (commented):
lgtm in general

src/lib/homestore/CMakeLists.txt
szmyd: try `retab` after setting `et` and `ts=4`.

src/lib/homestore/heap_chunk_selector.cpp
szmyd: I like `auto [it, happened]` since it reads like english.

JacksonYao287: change to happend to keep the same style with current code

src/lib/homestore/heap_chunk_selector.cpp
JacksonYao287: should be 1<<16(64K) and 1 <<4 ?

src/lib/homestore/heap_chunk_selector.cpp
```diff
@@ -0,0 +1,116 @@
...
+HeapChunkSelector::HeapChunkSelector() : 
+m_pdev_heap_map(PdevHeapMap(pdev_atomicmap_init_size)), 
+m_pdev_avalable_blk_map(pdev_atomicmap_init_size), 
+m_chunks(chunk_atomicmap_init_size) {}
+
+HeapChunkSelector::HeapChunkSelector(const uint32_t& pdev_heap_map_initial_size, 
+const uint32_t& pdev_avalable_blk_map_initial_size, const uint32_t& chunk_initial_num) : 
+m_pdev_heap_map(PdevHeapMap(pdev_heap_map_initial_size)), 
+m_pdev_avalable_blk_map(PdevAvalableBlkMap(pdev_avalable_blk_map_initial_size)),
+m_chunks(ChunkMap(chunk_initial_num)) {}
```
szmyd, line 21: Looks like this code needs to be run through `apply_clang_format.sh`?

szmyd:
Need code coverage: https://app.codecov.io/gh/eBay/HomeObject/tree/HeapChunkSelector/src%2Flib%2Fhomestore

Otherwise looks good!

src/lib/homestore/heap_chunk_selector.cpp
JacksonYao287: sure , will run apply_clang_format.sh and add UT. thanks

src/lib/homestore/heap_chunk_selector.cpp
```diff
@@ -0,0 +1,116 @@
...
+
+
+//add_chunk may be called in Homeobject(when sealing a shard) or in Homestore(when adding chunks to vdev) concurently
+void HeapChunkSelector::add_chunk(csharedChunk& chunk) {
```
xiaoxichen, line 25: IIRC we need a add::chunk(uint64 chunkID) overload for HO?

JacksonYao287: we already has a `release_chunk` for this purpose.
https://github.com/eBay/HomeObject/blob/27e9f1de6ae598da51b55f71ff599575db5ef893/src/lib/homestore/heap_chunk_selector.cpp#L108

xiaoxichen: sorry my bad. Yes
LGTM \

xiaoxichen (dismissed):

src/lib/homestore/heap_chunk_selector.cpp
```diff
@@ -0,0 +1,116 @@
...
+    VChunk vchunk(chunk);
+    auto&& pdevID = vchunk.get_pdev_id();
+    auto&& [it, _] = m_pdev_heap_map.emplace(pdevID, std::make_shared<std::pair<std::mutex, VChunkHeap>>());
```
hkadayam, line 28: What style of casing are we using, camelCase or underscore_case. It seems like code has both these cases in same method, would prefer consistency one way or other. Let us pick one as a team and follow.

src/lib/homestore/heap_chunk_selector.h
```diff
@@ -0,0 +1,65 @@
...
+
+    using PdevHeapMap = folly::AtomicHashMap< uint32_t, std::shared_ptr< std::pair< std::mutex, VChunkHeap > > >;
+    using PdevAvalableBlkMap = folly::AtomicHashMap< uint32_t, std::atomic_uint32_t >;
+    using ChunkMap = folly::AtomicHashMap< uint32_t, csharedChunk >;
+
+private:
+    //this holds all the unselected chunks for each pdev.
+    PdevHeapMap m_pdev_heap_map;
+
+    //for now, uint32_t is enough for the sum of all the available blocks of a pdev.
```
xiaoxichen, line 58: better to do a calculation here, I am not sure if uint32 is enough?

4B (2^32) * 4K = 16TB that is the limit of single pdev? if that is the case lets use uint64 as our current SKU is already running 16TB drive.

src/lib/homestore/heap_chunk_selector.cpp
szmyd: I think we have yet to discuss this. :)

My preference (let it be known) is for Classes and Structs to use CamelCase; but internal members to use snake_case. Parameters to functions should also prefer snake_case.

src/lib/homestore/heap_chunk_selector.h
szmyd: why not make it `size_t` and let the architecture decide? If there is a disk larger than the available addressable space on a platform it's not something we should support anyways?

src/lib/homestore/heap_chunk_selector.h
```diff
@@ -0,0 +1,65 @@
...
+    };
+
+    using VChunkHeap = std::priority_queue<VChunk, std::vector<VChunk>, VChunkComparator>;
+
+    void add_chunk(csharedChunk&) override;
+    void foreach_chunks(std::function< void(csharedChunk&) >&& cb) override;
+    csharedChunk select_chunk([[maybe_unused]]homestore::blk_count_t nblks, const homestore::blk_alloc_hints& hints);
+
+    // this function is used to return a chunk back to ChunkSelector when sealing a shard, and will only be used by Homeobject.
+    void release_chunk(const uint32_t&);
```
hkadayam, line 42: Just curious do we need to const pass-by-reference for uint32_t or easily be `void release_chunk(uint32_t)`. Isn't pass-by-value for this is more efficient and also simpler?

src/lib/homestore/heap_chunk_selector.h
xiaoxichen: yes that is right

src/lib/homestore/heap_chunk_selector.h
szmyd: Yes; reference are really `pass by pointer` anyways; which on a 64-bit platform is a "technically" more costly than a uint32_t.

src/lib/homestore/heap_chunk_selector.cpp
```diff
@@ -0,0 +1,116 @@
...
+        auto& heap = it->second->second;
+        std::lock_guard<std::mutex> l(heapLock);
+        if(heap.empty()) return VChunk(nullptr);
+        const auto& vchunk = heap.top();
+        heap.pop();
+        return vchunk;
+    }();
+
+    if(vchunk.get_internal_chunk()){
+        auto&& avalableBlkCounter = m_pdev_avalable_blk_map.find(pdevID)->second;
```
hkadayam, line 91: type m_pdev_avalable_blk_map -> m_pdev_available_blk_map

szmyd: I also don't get this:
```
auto&& ...
```
What does this mean? I think the result is...a copy? It's not a reference, so it's moving the temporary? Can we just make it `auto`?

src/lib/homestore/heap_chunk_selector.h
```diff
@@ -0,0 +1,65 @@
...
+    // we suppose our max pdev number is 64(64 physical disks)
+    static constexpr uint32_t pdev_atomicmap_init_size = 1<<6;
+
+    using PdevHeapMap = folly::AtomicHashMap< uint32_t, std::shared_ptr< std::pair< std::mutex, VChunkHeap > > >;
+    using PdevAvalableBlkMap = folly::AtomicHashMap< uint32_t, std::atomic_uint32_t >;
+    using ChunkMap = folly::AtomicHashMap< uint32_t, csharedChunk >;
+
+private:
+    //this holds all the unselected chunks for each pdev.
+    PdevHeapMap m_pdev_heap_map;
```
hkadayam, line 56: I am unable to see the reason we need AtomicHashMap of pdev_id -> Heap. The number of entries in this map will be constant and don't change run_time. So isn't much better to put one structure instead of 2 hash maps, something like
```
struct PerDevHeap {
    std::mutex mtx;
    VChunkHeap m_heap;
     uint32_t available_blk_count;
};

and
HeapChunkSelector has
```
std::map< uint32_t, PerDevHeap > m_per_dev_heap;
```
We need to initialize this m_per_dev_heap during add_chunk(). Given that HomeObject doesn't need dynamic chunk requirement, it will be single threaded and we emplace the PerDevHeap inside the map.

File: src/lib/homestore/heap_chunk_selector.h (hunk 16 above)
Comment (JacksonYao287) [reply src/lib/homestore/heap_chunk_selector.h:58]: correct

File: src/lib/homestore/heap_chunk_selector.h (hunk 17 above)
Comment (JacksonYao287) [reply src/lib/homestore/heap_chunk_selector.h:42]: yes that is right

File: src/lib/homestore/heap_chunk_selector.cpp (hunk 18 above)
Comment (JacksonYao287) [reply src/lib/homestore/heap_chunk_selector.cpp:91]: actually, `auto&&` here is universal reference, and `->second` is a lvalue， so here it will be deduced to a lvalue reference。to make it more straightforward, we can change it to `auto&`

Comment (szmyd) [reply src/lib/homestore/heap_chunk_selector.cpp:91]: :+1:

**Comment by JacksonYao287** [comment]
@szmyd i do not link `libhomeobject_homestore.a` when writing UT for HeapChunkSelector, but the UT covers all the functions of HeapChunkselector, not sure if this will be captured by Codecov and update the code coverage.

let me explaint this in more detail. if we want to test the correctness of HeapChunkSelector, we need to check whether the returned chunk is the one we expected. but we know nothing detail of chunk and it is only known by homstore, we just only have a raw declaration in homeobject, and we can not derive or mock it.

so i write a chunk implementation in the test,  and if we link `libhomeobject_homestore.a` which will link `libhomestore.a`, there will be a conflict that two defination of homestore::chunk exist. this is why i do not link `libhomeobject_homestore.a`.

File: src/lib/homestore/heap_chunk_selector.cpp [hunk 20]
```diff
@@ -0,0 +1,127 @@
...
+    auto& heap = m_per_dev_heap[pdevID]->m_heap;
+    std::vector< VChunk > temp;
+    std::lock_guard< std::mutex > l(heapLock);
+    for (; !heap.empty();) {
+        VChunk vchunk = heap.top();
+        heap.pop();
+        if (vchunk.get_chunk_id() == chunkID) break;
+        temp.emplace_back(vchunk);
+    }
+    for (auto& vchunk : temp) {
```
zichanglai, line 123: it is also required to reduce the pdev available blk num, right?

JacksonYao287: good catch , thanks , will fix it

src/lib/homestore/heap_chunk_selector.h
```diff
@@ -0,0 +1,54 @@
...
+
+    // homestore will initialize HeapChunkSelector by adding all the chunks. but some of them are already
+    // selected by open shards. so after homeobject restarts, we need to mark all these chunks as selected
+    void mark_chunk_selected(const uint16_t);
+
+private:
+    std::unordered_map< uint32_t, std::shared_ptr< PerDevHeap > > m_per_dev_heap;
+
+    // hold all the chunks , selected or not
+    std::unordered_map< uint16_t, csharedChunk > m_chunks;
```
yamingk, line 52: uint16_t -> chunk_numt_t

src/lib/homestore/heap_chunk_selector.h
```diff
@@ -0,0 +1,54 @@
...
+    // this function is used to return a chunk back to ChunkSelector when sealing a shard, and will only be used by
+    // Homeobject.
+    void release_chunk(const uint16_t);
+
+    // homestore will initialize HeapChunkSelector by adding all the chunks. but some of them are already
+    // selected by open shards. so after homeobject restarts, we need to mark all these chunks as selected
+    void mark_chunk_selected(const uint16_t);
+
+private:
+    std::unordered_map< uint32_t, std::shared_ptr< PerDevHeap > > m_per_dev_heap;
```
yamingk, line 49: It would be great if we can have a pdev_id_t (also needs to change in HomeStore), so we don't need to worry about future expansion or downgrade to another intx_t. It doesn't need to happen in this PR.

src/lib/homestore/tests/test_heap_chunk_selector.cpp
```diff
@@ -0,0 +1,148 @@
...
+    Chunk(uint32_t pdev_id, uint16_t chunk_id, uint32_t available_blks) {
+        m_available_blks = available_blks;
+        m_pdev_id = pdev_id;
+        m_chunk_id = chunk_id;
+    }
+
+private:
+    uint32_t m_available_blks;
+    uint32_t m_pdev_id;
+    uint16_t m_chunk_id;
```
yamingk, line 39: chunk_num_t

src/lib/homestore/heap_chunk_selector.cpp
```diff
@@ -0,0 +1,131 @@
...
+    m_chunks.emplace(vchunk.get_chunk_id(), chunk);
+
+    auto& heapLock = it->second->mtx;
+    auto& heap = it->second->m_heap;
+    std::lock_guard< std::mutex > l(heapLock);
+    heap.emplace(chunk);
+}
+
+// select_chunk will only be called in homestore when creating a shard.
+csharedChunk HeapChunkSelector::select_chunk(homestore::blk_count_t count, const homestore::blk_alloc_hints& hint) {
```
yamingk, line 36: `count` is not used anywhere in this function. If this count is meant for seleted_chunk should at least have more than this number of blks, you probably need to compare it before return, or is it for any other puspose?

src/lib/homestore/heap_chunk_selector.cpp
```diff
@@ -0,0 +1,131 @@
...
+    auto& chunkIdHint = hint.chunk_id_hint;
```
yamingk, line 37: Do you want to use consistent naming conventions? either myVarName or my_var_name? I see both of these in this, can we use consistent one?

yamingk, line 37: chunk_id_hint or chunkIdHint is not used anywhere in this function, what is the reason of checking it here? 

There is no use case consumer will call select_chunk with a chunk_id_hint, right?

The only use case I can see is remove `mark_chunk_selected`, replace by calling `select_chunk` too with a provided `chunk_id_hint`; But I also don't see why `mark_chunk_selected` is needed during recovery boot. 

Ideally HeapChunkSelector should wait to build heap until ShardManager completes its recovery, so that at that point, ShardManager knows which are the chunks that are with open_shards and those chunks should not participate heap building but only be added to m_chunks. It is okay to do this with another PR.

src/lib/homestore/heap_chunk_selector.cpp
```diff
@@ -0,0 +1,131 @@
...
+    if (!pdevIdHint.has_value()) {
+        // this is the first shard of this pg, select a pdev with the most available blocks for it
+        auto&& it =
+            std::max_element(m_per_dev_heap.begin(), m_per_dev_heap.end(),
+                             [](const std::pair< const uint32_t, std::shared_ptr< PerDevHeap > >& lhs,
+                                const std::pair< const uint32_t, std::shared_ptr< PerDevHeap > >& rhs) {
+                                 return lhs.second->available_blk_count.load() < rhs.second->available_blk_count.load();
+                             });
+        if (it == m_per_dev_heap.end()) {
+            LOGINFO("No pdev found for new pg");
```
yamingk, line 57: For all error cases in this function, please LOGERROR

src/lib/homestore/heap_chunk_selector.cpp
```diff
@@ -0,0 +1,131 @@
...
+        pdevID = pdevIdHint.value();
+    }
+
+    auto it = m_per_dev_heap.find(pdevID);
+    if (it == m_per_dev_heap.end()) {
+        LOGINFO("No pdev found for pdev {}", pdevID);
+        return nullptr;
+    }
+
+    const auto& vchunk = [it = std::move(it), pdevID]() {
```
yamingk, line 71: Why do you want to pdevID into lambda if it is not used anywhere inside lambda?

src/lib/homestore/heap_chunk_selector.cpp
xiaoxichen: I think Yaojie probabaly right.  The chunk information recovery is happending in HS layer, `PhysicalDev::load_chunks` will call `chunk_found_cb` and I believe the chunk should be added into selector here (i.e in the lambda of  DeviceManager::load_vdevs() )

Later when HO recover its shards, for every open shards it recovered , it marks the chunk as selected from selector.

Shard/Chunk are from different layer so if we want to wait ShardManager recover completes, we need to let HO get all chunks belongs to a vdev, which,  IMO is breaking the isolation

yamingk: ShardManager only need to inform HeapChunkSelector with  a list of "chunk ids" (which has open shards) and at that point, HeapChunkSelector can start building its heap by excluding those chunks with open shards on. 
ShardManager is already recovering from metablks and it doesn't need to know anything about chunks/vdev, just chunk ids is enough, right?

xiaoxichen: ShardMgr has a list of open shards for sure but doesnt necessary has a list of "free shard" right?

src/lib/homestore/heap_chunk_selector.h
JacksonYao287: sure , let do it in another pr

src/lib/homestore/tests/test_heap_chunk_selector.cpp
JacksonYao287: ok

src/lib/homestore/heap_chunk_selector.cpp
JacksonYao287: sure

src/lib/homestore/heap_chunk_selector.cpp
JacksonYao287: good point!

src/lib/homestore/heap_chunk_selector.cpp
JacksonYao287: this is just a function from homestore, and i just derived from it. I have marked it as `[[maybe_unused]]` in the header file. if we want to change this , we need to change the header file in homestore.

src/lib/homestore/heap_chunk_selector.cpp
JacksonYao287: when homeobject restarts, it should first initialize homestore, because it need to load some metadata (for example, shard >chunk mapping) from metaservice. and then it can intialize shardmanager, and now shardmanager is able to know which shards are open. when intializing homestore, all the chunks will be added to ChunkSelector when initializing vdev.  
so we can not do this `m_shardMgr_inst->m_chunk_selector->start_recovery(open_chunk_ids);`. what we can do is only add all the chunks , and then get what shards are open ,and then mark them selected.

JacksonYao287: i get your point @yamingk , i thinks the only key difference between our ideas is when should HeapChunkSelector build the per Pdev heap.

My implementation is buiding per pdev heap when adding chunk.
your idea is not building per pdev heap when adding chunk, but building the heaps after shardManager is recovered, so the HeapChunkSelector can build the heaps with the opened-shard-chunk information told by shardManager to exclude them from the heap.

logically, they can both achieve the goal. but `add_chunk` is also called by `release_chunk`(seal shard), so if using your idea, we need to add additional logic in `add_Chunk` to identify whether this add is a `release_chunk` call or just a `normal add chunk` when initializing, which may let the logic in this function a little confusing and complex.

so i prefer my current implementation

JacksonYao287:
error mesage is 

`/home/runner/work/HomeObject/HomeObject/src/lib/homestore/homeobject.cpp:46:99: error: ‘homestore::hs_input_params’ has no non-static data member named ‘services’
   46 |         hs_input_params{.devices = device_info, .app_mem_size = app_mem_size, .services = services},`

seems the build error is not caused by my code, i build it successfully in my local environment.

zichanglai:
> error mesage is
> 
> `/home/runner/work/HomeObject/HomeObject/src/lib/homestore/homeobject.cpp:46:99: error: ‘homestore::hs_input_params’ has no non-static data member named ‘services’ 46 | hs_input_params{.devices = device_info, .app_mem_size = app_mem_size, .services = services},`
> 
> seems the build error is not caused by my code, i build it successfully in my local environment.

I think the error comes from the latest interface changes in HS, have you updated HS in your local env to the latest?

src/lib/homestore/heap_chunk_selector.cpp
xiaoxichen: The count hint is for selector to pick up a chunk that "best-effort" fulfill the `count` blks requirement.  
HeapChunkSelector by nature implemented this as it always pick the largest chunk so if the chunk picked doesnt have `count` blks then no other chunk can have `count` blks.

However other selector doesnt behave like this, for example the round-robin selector.  It should best check size before returning.

JacksonYao287:
@zichanglai  the error is caused by the upate of Homestore.

if I update to the latest homestore , there will be a test failure cased by ShardManagerTest.

if i do not update, it will fail the building here.
Screenshot 2023-09-21 at 21 08 05

so will try to address those problem

src/lib/homestore/heap_chunk_selector.cpp
yamingk: @JacksonYao287 are you going to remove `chunk_id_hint` check  and keep `mark_chunk_selected` API or keep it and do consume it in `select_chunk` and retire `mark_chunk_selected` api? Either way if fine. you can make the call.

yamingk: @JacksonYao287 They can both achieve the goal. Difference is more efficient during bootup, e.g. we don't need to remove every elements from the top, put them into a tmp vector until we found a match for the chunkid being marked as selected, them add all the chunk in the tmp vector back to the heap. 

Let's say we have 64K chunks, and 200 open chunks, during boot up, this single mark chunks selected operation will do 200 * 64K * Log (64K), and these operations can all be zeroed, if we build heap after ShardManager recovery completes.

src/lib/homestore/heap_chunk_selector.cpp
yamingk: it will be weird for the consumer to provide a `count` when calling heap to select a chunk.
We can make change to API signature from both sides later.

yamingk (approved):

src/lib/homestore/heap_chunk_selector.cpp
JacksonYao287: > Difference is more efficient during bootup
agree, it is more efficient to build per dev heap after we know all the selected chunk.

>are you going to remove chunk_id_hint check
we can , but after that it is the reposibility of homestore to make sure using selector correctly. since it will not waste a lot efficiency , we could keep it like this as a sanity check.

JacksonYao287: @yamingk i have make the change， pls take a look

xiaoxichen (approved):

yamingk (approved):

PR #34, opened 2023-09-01
SDSTOR-11601 add create_shard implementation

zichanglai:
this is the draft version of create_shard implementation in homeobject and can not build successfully until ReplDev in homestore is finalized.

### DIFF
--- a/src/include/homeobject/shard_manager.hpp
+++ b/src/include/homeobject/shard_manager.hpp
@@ -27,11 +27,15 @@ struct ShardInfo {
     uint64_t deleted_capacity_bytes;
     std::optional< peer_id > current_leader{std::nullopt};
 };
-using InfoList = std::list< ShardInfo >;
+
+using InfoList = std::list<ShardInfo>;
+using CShardInfo = std::shared_ptr<ShardInfo>;
+using CShardInfoList = std::list< CShardInfo >;
 
 class ShardManager : public Manager< ShardError > {
 public:
     static uint64_t max_shard_size(); // Static function forces runtime evaluation.
+    static uint64_t max_shard_num_in_pg();
 
     virtual AsyncResult< ShardInfo > get_shard(shard_id id) const = 0;
     virtual AsyncResult< InfoList > list_shards(pg_id id) const = 0;
--- a/src/include/homeobject/shard_manager.hpp
+++ b/src/include/homeobject/shard_manager.hpp
@@ -28,9 +28,7 @@ struct ShardInfo {
     std::optional< peer_id > current_leader{std::nullopt};
 };
 
-using InfoList = std::list<ShardInfo>;
-using CShardInfo = std::shared_ptr<ShardInfo>;
-using CShardInfoList = std::list< CShardInfo >;
+using InfoList = std::list< ShardInfo >;
 
 class ShardManager : public Manager< ShardError > {
 public:
--- a/src/lib/homeobject_impl.hpp
+++ b/src/lib/homeobject_impl.hpp
@@ -7,12 +7,6 @@
 
 #include <sisl/logging/logging.h>
 
-
-template <>
-struct std::hash< homeobject::ShardInfo > {
-    std::size_t operator()(homeobject::ShardInfo const& i) const noexcept { return std::hash< uint64_t >()(i.id); }
-};
-
 namespace home_replication {
 class ReplicationService;
 }
@@ -21,6 +15,26 @@ namespace homeobject {
 
 inline bool operator<(ShardInfo const& lhs, ShardInfo const& rhs) { return lhs.id < rhs.id; }
 inline bool operator==(ShardInfo const& lhs, ShardInfo const& rhs) { return lhs.id == rhs.id; }
+
+struct ShardInfoExt {
+    uint16_t chunk_id;
+};
+
+struct Shard {
+    explicit Shard(ShardInfo info) : info(std::move(info)) {}
+    ShardInfo info;
+    ShardInfoExt ext_info;
+};
+
+using ShardPtr = std::shared_ptr< Shard >;
+using ShardPtrList = std::list< ShardPtr >;
+
+struct PG {
+    explicit PG(PGInfo info) : pg_info(std::move(info)) {}
+    PGInfo pg_info;
+    uint64_t shard_sequence_num{0};
+    ShardPtrList shards;
+};
 
 class HomeObjectImpl : public HomeObject,
                        public BlobManager,
@@ -54,7 +68,7 @@ protected:
     std::map< pg_id, PG > _pg_map;
 
     mutable std::shared_mutex _shard_lock;
-    std::map< shard_id, CShardInfo > _shard_map;
+    std::map < shard_id, ShardPtr > _shard_map;
     ///
 public:
     explicit HomeObjectImpl(std::weak_ptr< HomeObjectApplication >&& application) :
@@ -84,6 +98,7 @@ public:
     ShardManager::AsyncResult< ShardInfo > create_shard(pg_id pg_owner, uint64_t size_bytes) final;
     ShardManager::AsyncResult< InfoList > list_shards(pg_id pg) const final;
     ShardManager::AsyncResult< ShardInfo > seal_shard(shard_id id) final;
+    uint64_t get_current_timestamp();
 
     static const std::string s_shard_info_sub_type;
     /// BlobManager
--- a/src/lib/homeobject_impl.hpp
+++ b/src/lib/homeobject_impl.hpp
@@ -16,24 +16,20 @@ namespace homeobject {
 inline bool operator<(ShardInfo const& lhs, ShardInfo const& rhs) { return lhs.id < rhs.id; }
 inline bool operator==(ShardInfo const& lhs, ShardInfo const& rhs) { return lhs.id == rhs.id; }
 
-struct ShardInfoExt {
-    uint16_t chunk_id;
-};
-
 struct Shard {
     explicit Shard(ShardInfo info) : info(std::move(info)) {}
     ShardInfo info;
-    ShardInfoExt ext_info;
+    uint16_t chunk_id;
 };
 
-using ShardPtr = std::shared_ptr< Shard >;
-using ShardPtrList = std::list< ShardPtr >;
+using ShardList = std::list< Shard >;
+using ShardIterator = ShardList::iterator;
 
 struct PG {
     explicit PG(PGInfo info) : pg_info(std::move(info)) {}
     PGInfo pg_info;
     uint64_t shard_sequence_num{0};
-    ShardPtrList shards;
+    ShardList shards;
 };
 
 class HomeObjectImpl : public HomeObject,
@@ -50,8 +46,7 @@ class HomeObjectImpl : public HomeObject,
     virtual BlobManager::Result< Blob > _get_blob(ShardInfo const&, blob_id) const = 0;
     virtual BlobManager::NullResult _del_blob(ShardInfo const&, blob_id) = 0;
     ///
-
-    folly::Future< ShardManager::Result< ShardInfo > > _get_shard(shard_id id) const;
+    folly::Future< ShardManager::Result< Shard > > _get_shard(shard_id id) const;
     auto _defer() const { return folly::makeSemiFuture().via(folly::getGlobalCPUExecutor()); }
 
 protected:
@@ -68,8 +63,9 @@ protected:
     std::map< pg_id, PG > _pg_map;
 
     mutable std::shared_mutex _shard_lock;
-    std::map < shard_id, ShardPtr > _shard_map;
+    std::map < shard_id, ShardIterator > _shard_map;
     ///
+    PGManager::Result< PG > _get_pg(pg_id pg);
 public:
     explicit HomeObjectImpl(std::weak_ptr< HomeObjectApplication >&& application) :
             _application(std::move(application)) {}
@@ -99,8 +95,6 @@ public:
     ShardManager::AsyncResult< InfoList > list_shards(pg_id pg) const final;
     ShardManager::AsyncResult< ShardInfo > seal_shard(shard_id id) final;
     uint64_t get_current_timestamp();
-
-    static const std::string s_shard_info_sub_type;
     /// BlobManager
     BlobManager::AsyncResult< blob_id > put(shard_id shard, Blob&&) final;
     BlobManager::AsyncResult< Blob > get(shard_id shard, blob_id const& blob, uint64_t off, uint64_t len) const final;
--- a/src/lib/homeobject_impl.hpp
+++ b/src/lib/homeobject_impl.hpp
@@ -20,6 +20,7 @@ struct Shard {
     explicit Shard(ShardInfo info) : info(std::move(info)) {}
     ShardInfo info;
     uint16_t chunk_id;
+    void* metablk_cookie{nullptr};
 };
 
 using ShardList = std::list< Shard >;
@@ -47,7 +48,7 @@ class HomeObjectImpl : public HomeObject,
     virtual BlobManager::NullResult _del_blob(ShardInfo const&, blob_id) = 0;
     ///
     folly::Future< ShardManager::Result< Shard > > _get_shard(shard_id id) const;
-    auto _defer() const { return folly::makeSemiFuture().via(folly::getGlobalCPUExecutor()); }
+    auto _defer() const { return folly::makeSemiFuture().via(folly::getGlobalCPUExecutor());}
 
 protected:
     std::mutex _repl_lock;
@@ -63,7 +64,7 @@ protected:
     std::map< pg_id, PG > _pg_map;
 
     mutable std::shared_mutex _shard_lock;
-    std::map < shard_id, ShardIterator > _shard_map;
+    std::map< shard_id, ShardIterator > _shard_map;
     ///
     PGManager::Result< PG > _get_pg(pg_id pg);
 public:
--- a/src/lib/homeobject_impl.hpp
+++ b/src/lib/homeobject_impl.hpp
@@ -79,6 +79,8 @@ public:
     // This is public but not exposed in the API above
     void init_repl_svc();
 
+    std::shared_ptr< home_replication::ReplicationService > get_repl_svc()  { return _repl_svc;}
+
     std::shared_ptr< BlobManager > blob_manager() final;
     std::shared_ptr< PGManager > pg_manager() final;
     std::shared_ptr< ShardManager > shard_manager() final;
--- a/src/lib/homeobject_impl.hpp
+++ b/src/lib/homeobject_impl.hpp
@@ -7,6 +7,7 @@
 
 #include <sisl/logging/logging.h>
 
+
 template <>
 struct std::hash< homeobject::ShardInfo > {
     std::size_t operator()(homeobject::ShardInfo const& i) const noexcept { return std::hash< uint64_t >()(i.id); }
@@ -27,9 +28,6 @@ class HomeObjectImpl : public HomeObject,
                        public ShardManager,
                        public std::enable_shared_from_this< HomeObjectImpl > {
 
-    std::mutex _repl_lock;
-    std::shared_ptr< home_replication::ReplicationService > _repl_svc;
-
     /// Implementation defines these
     virtual ShardManager::Result< ShardInfo > _create_shard(pg_id, uint64_t size_bytes) = 0;
     virtual ShardManager::Result< ShardInfo > _seal_shard(shard_id) = 0;
@@ -43,6 +41,9 @@ class HomeObjectImpl : public HomeObject,
     auto _defer() const { return folly::makeSemiFuture().via(folly::getGlobalCPUExecutor()); }
 
 protected:
+    std::mutex _repl_lock;
+    std::shared_ptr< home_replication::ReplicationService > _repl_svc;
+    //std::shared_ptr<homestore::ReplicationService> _repl_svc;  
     peer_id _our_id;
 
     /// Our SvcId retrieval and SvcId->IP mapping
@@ -50,19 +51,20 @@ protected:
 
     ///
     mutable std::shared_mutex _pg_lock;
-    using shard_set = std::unordered_set< ShardInfo >;
-    using pg_pair = std::pair< PGInfo, shard_set >;
-    std::map< pg_id, pg_pair > _pg_map;
+    std::map< pg_id, PG > _pg_map;
 
     mutable std::shared_mutex _shard_lock;
-    std::map< shard_id, shard_set::const_iterator > _shard_map;
+    std::map< shard_id, CShardInfo > _shard_map;
     ///
-
 public:
     explicit HomeObjectImpl(std::weak_ptr< HomeObjectApplication >&& application) :
             _application(std::move(application)) {}
 
     ~HomeObjectImpl() override = default;
+    HomeObjectImpl(const HomeObjectImpl&) = delete;
+    HomeObjectImpl(HomeObjectImpl&&) noexcept = delete;
+    HomeObjectImpl& operator=(const HomeObjectImpl&) = delete;
+    HomeObjectImpl& operator=(HomeObjectImpl&&) noexcept = delete;
 
     // This is public but not exposed in the API above
     void init_repl_svc();
@@ -83,6 +85,7 @@ public:
     ShardManager::AsyncResult< InfoList > list_shards(pg_id pg) const final;
     ShardManager::AsyncResult< ShardInfo > seal_shard(shard_id id) final;
 
+    static const std::string s_shard_info_sub_type;
     /// BlobManager
     BlobManager::AsyncResult< blob_id > put(shard_id shard, Blob&&) final;
     BlobManager::AsyncResult< Blob > get(shard_id shard, blob_id const& blob, uint64_t off, uint64_t len) const final;
--- a/src/lib/homestore/homeobject.cpp
+++ b/src/lib/homestore/homeobject.cpp
@@ -94,12 +94,11 @@ void HSHomeObject::on_shard_meta_blk_found(homestore::meta_blk* mblk, sisl::byte
     shard_info_str.append(r_cast<const char*>(buf.bytes()), size);
 
     auto shard = deserialize_shard(shard_info_str);
-    if (shard.info.state == ShardInfo::State::OPEN) {
-        // create shard;
-        do_commit_new_shard(shard);
-    } else {
-        do_commit_seal_shard(shard);
-    }
+    shard.metablk_cookie = mblk;
+
+    //As shard info in the homestore metablk is always the latest state(OPEN or SEALED),
+    //we can always create a shard from this shard info and once shard is deleted, the associated metablk will be deleted too.
+    do_commit_new_shard(shard);
 }
 
 } // namespace homeobject
--- a/src/lib/homestore/homeobject.cpp
+++ b/src/lib/homestore/homeobject.cpp
@@ -2,9 +2,12 @@
 
 #include <homestore/homestore.hpp>
 #include <homestore/index_service.hpp>
+#include <homestore/meta_service.hpp>
 #include <iomgr/io_environment.hpp>
 
 namespace homeobject {
+
+const std::string HomeObjectImpl::s_shard_info_sub_type = "shard_info";
 
 extern std::shared_ptr< HomeObject > init_homeobject(std::weak_ptr< HomeObjectApplication >&& application) {
     LOGINFOMOD(homeobject, "Initializing HomeObject");
@@ -25,6 +28,14 @@ void HSHomeObject::init_homestore() {
 
     LOGINFO("Starting iomgr with {} threads, spdk: {}", app->threads(), false);
     ioenvironment.with_iomgr(iomgr::iomgr_params{.num_threads = app->threads(), .is_spdk = app->spdk_mode()});
+
+    //register some callbacks for metadata recovery;
+    HomeStore::instance()->meta_service().register_handler(HomeObjectImpl::s_shard_info_sub_type,
+                                                     [this](homestore::meta_blk* mblk, sisl::byte_view buf, size_t size) {
+                                                         on_shard_meta_blk_found(mblk, buf, size);
+                                                     },
+                                                     nullptr,
+                                                     true);
 
     /// TODO Where should this come from?
     const uint64_t app_mem_size = 2 * Gi;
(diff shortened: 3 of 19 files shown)


szmyd:
I have one more abstraction to commit; please wait for that and then feel free to implement after rebasing.

szmyd:
This looks pretty good! Unfortunately I re-arranged things to be a bit cleaner/clearer when reading the code. Let me know if you need help merging my latest changes in.

zichanglai:
> This looks pretty good! Unfortunately I re-arranged things to be a bit cleaner/clearer when reading the code. Let me know if you need help merging my latest changes in.

thanks the reminder and do not worry, I will rebase on the latest code. :)

szmyd (changes requested):
Try and minimize changes to the API `src/include`

src/include/homeobject/shard_manager.hpp
```diff
@@ -27,11 +27,15 @@ struct ShardInfo {
     uint64_t deleted_capacity_bytes;
     std::optional< peer_id > current_leader{std::nullopt};
 };
+
 using InfoList = std::list< ShardInfo >;
+using ShardInfoPtr = std::shared_ptr< ShardInfo >;
+using ShardInfoPtrList = std::list< ShardInfoPtr >;
```
szmyd, line 33: It seems we do not need to expose these types above correct (move to `src/lib/homeobject_impl.hpp`)? We still return `std::list<ShardInfo>` in the Result<>.

src/lib/homeobject_impl.hpp
```diff
@@ -18,15 +19,22 @@ namespace homeobject {
 inline bool operator<(ShardInfo const& lhs, ShardInfo const& rhs) { return lhs.id < rhs.id; }
 inline bool operator==(ShardInfo const& lhs, ShardInfo const& rhs) { return lhs.id == rhs.id; }
 
+static constexpr uint32_t init_crc32 = 0;
+
+struct PG {
+    explicit PG(PGInfo info) : pg_info(std::move(info)) {}
+    PGInfo pg_info;
+    uint64_t shard_sequence_num{0};
+    ShardInfoPtrList shards;
```
szmyd, line 28: Why a `list`? ShardInfo's are sortable and should be unique in a given Pg. Should we not retain this being a `set`? Maybe we need to revisit the public API if that's the reason to use a `list` here?

src/lib/memory/shard_manager.cpp
```diff
@@ -6,38 +6,30 @@ namespace homeobject {
...
     auto pg_it = _pg_map.find(pg_owner);
     if (_pg_map.end() == pg_it) return folly::makeUnexpected(ShardError::UNKNOWN_PG);
 
     auto const now = get_current_timestamp();
-    auto& s_set = pg_it->second.second;
-    auto s_id = (((uint64_t)pg_owner) << 48) + s_set.size();
+    auto& shard_list = pg_it->second.shards;
+    auto s_id = (((uint64_t)pg_owner) << 48) + shard_list.size();
     auto info = ShardInfo{s_id, pg_owner, ShardInfo::State::OPEN, now, now, size_bytes, size_bytes, 0};
+    auto cinfo = std::make_shared<ShardInfo>(info);
```
szmyd, line 18: Can we combine the above two lines? No need for the temporary variable that I see.

szmyd (commented):
This looks good. I had a couple change requests so far. My main question is around the use of a `shared_ptr` around the `ShardInfo`s. I think it instead of the `iterator` method I was using, but this is more costly right and does not achieve much since we never have these `shared_ptr`'s leave *HomeObject*'s own context (we control them) (`std::set and std::list` values are already stored on the heap right, so we are adding another indirection)

src/lib/homestore/homeobject.cpp
```diff
@@ -20,4 +25,27 @@ void HomeObjectImpl::init_repl_svc() {
     }
 }
 
+HSHomeObject::HSHomeObject(std::weak_ptr< HomeObjectApplication >&& application) : HomeObjectImpl(std::move(application)) {
+    //register some callbacks for metadata recovery;
+    homestore::hs()->meta_service().register_handler(HomeObjectImpl::s_shard_info_sub_type,
```
szmyd, line 30: I plan on having an `init_homestore` in my other "Initialize HomeStore" PR since we'll likely have to hand it a weak/shared ptr to ourself for callbacks and can't happen in the constructor. We can move this there and keep the `using` default constructor after that PR merges.

src/include/homeobject/shard_manager.hpp
zichanglai: yes, these two types does not need to expose outside, will move to src/lib.

src/lib/homeobject_impl.hpp
zichanglai: shard info is sortable and we can use std::set here,  but I am not sure if it is needed to sort the shard info inside one pg. if we do not need sort, list is enough.

src/lib/memory/shard_manager.cpp
zichanglai: sure

src/lib/homestore/homeobject.cpp
zichanglai: sure, will do that after init homestore pr is merged.

zichanglai:
> This looks good. I had a couple change requests so far. My main question is around the use of a `shared_ptr` around the `ShardInfo`s. I think it instead of the `iterator` method I was using, but this is more costly right and does not achieve much since we never have these `shared_ptr`'s leave _HomeObject_'s own context (we control them) (`std::set and std::list` values are already stored on the heap right, so we are adding another indirection)

the reason why changes to std::shared_ptr is to simplify the seal_shard process. when we use shared_ptr, during the seal shard we only need to find the shard info ptr from _shard_map and update its content, but without touching anything inside _pg_map.  for now it needs to insert the new shard info to pg.shards to replace and update the returned iterator to _shard_map.

src/lib/homestore/shard_manager.cpp
```diff
@@ -20,4 +124,74 @@ ShardManager::Result< InfoList > HSHomeObject::_list_shards(pg_id pg) const {
     return folly::makeUnexpected(ShardError::UNKNOWN_PG);
 }
 
+void HSHomeObject::on_shard_message_commit(int64_t lsn, sisl::blob const& header, sisl::blob const& key,
+                                           void* user_ctx) {
```
JacksonYao287, line 128: this will be called in `on_commit` of statemachine , right? what will we do in `on_precommit`

src/lib/homestore/shard_manager.cpp
```diff
@@ -20,4 +124,74 @@ ShardManager::Result< InfoList > HSHomeObject::_list_shards(pg_id pg) const {
...
+    if (msg_header->header_crc != msg_header->calculate_crc()) {
+        LOGWARN("replication message header is corrupted with crc error, lsn:{}", lsn);
+        if (promise) {
+            promise->setValue(folly::makeUnexpected(ShardError::INVALID_ARG));	  
+        }
+        return;
+    }
+
+    std::string string_value;
+    string_value.append(r_cast<char*>(header.bytes + sizeof(ReplicationMessageHeader)), msg_header->payload_size);
```
JacksonYao287, line 147: should be
`string_value.append(r_cast<char*>(msg_header), sizeof(ReplicationMessageHeader))`? 

please correct me if i am wrong

src/lib/homestore/shard_manager.cpp
```diff
@@ -20,4 +124,74 @@ ShardManager::Result< InfoList > HSHomeObject::_list_shards(pg_id pg) const {
...
+    if (msg_header->payload_crc != crc) {
+        LOGWARN("replication message body is corrupted with crc error, lsn:{}", lsn);
+        if (promise) {
+            promise->setValue(folly::makeUnexpected(ShardError::INVALID_ARG));	  
+        }
+        return;
+    }
+
+    //TODO: preapre ShardInfoExt msg with chunk_id and persisted the ShardInfoExt to homestore MetaBlkService;
+    void* cookie = nullptr;
```
JacksonYao287, line 159: where is cookie saved? how to get cookie when restarting

JacksonYao287 (commented):
do we miss the implement `on_precommit` and `async_alloc_write` for now, and will be implemented later?

src/lib/homeobject_impl.hpp
zichanglai: if you want to get a specified shard info from the shard_id, you can use the _shard_map, no need to touch this structure

szmyd: The only time we touch this is when we move/delete a shard, so I'm okay with a list. It makes the list_shards() operation potentially simpler since the result is a list as well.

src/lib/homeobject_impl.hpp
```diff
@@ -39,26 +47,30 @@ class HomeObjectImpl : public HomeObject,
...
     ///
     mutable std::shared_mutex _pg_lock;
-    using shard_set = std::unordered_set< ShardInfo >;
-    using pg_pair = std::pair< PGInfo, shard_set >;
-    std::map< pg_id, pg_pair > _pg_map;
+    std::map< pg_id, PG > _pg_map;
 
     mutable std::shared_mutex _shard_lock;
-    std::map< shard_id, shard_set::const_iterator > _shard_map;
+    std::map< shard_id, ShardInfoPtr > _shard_map;
```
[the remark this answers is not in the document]
szmyd: `map` is better, iterators are not invalidated and we can continue to not need another level of indirection with the shared_ptr which is unnecessary.

src/lib/homestore/shard_manager.cpp
```diff
@@ -1,11 +1,115 @@
...
+    return (iter == _pg_map.end()) ? false : true;
+}
+
+shard_id HSHomeObject::generate_new_shard_id(pg_id pg) {
+    std::scoped_lock lock_guard(_pg_lock);
+    auto iter = _pg_map.find(pg);
+    if (iter == _pg_map.end()) {
+        auto created_pg = PG(std::move(PGInfo(pg)));
+        iter = _pg_map.insert(std::make_pair(pg, std::move(created_pg))).first;
+    }
```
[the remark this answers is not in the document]
szmyd: Yes; Pg creation can not be implicit since we need to create a new ReplDev etc which can fail. This operation would take too long.

src/lib/homestore/shard_manager.cpp
```diff
@@ -1,11 +1,115 @@
...
+    auto [p, sf] = folly::makePromiseContract< ShardManager::Result<ShardInfo> >();
+    repl_dev->async_alloc_write(sisl::blob(buf->data_begin(), needed_size),
+                                sisl::blob(),
+                                value,
+                                static_cast<void*>(&p));
+    auto info = std::move(sf).via(folly::getGlobalCPUExecutor()).get();
+    if (!bool(info)) {
+        LOGWARN("create new shard [{}] on pg [{}] is failed with error:{}", new_shard_id, pg_owner, info.error());
+    }
+    header->~ReplicationMessageHeader();
```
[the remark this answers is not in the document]
zichanglai: no, raw_ptr is a std::shared_ptr<>, it will free internal buffer when destruction.

src/lib/homestore/shard_manager.cpp
```diff
@@ -1,11 +1,115 @@
...
+    }
+
+    if (!repl_dev->is_leader()) {
+        LOGWARN("pg [{}] replica set is not a leader, please retry other pg members", pg_owner);
+        return folly::makeUnexpected(ShardError::NOT_LEADER);
+    }
+
+    auto new_shard_id = generate_new_shard_id(pg_owner);
+    auto create_time = get_current_timestamp();
+    auto shard_info = std::make_shared<ShardInfo>(ShardInfo(new_shard_id, pg_owner, ShardInfo::State::OPEN,
```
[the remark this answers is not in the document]
zichanglai: sure

src/lib/homeobject_impl.hpp
```diff
@@ -78,7 +90,9 @@ class HomeObjectImpl : public HomeObject,
     ShardManager::AsyncResult< ShardInfo > create_shard(pg_id pg_owner, uint64_t size_bytes) final;
     ShardManager::AsyncResult< InfoList > list_shards(pg_id pg) const final;
     ShardManager::AsyncResult< ShardInfo > seal_shard(shard_id id) final;
+    uint64_t get_current_timestamp();
 
+    static const std::string s_shard_info_sub_type;
```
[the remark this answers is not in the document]
zichanglai: ok, will move to HSHomeObject.

src/lib/homestore/shard_manager.cpp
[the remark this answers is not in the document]
zichanglai: will add a assert() here to make sure pg exists because pg exists is already checked pass before.

src/lib/homestore/shard_manager.cpp
```diff
@@ -1,11 +1,115 @@
...
 
 ShardManager::Result< ShardInfo > HSHomeObject::_create_shard(pg_id pg_owner, uint64_t size_bytes) {
-    return folly::makeUnexpected(ShardError::UNKNOWN_PG);
+    bool pg_exist = check_if_pg_exist(pg_owner);
+    if (!pg_exist) {
+        LOGWARN("failed to create shard with non-exist pg [{}]", pg_owner);
+        return folly::makeUnexpected(ShardError::UNKNOWN_PG);
+    }
+
+    auto repl_dev = _repl_svc->get_replica_dev(fmt::format("{}", pg_owner));
```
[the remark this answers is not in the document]
zichanglai: this is the ReplDev interface on HomeStore, but this interface is on Repl_merge branch and has not been merged to master branch yet.

src/lib/homestore/homeobject.hpp
```diff
@@ -20,10 +24,22 @@ class HSHomeObject : public HomeObjectImpl {
...
+    bool check_if_pg_exist(pg_id pg) const;
+    shard_id generate_new_shard_id(pg_id pg);
+    shard_id make_new_shard_id(pg_id pg, uint64_t sequence_num) const;
+    uint64_t get_sequence_num_from_shard_id(uint64_t shard_id);
+    std::string serialize_shard_info(ShardInfoPtr shard_info) const;
+    ShardInfoPtr deserialize_shard_info(const std::string& shard_info_str) const;
+    void do_commit_new_shard(ShardInfoPtr shard_info);
+    void do_commit_seal_shard(ShardInfoPtr shard_info);
 public:
     using HomeObjectImpl::HomeObjectImpl;
```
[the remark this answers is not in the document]
zichanglai: as HSHomeObject constructor is rollbacked and it is needed now.

src/lib/homestore/shard_manager.cpp
zichanglai: no, string_value is used to hold the msg body.

src/lib/homestore/shard_manager.cpp
zichanglai: cookie is not saved any more, it is usually to save cookie when you need to update homestore meta blks, but in our cases,   this meta blk will not be updated until now. when shard is sealed,  it will write to a new meta blks to homestore according to homestore design doc.

src/lib/homestore/shard_manager.cpp
zichanglai: on_precommit will call ReplDev to  alloc_blk() and it gives HeapChunkSelector a chance to bind a shard with the chunk.

src/lib/homestore/shard_manager.cpp
```diff
@@ -20,4 +124,74 @@ ShardManager::Result< InfoList > HSHomeObject::_list_shards(pg_id pg) const {
...
+}
+
+void HSHomeObject::do_commit_new_shard(ShardInfoPtr shard_info) {
+    std::scoped_lock lock_guard(_pg_lock, _shard_lock);  
+    auto pg_iter = _pg_map.find(shard_info->placement_group);
+    RELEASE_ASSERT(pg_iter != _pg_map.end(), "Missing PG info");
+    pg_iter->second.shards.push_back(shard_info);
+
+    //following part is must for follower members or when member is restarted;
+    auto sequence_num = get_sequence_num_from_shard_id(shard_info->id);
```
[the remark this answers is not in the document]
zichanglai: 1.follower will happens,  shard is generated from the leader side and  sequence num s increased only in leader side,  this gives the chance for follower to catch up sequence num with leader.
2. it is better to flush pg_sequence_number to disk but it will cause extra IO and I am trying to avoid it by recovering the lastest sequence num from meta blks, as all commited  shard creation is already persisted in meta blks.  there are still another two options for this from my side:   using raft LSN or  combined with (raft term , local sequence num) to be new proposal shard id

src/lib/homestore/shard_manager.cpp
```diff
@@ -1,11 +1,115 @@
...
+    std::memcpy(raw_ptr, create_shard_message.c_str(), create_shard_message.size());
+    
+    sisl::sg_list value;
+    //replicate this create shard message to PG members;
+    auto [p, sf] = folly::makePromiseContract< ShardManager::Result<ShardInfo> >();
+    repl_dev->async_alloc_write(sisl::blob(buf->data_begin(), needed_size),
+                                sisl::blob(),
+                                value,
+                                static_cast<void*>(&p));
+    auto info = std::move(sf).via(folly::getGlobalCPUExecutor()).get();
```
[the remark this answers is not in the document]
zichanglai: folly::SemiFuture always need a via() before get(), right?

src/lib/homeobject_impl.hpp
```diff
@@ -21,15 +16,32 @@ namespace homeobject {
 inline bool operator<(ShardInfo const& lhs, ShardInfo const& rhs) { return lhs.id < rhs.id; }
 inline bool operator==(ShardInfo const& lhs, ShardInfo const& rhs) { return lhs.id == rhs.id; }
 
+struct ShardInfoExt {
```
szmyd, line 19: How about just:
```
struct ShardInfoExt : public ShardInfo {
```
Then we can change `_get_shard` to return a `Result<ShardInfoExt>` and get rid of `Shard`? (or use `Shard` instead of `ShardInfoExt`; but just one new struct i think is needed, not two).

src/lib/homeobject_impl.hpp
```diff
@@ -43,26 +55,30 @@ class HomeObjectImpl : public HomeObject,
...
     ///
     mutable std::shared_mutex _pg_lock;
-    using shard_set = std::unordered_set< ShardInfo >;
-    using pg_pair = std::pair< PGInfo, shard_set >;
-    std::map< pg_id, pg_pair > _pg_map;
+    std::map< pg_id, PG > _pg_map;
 
     mutable std::shared_mutex _shard_lock;
-    std::map< shard_id, shard_set::const_iterator > _shard_map;
+    std::map < shard_id, ShardPtr > _shard_map;
```
szmyd, line 71: This can be `ShardList::const_iterator`.

src/lib/homeobject_impl.hpp
```diff
@@ -21,15 +16,32 @@ namespace homeobject {
...
+};
+
+using ShardPtr = std::shared_ptr< Shard >;
+using ShardPtrList = std::list< ShardPtr >;
+
+struct PG {
+    explicit PG(PGInfo info) : pg_info(std::move(info)) {}
+    PGInfo pg_info;
+    uint64_t shard_sequence_num{0};
+    ShardPtrList shards;
```
szmyd, line 36: We don't need `Shard` to be in a shared ptr; this can just be `ShardList` and we hold iterators in the `_shard_map` since the list will never be mutated. We can tombstone deleted shards etc. Removes an atomic counter that's not serving a purpose.

src/lib/homestore/shard_manager.cpp
```diff
@@ -1,15 +1,199 @@
...
+
+#include <homestore/homestore.hpp>
+#include <homestore/meta_service.hpp>
 
 namespace homeobject {
 
 uint64_t ShardManager::max_shard_size() { return Gi; }
+uint64_t ShardManager::max_shard_num_in_pg() {return ((uint64_t)0x01) << 48;}
+
+bool HSHomeObject::check_if_pg_exist(pg_id pg) const {
```
szmyd, line 12: Can we follow the pattern for `_get_shard` and do a `_get_pg` that returns the `Result<PG>` itself? The user doesn't have to consume it, but makes the code consistent for anything that looks up pgs.

szmyd (changes requested):
Have a couple requests to the HomeObjectImpl changes. HSHomeObject looks good so far, need the real HomeStore I realize.

src/lib/homeobject_impl.hpp
zichanglai: how about ShardList::Iterator?  because seal shard will change the status of shard info, so it can not be const, right?

src/lib/homeobject_impl.hpp
zichanglai: ok, will do that.

src/lib/homeobject_impl.hpp
zichanglai: ok, will do that

src/lib/homestore/shard_manager.cpp
zichanglai: good point, it is added now.

src/lib/homeobject_impl.hpp
szmyd: Yes `iterator` is fine; i just tend to make things more `const` if possible first and weaken the type if it doesn't work. You can always convert a const_iterator to non-const with:
```auto it = container.erase(const_it, const_it);```
but it looks odd in the code.

File: src/lib/homestore/shard_manager.cpp [hunk 21]
```diff
@@ -1,15 +1,263 @@
...
+    RELEASE_ASSERT(new_sequence_num < ShardManager::max_shard_num_in_pg(), "new shard id must be less than ShardManager::max_shard_num_in_pg()");
+    return make_new_shard_id(pg, new_sequence_num);
+}
+
+shard_id HSHomeObject::make_new_shard_id(pg_id pg, uint64_t sequence_num) const {
+    return ((uint64_t)pg << 48) | sequence_num; 
+}
+
+uint64_t HSHomeObject::get_sequence_num_from_shard_id(uint64_t shard_id) {
+    return shard_id & (0x0000FFFFFFFFFFFF);
```
szmyd, line 27: Should we do `& (max_shard_num_in_pg() - 1)` rather than a raw bitmask?

szmyd:
This PR still needs to take the create_shared tests from MemoryShardManager tests so we can get a code-coverage check. The few tests there should cover most paths we care about as is I think.

src/lib/homestore/shard_manager.cpp
zichanglai: good point, will update that.

src/lib/homestore/shard_manager.cpp
zichanglai: on_precommit and on_rollback is already added now. mostly their jobs is aligned with HomeObject design doc to write    ShardHeader to the HomeStore with BlkId returned, by that time the newly shard is binding with the specified chunk.

src/lib/homestore/shard_manager.cpp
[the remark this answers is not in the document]
zichanglai: yes, we can do it and it is updated.

src/lib/homestore/shard_manager.cpp
```diff
@@ -1,15 +1,263 @@
...
+    std::scoped_lock lock_guard(_pg_lock);
+    auto iter = _pg_map.find(pg);
+    RELEASE_ASSERT(iter != _pg_map.end(), "Missing pg info");
+    auto new_sequence_num = ++(iter->second.shard_sequence_num);
+    RELEASE_ASSERT(new_sequence_num < ShardManager::max_shard_num_in_pg(), "new shard id must be less than ShardManager::max_shard_num_in_pg()");
+    return make_new_shard_id(pg, new_sequence_num);
+}
+
+shard_id HSHomeObject::make_new_shard_id(pg_id pg, uint64_t sequence_num) const {
+    return ((uint64_t)pg << 48) | sequence_num;
```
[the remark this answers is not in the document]
zichanglai: a const variable SEQUENCE_BIT_NUM_IN_SHARD is added which will represent the bits of sequence num and max_shard_num_in_pg() is also aligned with it.

zichanglai:
> This PR still needs to take the create_shared tests from MemoryShardManager tests so we can get a code-coverage check. The few tests there should cover most paths we care about as is I think.

yes, homestore version ShardManager unit test is added in the latest commit.

src/lib/homestore/homeobject.cpp
```diff
@@ -20,4 +25,27 @@ void HomeObjectImpl::init_repl_svc() {
...
+    std::string shard_info_str;
+    shard_info_str.append(r_cast<const char*>(buf.bytes()), size);
+
+    ShardInfoPtr shard_info = deserialize_shard_info(shard_info_str);
+    if (shard_info->state == ShardInfo::State::OPEN) {
+        // create shard;
+        do_commit_new_shard(shard_info);
+    } else {
+        do_commit_seal_shard(shard_info);
+    }
```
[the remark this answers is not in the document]
JacksonYao287: my point here is if we do this synchrnously, we can only handle a metablk until finish handling previous one. if asynchrnously, we might handle all the metablk concurrently in the background threads.

well, for now, we can just keep it like this, and asynchronize them if necessary in the future.

src/lib/homestore/shard_manager.cpp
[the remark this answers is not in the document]
JacksonYao287: i prefer  lastest sequence num from meta blks, that is the source of truth

src/lib/homestore/shard_manager.cpp
```diff
@@ -1,15 +1,276 @@
...
+
+    auto new_shard_id = generate_new_shard_id(pg_owner);
+    auto create_time = get_current_timestamp();
+    auto shard_info = ShardInfo(new_shard_id, pg_owner, ShardInfo::State::OPEN,
+                                create_time, create_time, size_bytes, size_bytes, 0);
+    std::string create_shard_message = serialize_shard(Shard(shard_info));
+    //preapre msg header;
+    const uint32_t needed_size = sizeof(ReplicationMessageHeader) + create_shard_message.size();
+    auto buf = nuraft::buffer::alloc(needed_size);
+    uint8_t* raw_ptr = static_cast<uint8_t*>(buf->data_begin());
```
JacksonYao287, line 90: `buf->data_begin()` will return `byte*`, why static_cast here

src/lib/homestore/shard_manager.cpp
```diff
@@ -1,15 +1,276 @@
...
+        if (iter == _flying_shards.end()) {
+            LOGWARN("can not find flying shards on lsn {}", lsn);
+            if (promise) {
+                promise->setValue(folly::makeUnexpected(ShardError::UNKNOWN));
+            }
+            return;
+        }
+        shard.chunk_id = iter->second.chunk_id;
+        _flying_shards.erase(iter);
+        //deserialize the finalized shard msg;
```
JacksonYao287, line 231: nit: deserialize -> serialize

szmyd:
Can you merge `main`? I think that may be the reason we're not seeing coverage reports...

src/lib/homestore/shard_manager.cpp
zichanglai: good catch, will correct it.

zichanglai:
> Can you merge `main`? I think that may be the reason we're not seeing coverage reports...
> 
> Also force-pushing makes reviewing incredibly difficult. If we can just merge that'd be best; we can always squash at the end.

sure, I do not realize it before and will take cake of this from now.

src/lib/homestore/shard_manager.cpp
```diff
@@ -1,15 +1,276 @@
...
+    if (shard.info.state != ShardInfo::State::OPEN) {
+        //it is not an create shard msg, nothing to do;
+        return;
+    }
+    
+    std::scoped_lock lock_guard(_flying_shard_lock);
+    auto iter = _flying_shards.find(lsn);
+    if (iter != _flying_shards.end()) {
+        _flying_shards.erase(iter);
+    }
```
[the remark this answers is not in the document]
zichanglai: No, I think.  I think the written shard header will be used for recovery in case of meta blks are lost from the previous discussion meeting.

src/lib/homestore/shard_manager.cpp
zichanglai: good catch.

src/lib/homestore/shard_manager.cpp
```diff
@@ -1,15 +1,276 @@
...
+
+    std::string shard_msg;
+    if (!precheck_and_decode_shard_msg(lsn, header, key, &shard_msg)) {
+        if (promise) {
+            promise->setValue(folly::makeUnexpected(ShardError::INVALID_ARG));
+        }
+    }
+
+    auto shard = deserialize_shard(shard_msg);
+    if (shard.info.state == ShardInfo::State::OPEN) {
```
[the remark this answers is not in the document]
zichanglai: make sense and will take care of this.

src/lib/homestore/shard_manager.cpp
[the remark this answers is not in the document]
JacksonYao287: if we rollback the `pre_commit`, this means this is a canceled `create_shard`, why to recovery this?

coming to this, i have another question. we rolling back a `pre_commit`, seems nuraft does not append a log for rollback, what the statemachine does is just call the `rollback` function(i just only read the nuraft doc, but not deep dive the code for now, so if this is wrong, please correct me). so , if it crashes when we called `pre_commit` but before or not finish calling `statemachin#rollback` , after restart who will take the task to finish the rollback job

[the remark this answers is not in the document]
zichanglai: 1. yes, it is absolutely more clear to truncate/overwrite the shard header which is already written in on_pre_commit(), but homestore DataService is append-only model, I think overwritten is conflict with this model.  even this shard header is leaked,  I think it will be finalized recycled by chunk compaction, right? because this log is rollbacked and will never be commited, so from the homeobject side,  it can not see this shard any more(whether now or after recovery).
2. I think your mentioned case may happens,  restart happens after on_pre_commit() but before on_rollback(), in such case, this shard header is leaked, but same to question 1, it will be recycled finalized.

src/lib/homestore/shard_manager.cpp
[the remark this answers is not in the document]
zichanglai: that is the current implementation behaviors.

src/lib/homestore/shard_manager.cpp
zichanglai: I think metablk for this shard will be deleted too if while shard is deleted. for now, the cookie is saved in shard and can be used to handle to delete this shard info from homestore metablk service.

szmyd:
Nice; I think once we get the CodeCoverage up above 80% we can consider merging this. Let me know if any help is needed to get those unit tests made.

zichanglai:
> Nice; I think once we get the CodeCoverage up above 80% we can consider merging this. Let me know if any help is needed to get those unit tests made.

thanks, Brian.  I had added some more unit tests to improve code coverage to 81.99% with the latest few commits.

src/lib/homestore/shard_manager.cpp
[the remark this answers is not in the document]
zichanglai: for now, create shard is using mocked replication interface inside homeobject, will submit anther pr to adapt to ReplDev interface

szmyd (approved):
let's merge this and make fixes/api changes later. :+1:

src/lib/homestore/shard_manager.cpp
```diff
@@ -1,15 +1,293 @@
...
+    if (std::holds_alternative< home_replication:: ReplServiceError>(replica_set_var)) {
+        LOGWARN("failed to get replica set instance for pg [{}]", pg_owner);
+        return folly::makeUnexpected(ShardError::UNKNOWN_PG);	
+    }
+
+    auto replica_set = std::get< home_replication::rs_ptr_t >(replica_set_var);
+    if (!replica_set->is_leader()) {
+        LOGWARN("pg [{}] replica set is not a leader, please retry other pg members", pg_owner);
+        return folly::makeUnexpected(ShardError::NOT_LEADER);
+    }
```
szmyd, line 80: Oh; shard creation should be able to happen everywhere. We can relax this in M3 I think when we have real replicas.

szmyd: Just to follow up; this _has_ to be the case since by the time we actually call `append_entries` in HomeStore the leadership has changed since it was checked. We can never pre-check leadership...

xiaoxichen: maybe let HS return an error in this case?  Same situation is also true for write_blob request right?

szmyd: Yes; if we want to prevent this we can just turn `forwarding` off in the raft_server; it will reject `append_entries` if it's not currently the leader. But I'm not sure why we care about leadership here...we're basically just doing what the DM would retry anyways by forwarding it.

zichanglai: yes, if raft server on follower side support msg forward, then this check will not be necessary. before I do not know this feature is enabled in nuraft.

xiaoxichen: OK yes, only for the optimize_path. DM and GW need to adjust as well. 

Actually I was kind of not sure why  OM/CM team  need that strict regarding sending to leader :

Issue #27, opened 2023-08-24
Initialize HomeStore.


szmyd:
HomeObject has the responsibility of initializing HomeStore and its respective services. Configuration parameters should come from application through init_homeobject() routine and return only after all services are operational.

PR #40, opened 2023-09-06
#27 - Initialize HomeStore.

szmyd:


src/lib/homestore/homeobject.cpp
```diff
@@ -1,14 +1,60 @@
...
+void HSHomeObject::init_homestore() {
+    auto app = _application.lock();
+    RELEASE_ASSERT(app, "HomeObjectApplication lifetime unexpected!");
+
+    LOGINFO("Starting iomgr with {} threads, spdk: {}", app->threads(), false);
+    ioenvironment.with_iomgr(iomgr::iomgr_params{.num_threads = app->threads(), .is_spdk = app->spdk_mode()});
+
+    /// TODO Where should this come from?
+    const uint64_t app_mem_size = 2 * Gi;
+    LOGINFO("Initialize and start HomeStore with app_mem_size = {}", homestore::in_bytes(app_mem_size));
```
xiaoxichen, line 31: we usually do like this in POD spec,  that can copy any field in POD spec into ENV and make it available inside container.  Application can read from system env via `std::getenv()`.

In this case I believe we want to have `AVAIL_MEM_SIZE` set to POD's memory limit and take certain percentage from that for  each component. Maytbe same thing for the #thread? 

But these logic should better stay in SM? and pass in via APP?      
```
    env:
    - name: MINIO_GATEWAY_ACCOUNT_NAME
      valueFrom:
        fieldRef:
          apiVersion: v1
          fieldPath: metadata.labels['nuobject-account']
         
```

xiaoxichen:
```
    - name: POD_IP
      valueFrom:
        fieldRef:
          apiVersion: v1
          fieldPath: status.podIP
```

PR #42, opened 2023-09-07
SDSTOR-11612 get_shard

zichanglai:
get_shard should be common among homestore version and memory version.


szmyd (approved):
Agree'd; thanks!

PR #44, opened 2023-09-08
Split index from index svc

szmyd:


src/lib/memory/blob_manager.cpp
```diff
@@ -3,33 +3,43 @@
...
-    auto lg = std::scoped_lock(_index_lock);
-    auto [btree_it, h1] = _in_memory_index.try_emplace(shard.id, std::make_pair(btree(), blob_id(0ull)));
-    RELEASE_ASSERT(_in_memory_index.end() != btree_it, "Could not create BTree!");
-
-    auto route = BlobRoute{shard.id, btree_it->second.second++};
-    LOGDEBUGMOD(homeobject, "Writing BLOB {} to: BlkId:[{}]", route.blob, fmt::ptr(&blob));
+    auto route = BlobRoute{shard.id, 0ull};
+    auto index_it = _in_memory_index.end();
+    {
+        auto lg = std::scoped_lock(_index_lock);
```
xiaoxichen, line 9: It will better simulate the behavior (fail put_blob if shard doestn exists) if we put  shardIndex creation into `_create_shard` 
so that. in `_put_blob` we only take `shared_lock` and return an error if shard doestn exists

szmyd: Let me try that; one moment :+1:

src/lib/memory/homeobject.hpp
```diff
@@ -32,11 +32,18 @@ struct std::hash< homeobject::BlobRoute > {
 
 namespace homeobject {
 
+using btree = std::unordered_map< BlobRoute, Blob >;
+
+struct ShardIndex {
+    mutable std::shared_mutex _btree_lock;
+    btree _btree;
+    blob_id _shard_seq_num{0ull};
```
xiaoxichen, line 40: shard_id here?

src/lib/memory/homeobject.hpp
```diff
@@ -32,11 +32,18 @@ struct std::hash< homeobject::BlobRoute > {
 
 namespace homeobject {
 
+using btree = std::unordered_map< BlobRoute, Blob >;
+
+struct ShardIndex {
+    mutable std::shared_mutex _btree_lock;
+    btree _btree;
```
xiaoxichen, line 39: `bool sealed` ?

src/lib/memory/blob_manager.cpp
xiaoxichen: also if we can check `sealed` flag

src/lib/memory/homeobject.hpp
szmyd: it's "a blob_id" which represents the SSN. blob_id is the type.

src/lib/memory/homeobject.hpp
szmyd: sealed is stored in the metadata (ShardInfo) in the PGMap

src/lib/memory/homeobject.hpp
xiaoxichen: yes although blob_id is also uint64_t I mean `using shard_id uint64_t` probabaly make it cleaner.  Nevermind

src/lib/memory/blob_manager.cpp
```diff
@@ -2,34 +2,48 @@
...
+    auto [new_it, happened] = index_it->second._btree.try_emplace(route, std::move(blob));
     RELEASE_ASSERT(happened, "Generated duplicate BlobRoute!");
 
+    LOGDEBUGMOD(homeobject, "Wrote BLOB {} to: BlkId:[{}]", route.blob, fmt::ptr(&new_it->second));
     return route.blob;
 }
 
 BlobManager::Result< Blob > MemoryHomeObject::_get_blob(ShardInfo const& shard, blob_id blob) const {
+    auto index_it = _in_memory_index.cend();
+    {
```
xiaoxichen, line 30: replace this section with _find_index?  as the `iterator` will be implicitly cast to `const iterator`

szmyd: I wanted to do that; but i can't call a non-const method from const, and making it const makes it const so i can't _put on it anymore. I know there's a trick...but it's eluding me at the moment.

Issue #19, opened 2023-08-15
Implement HSHomeObject::_del


szmyd:
del should generate a json message that is appended to the ReplicaSet. The commit context should perform all data reservation cleanup required using the Homestore Index and Data services.

JacksonYao287:
#86

PR #50, opened 2023-09-13
#19: Tombstone blobs in emulated blob_manager as we will in HomeStore.

szmyd:


szmyd:
> still thinking about the scrubbing , just trying to propose doing soft-delete on Index when we deleting a blob and defer the hard delete till GC.
> 
> What i was trying to achieve is at any given time, let LSNm = min(committed LSN across all members of PG). The btree leafs (BlobRoutes) whose LSN <= LSNm should be 100% identical across all replicas. That gives us sort of consistent snapshot that we can compare across replicas w/o worrying the replication lag or new write IOs.
> 
> The blobID by nature have LSN but for delete we need to record LSN of the delete operation to the btree(as part of the soft delete).
> 
> Just some ideas in my mind, not solid enough about the cost/overhead

Tomb-stoning the entry is a "soft-delete" correct? It's not possible for there to be entries in the BTree that are greater than the current log_index (BlobID); maybe i'm not following what you want to do?

GC is unrelated to scrubbing right? GC is the act of reacquiring previously allocated disk space that is unaddressable without compacting/moving the Shard.

Scrubbing is the the act of marking Blobs no longer (or never) referenced by the MetaData as available for GC. Scrubbing member A should always require the same "Scrub" on all members and will be a normal BLOB Delete done through quorum.

xiaoxichen:
>We don't have to worry about consistency across members; that's what RAFT does for us.
Raft does that for online operations  i.e it ensures new modification are consistent across members but not taking care about the consistency of previous committed entries affected by bit error, code error or others. 

Silent bit error is easy to understand, if that happens on data part it could be detected by local scrub(comparing index vs data),  but if happens on Index then we need to compare across replicas.

Code error like raft only ensure you have a quorum "commit' but not ensuring your commit is properly done, for example if our commit writes to index svc and return, however for some code corner cases that update was not properly done,  raft cannot catch that.

So the scrubber does, is actually comparing all members and check if  they are identical , if we freeze and flush(complete all commits) the state machine.   In practise we cannot freeze the state machine,  then we need to scrub to a given LSN and ignore all new commits  to the state machine.

szmyd:
> >We don't have to worry about consistency across members; that's what RAFT does for us.
> 
> Raft does that for online operations  i.e it ensures new modification are consistent across members but not taking care about the consistency of previous committed entries affected by bit error, code error or others. 
> 
> Silent bit error is easy to understand, if that happens on data part it could be detected by local scrub(comparing index vs data),  but if happens on Index then we need to compare across replicas.
> 
> Code error like raft only ensure you have a quorum "commit' but not ensuring your commit is properly done, for example if our commit writes to index svc and return, however for some code corner cases that update was not properly done,  raft cannot catch that.
> 
> So the scrubber does, is actually comparing all members and check if  they are identical , if we freeze and flush(complete all commits) the state machine.   In practise we cannot freeze the state machine,  then we need to scrub to a given LSN and ignore all new commits  to the state machine.
> 
> 
At some point we have to trust our code. We are writing crc's in homestore for everything. Silent error is almost impossible. And all blob/shard data is duplicated between index and header right? Everything in the index is derivable locally with shard data.

PR #56, opened 2023-09-14
Cleanup memory::blob_manager.

szmyd:


src/lib/memory/blob_manager.cpp
```diff
@@ -2,7 +2,7 @@
 
 namespace homeobject {
 
-ShardIndex& MemoryHomeObject::_find_index(shard_id shard) {
+ShardIndex& MemoryHomeObject::_find_index(shard_id shard) const {
```
xiaoxichen, line 5: Oh the tricks here is use unique_ptr(or any ptr) so the const is on ptr

src/lib/memory/blob_manager.cpp
```diff
@@ -2,7 +2,7 @@
...
     auto index_it = _in_memory_index.find(shard);
     RELEASE_ASSERT(_in_memory_index.end() != index_it, "Missing BTree!!");
     return *index_it->second;
```
xiaoxichen, line 8: NPE here if shard doesnt exists

src/lib/memory/blob_manager.cpp
```diff
@@ -12,62 +12,69 @@ BlobManager::Result< blob_id > MemoryHomeObject::_put_blob(ShardInfo const& shar
     // Lookup Shard Index (BTree and SSN)
     auto& our_shard = _find_index(shard.id);
 
-    // Lock the BTree
-    // TODO: Generate BlobID (with RAFT this will happen implicitly) and Route
-    auto route = BlobRoute{shard.id, our_shard._shard_seq_num++};
+    // Generate BlobID (with RAFT this will happen implicitly) and Route
+    auto const route = BlobRoute{shard.id, our_shard._shard_seq_num++};
+    if ((UINT64_MAX >> 16) < route.blob) { return folly::makeUnexpected(BlobError::UNKNOWN); }
```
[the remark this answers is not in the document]
szmyd: Hmmm, yes. I think I was thinking shard good catch. Although no modern machine has >48 bits of addressable memory anyways. Will fix next time I'm in that code.

src/lib/memory/blob_manager.cpp
szmyd: NPE?

src/lib/memory/blob_manager.cpp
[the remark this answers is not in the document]
szmyd: Still waiting in a change to make the Blob's body movable outside the unique_ptr; so can remove this check then.

src/lib/memory/blob_manager.cpp
xiaoxichen: nevermind,  you catch that by RELEASE_ASSERT.

src/lib/memory/blob_manager.cpp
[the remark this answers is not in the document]
xiaoxichen: yes that part looks good

PR #59, opened 2023-09-15
Movable io_blob_safe does not require wrapping.

szmyd:


src/lib/blob_manager.cpp
```diff
@@ -31,4 +31,10 @@ BlobManager::NullAsyncResult HomeObjectImpl::del(shard_id shard, blob_id const&
     });
 }
 
+Blob Blob::clone() const {
+    auto new_body = sisl::io_blob_safe(body.size);
+    std::memcpy(new_body.bytes, body.bytes, body.size);
```
xiaoxichen, line 36: can we add deep copy into the copy assignment of io_blob_safe? 
are we on purpose delete that to avoid unconscious deep-copy?

xiaoxichen (approved):

src/lib/blob_manager.cpp
szmyd: You mean move the memcpy down? Yeah; we can talk to hari about that; it'd be a sisl change.

PR #68, opened 2023-09-22
adapt create_shard to use ReplDev of homestore

zichanglai:
This PR is used to adapt HO create_shard flow to use HS ReplDev instead of mock ReplicaSet, which will make HO shards survive from reboot.


src/lib/homestore/shard_manager.cpp
```diff
@@ -61,183 +62,119 @@ ShardManager::Result< ShardInfo > HSHomeObject::_create_shard(pg_id pg_owner, ui
...
-    header->payload_crc =
-        crc32_ieee(init_crc32, r_cast< const uint8_t* >(create_shard_message.c_str()), create_shard_message.size());
-    header->header_crc = header->calculate_crc();
-    raw_ptr += sizeof(ReplicationMessageHeader);
-    std::memcpy(raw_ptr, create_shard_message.c_str(), create_shard_message.size());
 
+    // round up msg size to homestore datasvc blk size
+    const auto datasvc_blk_size = homestore::HomeStore::instance()->data_service().get_blk_size();
+    auto msg_size = create_shard_message.size();
+    if (msg_size % datasvc_blk_size != 0) { msg_size = (msg_size / datasvc_blk_size + 1) * datasvc_blk_size; }
```
yamingk, line 83: sisl::round_up(msg_size, homestore::HomeStore::instance()->data_service().get_blk_size());

src/lib/homestore/shard_manager.cpp
```diff
@@ -61,183 +62,119 @@ ShardManager::Result< ShardInfo > HSHomeObject::_create_shard(pg_id pg_owner, ui
...
+    std::memset(r_cast< uint8_t* >(msg_buf), 0, msg_size);
+    std::memcpy(r_cast< uint8_t* >(msg_buf), create_shard_message.c_str(), create_shard_message.size());
+
+    // preapre msg header;
+    ReplicationMessageHeader header;
+    header.repl_group_id = pg.value().pg_info.id;
+    header.msg_type = ReplicationMessageType::CREATE_SHARD_MSG;
+    header.payload_size = msg_size;
+    header.payload_crc = crc32_ieee(init_crc32, r_cast< const uint8_t* >(msg_buf), msg_size);
+    header.header_crc = header.calculate_crc();
```
yamingk, line 94: Should we make it an internal crc, so that crc itself is also protected?

src/lib/homestore/replication_state_machine.cpp
```diff
@@ -23,36 +28,51 @@ void ReplicationStateMachine::on_commit(int64_t lsn, const sisl::blob& header, c
...
+        auto list_shard_result = _home_object->shard_manager()->list_shards(msg_header->repl_group_id).get();
+        if (!list_shard_result) {
+            LOGWARN("list shards failed with unknown pg {}", msg_header->repl_group_id);
+            break;
+        }
 
-    switch (msg_header->message_type) {
-    case ReplicationMessageType::SHARD_MESSAGE: {
-        _home_object->on_rollback_shard_msg(lsn, header, key, ctx);
+        if (list_shard_result.value().empty()) {
```
yamingk, line 55: nit: can we switch if/else condition since else branch is more likely to be hit?

if (!list_shard_result.value.empty()) { ... // do stuff
} else {
... // pg empty
}

src/lib/homestore/shard_manager.cpp
```diff
@@ -61,183 +62,119 @@ ShardManager::Result< ShardInfo > HSHomeObject::_create_shard(pg_id pg_owner, ui
...
-
-    if (promise) { promise->setValue(ShardManager::Result< ShardInfo >(shard.info)); }
+    // read value from PBA;
+    auto value_size = homestore::HomeStore::instance()->data_service().get_blk_size() * blkids.blk_count();
+    auto value_buf = iomanager.iobuf_alloc(512, value_size);
+    sisl::sg_list value;
+    value.size = value_size;
+    value.iovs.emplace_back(iovec{.iov_base = value_buf, .iov_len = value_size});
+
+    repl_dev.async_read(blkids, value, value_size)
```
yamingk, line 138: I believe we would be able to skip this `async_read` by providing `ShardInfo` into the context (a pinter to heap memory)?
And free the buffer after MetaSvc write the sb?

src/lib/homestore/shard_manager.cpp
```diff
@@ -61,183 +62,119 @@ ShardManager::Result< ShardInfo > HSHomeObject::_create_shard(pg_id pg_owner, ui
...
+    header.payload_crc = crc32_ieee(init_crc32, r_cast< const uint8_t* >(msg_buf), msg_size);
+    header.header_crc = header.calculate_crc();
     sisl::sg_list value;
+    value.size = msg_size;
+    value.iovs.push_back(iovec(msg_buf, msg_size));
     // replicate this create shard message to PG members;
     auto [p, sf] = folly::makePromiseContract< ShardManager::Result< ShardInfo > >();
-    replica_set->write(sisl::blob(buf->data_begin(), needed_size), sisl::blob(), value, static_cast< void* >(&p));
+    repl_dev.value()->async_alloc_write(sisl::blob(r_cast< uint8_t* >(&header), sizeof(header)), sisl::blob(), value,
+                                        static_cast< void* >(&p));
```
yamingk, line 101: can we pass `msg_buf` to be part of the context to save `async_read` in `on_shard_message_commit` ?

src/lib/homestore/shard_manager.cpp
```diff
@@ -61,183 +62,119 @@ ShardManager::Result< ShardInfo > HSHomeObject::_create_shard(pg_id pg_owner, ui
...
     auto info = std::move(sf).get();
     if (!bool(info)) {
         LOGWARN("create new shard [{}] on pg [{}] is failed with error:{}", new_shard_id & shard_mask, pg_owner,
                 info.error());
     }
-    header->~ReplicationMessageHeader();
+    iomanager.iobuf_free(msg_buf);
```
yamingk, line 107: if agree on my previous comment, move this to `on_shard_message_commit`.

src/lib/homestore/shard_manager.cpp
szmyd: One argument I had made to myself before I had made a similar comment was that we are not exposing the `how` for this message to HomeObject...so technically knowing that it's not some magic replicated spinning disk that just throws off random bits is not an assumption it should make? /shrug

src/lib/homestore/shard_manager.cpp
yamingk: Should be `reinterpret_cast`.

yamingk, line 101: Should be `reinterpret_cast`.

yamingk, line 101: p is a local var on stack and will be recycled after function exit, right?
Later in `on_shard_message_commit`, we will be accessing garbage pointer?

yamingk, line 101: I believe we also need to save the either the last allocated chunk_id assigned to this PG or the pdev_id hint itself to the context, so that when repldev calls get_alloc_blk_hints via listener, it knows what to return back as the hint, right?

src/lib/homestore/shard_manager.cpp
zichanglai: good suggestion and make it more clear.

src/lib/homestore/replication_state_machine.cpp
zichanglai: sure, will do that.

src/lib/homestore/shard_manager.cpp
zichanglai: yes, I will. in  hari's latest PR,  newly repl_req_ctx struct is added and header/key/value will be all added to this structure, so in on_commit() no more read is needed.

zichanglai: r_cast is a macro for reinterpret_cast and is defined in sisl

zichanglai: yes, but L102 will make this thread to block to wait util promise is set in on_commit(), that is why promise is still accessable in on_pre_commit().

zichanglai: last allocated chunk_id is needed to pass down to HS, but not via ctx. it will be passed down by here: https://github.com/eBay/HomeObject/pull/68/files#diff-8f4c55993d40418d0dbb2db0f58c7d8696fd7d793f488104752d4b889f4efeaaR61.
I will complete this part since PR#70 is already merged now.

src/lib/homestore/shard_manager.cpp
zichanglai: After hari's PR is merged, I am considering to replace msg_buf with hdr_buf_ in  ho_repl_ctx struct, when this struct is destruction, it will release this buf automatically.

src/lib/homestore/shard_manager.cpp
```diff
@@ -61,183 +62,119 @@ ShardManager::Result< ShardInfo > HSHomeObject::_create_shard(pg_id pg_owner, ui
...
-        break;
-    }
-    default: {
-        break;
-    }
-    }
-
-    if (promise) { promise->setValue(ShardManager::Result< ShardInfo >(shard.info)); }
+    // read value from PBA;
+    auto value_size = homestore::HomeStore::instance()->data_service().get_blk_size() * blkids.blk_count();
```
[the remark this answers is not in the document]
zichanglai: yes, good point and I will change to use homestore::HomeStore::instance()->repl_service().get_blk_size() when this interface is added to HS ReplicationService

src/lib/homestore/shard_manager.cpp
zichanglai: yes, when switch to use ho_repl_ctx, no extra read is needed any more.

src/lib/homestore/shard_manager.cpp
yamingk: I think in this case, it is more of protection from memory corruption which can not be from inside. Do you think it is an overkill from doing that?

src/lib/homestore/shard_manager.cpp
yamingk: Sounds good, Thanks.

src/lib/homestore/replication_state_machine.cpp
```diff
@@ -23,36 +28,51 @@ void ReplicationStateMachine::on_commit(int64_t lsn, const sisl::blob& header, c
...
-        _home_object->on_pre_commit_shard_msg(lsn, header, key, ctx);
-        break;
-    }
-    case ReplicationMessageType::PG_MESSAGE:
-    case ReplicationMessageType::BLOB_MESSAGE:
-    default: {
-        break;
-    }
-    }
+    // For shard creation, since homestore repldev inside will write shard header to data service first before this
```
xiaoxichen, line 31: I think we still need to track the _flying_shards and checks the uniqueness of shard.
The shard ID is given by master,  if leader change the new leader might propose a shardID that had been proposed.

src/lib/homestore/replication_state_machine.cpp
```diff
@@ -23,36 +28,51 @@ void ReplicationStateMachine::on_commit(int64_t lsn, const sisl::blob& header, c
...
-    switch (msg_header->message_type) {
-    case ReplicationMessageType::SHARD_MESSAGE: {
-        _home_object->on_rollback_shard_msg(lsn, header, key, ctx);
+        if (list_shard_result.value().empty()) {
+            // pg is empty without any shards, we leave the decision the HeapChunkSelector to select a pdev
+            // with most available space and select one chunk based on that pdev
+        } else {
+            auto chunk_id = _home_object->get_shard_chunk(list_shard_result.value().front().id);
+            RELEASE_ASSERT(!!chunk_id, "unknown shard id to get binded chunk");
+            // TODO:HS will add a new interface to get alloc hint based on a reference chunk;
```
xiaoxichen, line 61: this interface is there.
Do we want to cache the result into PG?

src/lib/homestore/shard_manager.cpp
zichanglai: Hi @yamingk , please check if the latest change is what you wants

src/lib/homestore/homeobject.cpp
```diff
@@ -38,21 +39,30 @@ void HSHomeObject::init_homestore() {
...
     if (need_format) {
-        HomeStore::instance()->format_and_start(std::map< uint32_t, hs_format_params >{
+        hsi->format_and_start(std::map< uint32_t, hs_format_params >{
             {HS_SERVICE::META, hs_format_params{.size_pct = 5.0}},
             {HS_SERVICE::LOG_REPLICATED, hs_format_params{.size_pct = 10.0}},
             {HS_SERVICE::LOG_LOCAL, hs_format_params{.size_pct = 5.0}},
-            {HS_SERVICE::DATA, hs_format_params{.size_pct = 50.0}},
             {HS_SERVICE::INDEX, hs_format_params{.size_pct = 30.0}},
+            {HS_SERVICE::REPLICATION,
+             hs_format_params{.size_pct = 50.0,
```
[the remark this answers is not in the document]
zichanglai: no, the pct here is the percent of whole HS space and hari's pr had already set it and I will keep this part is untouched.

src/lib/homestore/homeobject.cpp
```diff
@@ -38,21 +39,30 @@ void HSHomeObject::init_homestore() {
...
+                              .alloc_type = homestore::blk_allocator_type_t::append,
+                              .chunk_sel_type = homestore::chunk_selector_type_t::CUSTOM}},
```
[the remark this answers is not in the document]
zichanglai: same as above, will rebase on hari's PR.

src/lib/homestore/replication_state_machine.cpp
zichanglai: for now,  chunk_id is cached in Shard, I think it is very little cost to call pg-> shard list and then get chunk_id from one of the shard list.  but I am ok as it is an optimization.

zichanglai:
comments are addressed and already merge all changes in main branch, it is ready to review now.

src/lib/homestore/shard_manager.cpp
yamingk: it is external crc now, e.g. crc itself is excluded in header.claculate_crc(). 
I am fine with this approach, given the magic in header can also provide some protection against memory corruption. But ideally an internal crc is with protection to the crc field itself. I will leave it to you to decide.

src/lib/homestore/replication_message.hpp
```diff
@@ -7,21 +7,24 @@
...
-    ReplicationMessageType message_type;
+    replication_group_id repl_group_id; // replication group id upon which the msg is being replicating;
+    ReplicationMessageType msg_type;    // message type
     uint32_t payload_size;
     uint32_t payload_crc;
-    uint8_t reserved_pad[6]{};
+    uint8_t reserved_pad[4]{};
     uint32_t header_crc;
     uint32_t calculate_crc() const {
         return crc32_ieee(init_crc32, r_cast< const unsigned char* >(this), sizeof(*this) - sizeof(header_crc));
```
yamingk: to echo my previous comment, e.g. header_crc is excluded when being calculated.

src/lib/homestore_backend/hs_homeobject.cpp
```diff
@@ -87,16 +86,41 @@ HSHomeObject::~HSHomeObject() {
...
-    // we can always create a shard from this shard info and once shard is deleted, the associated metablk will be
-    // deleted too.
-    do_commit_new_shard(shard);
+    if (pg_is_recovery) {
+        auto hs_shard = std::make_shared< HS_Shard >(sb);
+        add_new_shard_to_map(hs_shard);
+        return;
+    }
+
+    // There is no guarantee that pg info will be recovery before shard recovery
```
yamingk, line 105: typo: recovered

src/lib/homestore_backend/hs_homeobject.cpp
```diff
@@ -87,16 +86,41 @@ HSHomeObject::~HSHomeObject() {
...
+void HSHomeObject::on_shard_meta_blk_found(homestore::meta_blk* mblk, sisl::byte_view buf) {
+    homestore::superblk< shard_info_superblk > sb;
+    sb.load(buf, mblk);
 
-    auto shard = deserialize_shard(shard_info_str);
-    shard.metablk_cookie = mblk;
+    bool pg_is_recovery = false;
+    {
+        std::scoped_lock lock_guard(_pg_lock);
+        pg_is_recovery = _pg_map.find(sb->placement_group) != _pg_map.end();
```
yamingk, line 96: `pg_is_recovered`

src/lib/homestore/shard_manager.cpp
zichanglai: ok, I think I got your point here, yes, it will be more reasonable if header_crc is included too. but if I do that, there is an extra place needed to save the crc of the whole header, right?

src/lib/homestore_backend/tests/test_shard_manager.cpp
```diff
@@ -176,223 +161,144 @@ TEST_F(ShardManagerWithShardsTesting, ListShards) {
...
+    EXPECT_EQ(1, pg_result->shards_.size());
+    auto& check_shard = pg_result->shards_.front();
+    EXPECT_EQ(ShardInfo::State::SEALED, check_shard->info.state);
 }
 
-TEST_F(ShardManagerWithShardsTesting, ShardManagerRecovery) {
+class FixtureAppWithRecovery : public FixtureApp {
+public:
+    std::list< std::filesystem::path > devices() const override {
+        const std::string fpath{"/tmp/test_homestore.data"};
```
yamingk, line 224: This same file name will clash with other test binary (not test cases in this cpp), if they run in parallel, right?
Can we apply `/tmp/<test_binary_name>_homestore.data` to make sure every test binary use their unique file name?

src/lib/homeobject_impl.hpp
```diff
@@ -94,7 +91,7 @@ class HomeObjectImpl : public HomeObject,
     std::map< pg_id_t, shared< PG > > _pg_map;
 
     mutable std::shared_mutex _shard_lock;
-    std::map< shard_id_t, ShardIterator > _shard_map;
+    std::map< shard_id_t, ShardPtr > _shard_map;
```
szmyd, line 94: There doesn't seem to be any reason for this change that I can see? This is more expensive; so we should maintain current code if possibe.

src/lib/homeobject_impl.hpp
```diff
@@ -38,15 +38,12 @@ inline shard_id_t make_new_shard_id(pg_id_t pg, shard_id_t next_shard) {
 
 struct Shard {
     explicit Shard(ShardInfo info) : info(std::move(info)) {}
+    virtual ~Shard() = default;
```
szmyd, line 41: Where is this used poly-morphically? It's not exposed above HomeObject, or below; so we shouldn't need a vtable.

src/lib/homestore_backend/hs_homeobject.cpp
zichanglai: yes, will update.

src/lib/homestore_backend/hs_homeobject.cpp
zichanglai: ok, will update.

src/lib/homestore_backend/tests/test_shard_manager.cpp
zichanglai: ok, I will align with Hari's PR to add a random number to file path

src/lib/homestore/shard_manager.cpp
yamingk: no, crc itself can still be inside this data structure. 
you can refer to this if you decide to do similar (not a must): https://github.com/eBay/HomeStore/blob/master/src/lib/device/device.h#L50

src/lib/homeobject_impl.hpp
zichanglai: In hs_homeobject.hpp,  I had added a new derived class HS_Shard which will including some members such as HS superblock, which is only meaning to HS version homeobject and I do not want to expose these details to Shard level.

src/lib/homeobject_impl.hpp
zichanglai: ok, will change back.

src/lib/homestore/shard_manager.cpp
zichanglai: sure, will take a look and apply.

src/lib/homestore_backend/hs_homeobject.hpp
```diff
@@ -58,38 +70,53 @@ class HSHomeObject : public HomeObjectImpl {
         static PGInfo pg_info_from_sb(homestore::superblk< pg_info_superblk > const& sb);
     };
 
+    struct HS_Shard : public Shard {
+        homestore::superblk< shard_info_superblk > sb_;
+        HS_Shard(ShardInfo info, homestore::chunk_num_t chunk_id);
+        HS_Shard(homestore::superblk< shard_info_superblk > const& sb);
+        virtual ~HS_Shard() = default;
```
szmyd, line 77: `~HS_Shard() override = default`

src/lib/homeobject_impl.hpp
szmyd: I see; so it can be a `unique_ptr` then?

src/lib/homestore_backend/hs_homeobject.cpp
```diff
@@ -87,16 +86,41 @@ HSHomeObject::~HSHomeObject() {
...
+}
+
+void HSHomeObject::on_shard_meta_blk_recover_completed(bool success) {
+    // Find all shard with opening state and excluede their binding chunks from the HeapChunkSelector;
+    RELEASE_ASSERT(pending_recovery_shards_.empty(), "some shards is still pending on recovery");
+    std::unordered_set< homestore::chunk_num_t > excluding_chunks;
+    std::scoped_lock lock_guard(_pg_lock);
+    for (auto& pair : _pg_map) {
+        for (auto& shard : pair.second->shards_) {
+            if (shard->info.state == ShardInfo::State::OPEN) {
```
sanebay, line 117: Does this mean if we havent fill the shards and they are open and they are not reused after recovery ?

src/lib/homestore_backend/hs_shard_manager.cpp
```diff
@@ -27,34 +28,31 @@ uint64_t HSHomeObject::get_sequence_num_from_shard_id(uint64_t shard_id_t) {
...
     shard_info.available_capacity_bytes = shard_json["shard_info"]["available_capacity"].get< uint64_t >();
     shard_info.total_capacity_bytes = shard_json["shard_info"]["total_capacity"].get< uint64_t >();
     shard_info.deleted_capacity_bytes = shard_json["shard_info"]["deleted_capacity"].get< uint64_t >();
-    auto shard = Shard(shard_info);
-    shard.chunk_id = shard_json["ext_info"]["chunk_id"].get< uint16_t >();
-    return shard;
+    return shard_info;
 }
 
 ShardManager::Result< ShardInfo > HSHomeObject::_create_shard(pg_id_t pg_owner, uint64_t size_bytes) {
```
sanebay, line 58: This has to AsyncResult. Blob manager changed it in another PR.

src/lib/homestore_backend/hs_shard_manager.cpp
```diff
@@ -68,187 +66,194 @@ ShardManager::Result< ShardInfo > HSHomeObject::_create_shard(pg_id_t pg_owner,
...
+    sisl::blob header;
+    header.bytes = r_cast< uint8_t* >(&req->header_);
+    header.size = sizeof(req->header_);
+    sisl::sg_list value;
+    value.size = msg_size;
+    value.iovs.push_back(iovec(buf_ptr, msg_size));
     // replicate this create shard message to PG members;
-    repl_dev->async_alloc_write(req->hdr_buf_, sisl::blob{}, sisl::sg_list{}, req);
+    repl_dev->async_alloc_write(header, sisl::blob{}, value, req);
     auto info = req->result().get();
```
sanebay, line 99: Its blocking. As brian suggested, we need to change to AsyncResult.

src/lib/homestore_backend/hs_shard_manager.cpp
```diff
@@ -68,187 +66,194 @@ ShardManager::Result< ShardInfo > HSHomeObject::_create_shard(pg_id_t pg_owner,
...
-
-    header->~ReplicationMessageHeader();
+    if (!bool(info)) {
+        LOGW("create new shard [{}] on pg [{}] is failed with error:{}", new_shard_id & shard_mask, pg_owner,
+             info.error());
+    }
     return info;
 }
 
 ShardManager::Result< ShardInfo > HSHomeObject::_seal_shard(shard_id_t id) {
```
sanebay, line 107: AsyncResult for all api's

src/lib/homestore_backend/hs_shard_manager.cpp
```diff
@@ -68,187 +66,194 @@ ShardManager::Result< ShardInfo > HSHomeObject::_create_shard(pg_id_t pg_owner,
...
     repl_result_ctx< ShardManager::Result< ShardInfo > >* ctx{nullptr};
     if (hs_ctx != nullptr) {
         ctx = boost::static_pointer_cast< repl_result_ctx< ShardManager::Result< ShardInfo > > >(hs_ctx).get();
     }
 
-    std::string shard_msg;
-    if (!precheck_and_decode_shard_msg(lsn, header, key, &shard_msg)) {
+    if (header.corrupted()) {
+        LOGW("replication message header is corrupted with crc error, lsn:{}", lsn);
+        if (ctx) { ctx->promise_.setValue(folly::makeUnexpected(ShardError::INVALID_ARG)); }
```
sanebay, line 156: Its not invalid arg from user. A new error HEADER or MSG CORRUPT ?

src/lib/homestore_backend/hs_shard_manager.cpp
```diff
@@ -68,187 +66,194 @@ ShardManager::Result< ShardInfo > HSHomeObject::_create_shard(pg_id_t pg_owner,
...
+        return;
+    }
+
+    if (crc32_ieee(init_crc32, value.bytes, value.size) != header.payload_crc) {
+        // header & value is inconsistent;
+        LOGW("replication message header is inconsistent with value, lsn:{}", lsn);
         if (ctx) { ctx->promise_.setValue(folly::makeUnexpected(ShardError::INVALID_ARG)); }
```
sanebay, line 163: same here.

src/lib/homestore_backend/hs_shard_manager.cpp
```diff
@@ -68,187 +66,194 @@ ShardManager::Result< ShardInfo > HSHomeObject::_create_shard(pg_id_t pg_owner,
...
     auto iter = shards.emplace(shards.end(), shard);
-    auto [_, happened] = _shard_map.emplace(shard.info.id, iter);
+    auto [_, happened] = _shard_map.emplace(shard->info.id, iter);
     RELEASE_ASSERT(happened, "duplicated shard info");
 
-    // following part is must for follower members or when member is restarted;
-    auto sequence_num = get_sequence_num_from_shard_id(shard.info.id);
+    // following part gives follower members a chance to catch up shard sequence num;
+    auto sequence_num = get_sequence_num_from_shard_id(shard->info.id);
     if (sequence_num > pg_iter->second->shard_sequence_num_) { pg_iter->second->shard_sequence_num_ = sequence_num; }
```
sanebay, line 200: We need to update the PG metablk everytime a sequence number is updated. Else it will be lost.

src/lib/homestore_backend/hs_shard_manager.cpp
```diff
@@ -68,187 +66,194 @@ ShardManager::Result< ShardInfo > HSHomeObject::_create_shard(pg_id_t pg_owner,
...
 
-    std::scoped_lock lock_guard(_flying_shard_lock);
-    auto iter = _flying_shards.find(lsn);
-    if (iter != _flying_shards.end()) { _flying_shards.erase(iter); }
+    // hs_ctx will be nullptr when HS is restarting and replay all commited log entries from the last checkpoint;
+    // but do we really need to handle this for create_shard or seal_shard?
+    // because this raft log had been commit before HO restarts and commit result is already saved in HS metablks
+    // when HS restarts, all PG/Shard infos will be recovered from HS metablks and if we commit again, it will cause
+    // duplication;
+#if 0
```
yamingk, line 126: Can this #if 0 code be removed?

src/lib/homestore_backend/hs_shard_manager.cpp
yamingk: `ShardError::CRC_MISMATCH`?

src/lib/homestore_backend/replication_state_machine.cpp
```diff
@@ -2,67 +2,76 @@
...
 
-    switch (msg_header->message_type) {
-    case ReplicationMessageType::SHARD_MESSAGE: {
-        _home_object->on_rollback_shard_msg(lsn, header, key, ctx);
+        if (list_shard_result.value().empty()) {
+            // pg is empty without any shards, we leave the decision the HeapChunkSelector to select a pdev
+            // with most available space and then select one chunk based on that pdev
+        } else {
+            auto chunk_id = home_object_->get_shard_chunk(list_shard_result.value().back().id);
+            RELEASE_ASSERT(!!chunk_id, "unknown shard id to get binded chunk");
```
yamingk, line 53: nit: `chunk_id.hasValue()` can be more readable?

src/lib/homestore_backend/replication_state_machine.cpp
```diff
@@ -2,67 +2,76 @@
...
 
-void ReplicationStateMachine::on_rollback(int64_t lsn, sisl::blob const& header, sisl::blob const& key,
-                                          cintrusive< homestore::repl_req_ctx >& ctx) {
-    LOGI("rollback with lsn:{}", lsn);
+homestore::blk_alloc_hints ReplicationStateMachine::get_blk_alloc_hints(sisl::blob const& header,
+                                                                        cintrusive< homestore::repl_req_ctx >& ctx) {
     const ReplicationMessageHeader* msg_header = r_cast< const ReplicationMessageHeader* >(header.bytes);
+    switch (msg_header->msg_type) {
+    case ReplicationMessageType::CREATE_SHARD_MSG: {
+        auto list_shard_result = home_object_->shard_manager()->list_shards(msg_header->pg_id).get();
```
yamingk, line 42: can we simplify the logic here a little bit by `home_object_->get_any_chunk_id(msg_header->pg_id)` which in its implementation just call shard_manager to pick *any* shard and return's chunk_id? It seems uncessary to return the full shard lists which can be thousands of shards, right?

src/lib/homestore_backend/hs_homeobject.cpp
zichanglai: no, they will be reused after recovery because all shards will be recovered from HS metablks and opening shards are included too. here logic is to excluded chunks which has opening shards from the heap chunkselector, so they will not be re-used to create shard util current shards is sealed.

src/lib/homestore_backend/hs_homeobject.hpp
zichanglai: will update it.

src/lib/homestore_backend/hs_shard_manager.cpp
zichanglai: Agreed.

src/lib/homestore_backend/hs_shard_manager.cpp
zichanglai: Agreed.

src/lib/homestore_backend/hs_shard_manager.cpp
zichanglai: no, when shard info is recovered from HS meta blks,  it will decode the sequence num from the shard_id and update pg sequence num if shard has larger sequence num

src/lib/homestore_backend/hs_shard_manager.cpp
zichanglai: will do it.

src/lib/homestore/replication_state_machine.cpp
zichanglai: this optimization is already added in the latest PR.

src/lib/homeobject_impl.hpp
zichanglai: yes, it is already updated.

src/lib/homestore_backend/hs_shard_manager.cpp
```diff
@@ -68,187 +66,192 @@ ShardManager::Result< ShardInfo > HSHomeObject::_create_shard(pg_id_t pg_owner,
...
-    return (*shard_iter->second).metablk_cookie;
+    if (shard_iter == _shard_map.end()) { return folly::makeUnexpected(ShardError::UNKNOWN_SHARD); }
+    auto hs_shard = d_cast< HS_Shard* >((*shard_iter->second).get());
+    return hs_shard->sb_->chunk_id;
+}
+
+std::optional< homestore::chunk_num_t > HSHomeObject::get_any_chunk_id(pg_id_t const pg_id) {
+    std::scoped_lock lock_guard(_pg_lock);
+    auto pg_iter = _pg_map.find(pg_id);
+    RELEASE_ASSERT(pg_iter != _pg_map.end(), "Missing PG info");
```
yamingk, line 202: Suggest not do assert, but return empty and let caller to check if this pg_id has been created in the map before, since ret value is `std::optional`, which also gives caller some flexibility to call with pg_id that may or may not be added to the map yet.

src/lib/homestore_backend/replication_state_machine.cpp
```diff
@@ -2,67 +2,69 @@
...
+            return home_object_->chunk_selector()->chunk_to_hints(any_allocated_chunk_id.value());
+        }
         break;
     }
-    case ReplicationMessageType::PG_MESSAGE:
-    case ReplicationMessageType::BLOB_MESSAGE:
+
+    case ReplicationMessageType::SEAL_SHARD_MSG: {
+        auto chunk_id = home_object_->get_shard_chunk(msg_header->shard_id);
+        RELEASE_ASSERT(!!chunk_id, "unknown shard id to get binded chunk");
```
yamingk, line 54: `!!chunk_id -> chunk_id.hasValue()` ?

src/lib/homestore_backend/tests/test_shard_manager.cpp
```diff
@@ -104,26 +109,7 @@ TEST_F(ShardManagerTesting, ListShardsOnEmptyPg) {
...
-        auto replica_set =
-            dynamic_cast< home_replication::MockReplicaSet* >(std::get< home_replication::rs_ptr_t >(rs).get());
-        _listener = std::make_shared< homeobject::ReplicationStateMachine >(ho);
-        replica_set->set_listener(_listener);
-#endif
-    }
-};
-
-TEST_F(ShardManagerWithShardsTesting, CreateShardSuccess) {
+TEST_F(ShardManagerTesting, CreateShardSuccess) {
```
yamingk, line 112: We need a test case with one PG, and create more than one Shard inside this PG. The 2nd and later shards created in this PG should arrive on chunk which belongs to same pdev_id. 

Also can we create 2 PGs and each PG with more than 1 shard created?

zichanglai:
The build failure comes from UT fails by the latest HomeStore.  currently HomeObject UT will be failed without this PR too but it is ok with HS 4.5.1, still not sure why the latest change in HS will cause UT failed. error msg shows ut will crash when HS indexservice is stopping.

src/lib/homestore_backend/hs_homeobject.cpp
```diff
@@ -87,16 +86,40 @@ HSHomeObject::~HSHomeObject() {
...
-    // deleted too.
-    do_commit_new_shard(shard);
+    if (pg_is_recovered) {
+        add_new_shard_to_map(std::make_unique< HS_Shard >(sb));
+        return;
+    }
+
+    // There is no guarantee that pg info will be recovered before shard recovery
+    std::scoped_lock lock_guard(recovery_mutex_);
+    pending_recovery_shards_[sb->placement_group].push_back(std::move(sb));
```
JacksonYao287, line 106: IIRC, all the callbacks of all the subsystems of metaservice will not be callled concurrently , but sequenciallly in this fuction (https://github.com/eBay/HomeStore/blob/bf8e6fdf9a1a417dd5aaf7be7328b4d015dd1caa/src/lib/meta/meta_blk_service.cpp#L1074). do we need this lock here? 

please correct me if i am wrong

src/lib/homestore_backend/hs_homeobject.cpp
```diff
@@ -87,16 +86,40 @@ HSHomeObject::~HSHomeObject() {
...
-    // we can always create a shard from this shard info and once shard is deleted, the associated metablk will be
-    // deleted too.
-    do_commit_new_shard(shard);
+    if (pg_is_recovered) {
+        add_new_shard_to_map(std::make_unique< HS_Shard >(sb));
+        return;
+    }
+
+    // There is no guarantee that pg info will be recovered before shard recovery
+    std::scoped_lock lock_guard(recovery_mutex_);
```
JacksonYao287, line 105: IIRC, the callbacks of metaservice will not be called concurrently, but sequentially in the function(https://github.com/eBay/HomeStore/blob/bf8e6fdf9a1a417dd5aaf7be7328b4d015dd1caa/src/lib/meta/meta_blk_service.cpp#L1074). so , do we need this lock here

src/lib/homestore_backend/hs_shard_manager.cpp
```diff
@@ -68,187 +66,192 @@ ShardManager::Result< ShardInfo > HSHomeObject::_create_shard(pg_id_t pg_owner,
...
-                                                   shard.metablk_cookie);
-        // update in-memory shard map;
-        do_commit_new_shard(shard);
+    if (crc32_ieee(init_crc32, value.bytes, value.size) != header.payload_crc) {
+        // header & value is inconsistent;
+        LOGW("replication message header is inconsistent with value, lsn:{}", lsn);
+        if (ctx) { ctx->promise_.setValue(folly::makeUnexpected(ShardError::CRC_MISMATCH)); }
+        return;
+    }
+
```
JacksonYao287, line 146: if the header or value corrupted, the log can not be applied anymore, should we just crash ?

src/lib/homestore_backend/hs_homeobject.hpp
```diff
@@ -45,51 +45,80 @@ class HSHomeObject : public HomeObjectImpl {
...
+        uint64_t deleted_capacity_bytes;
+        homestore::chunk_num_t chunk_id;
+    };
 #pragma pack()
 
     struct HS_PG : public PG {
         homestore::superblk< pg_info_superblk > pg_sb_;
         shared< homestore::ReplDev > repl_dev_;
-
+        std::optional< homestore::chunk_num_t > any_allocated_chunk_id_{};
```
JacksonYao287, line 65: where will this be used, want to get the pdev_id here?
it is a little wired

src/lib/homestore_backend/hs_shard_manager.cpp
zichanglai: updated.

src/lib/homestore_backend/hs_shard_manager.cpp
zichanglai: same as above.

src/lib/homestore_backend/hs_shard_manager.cpp
zichanglai: such cases are added now.

src/lib/homestore_backend/replication_state_machine.cpp
zichanglai: updated.

src/lib/homestore_backend/tests/test_shard_manager.cpp
zichanglai: such test cases are added in the latest pr.

src/lib/homestore_backend/hs_homeobject.hpp
zichanglai: no, just to cache any chunk_id from random shard for pg.  thus it can use it to call chunk selector->chunk_to_hints() and no need to list shard from the pg any more for per shard create request

src/lib/homestore_backend/hs_homeobject.cpp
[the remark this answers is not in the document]
zichanglai: yes, but on_pg_meta_blk_found() is a async function.

src/lib/homestore_backend/hs_homeobject.hpp
JacksonYao287: chunk_to_hint() is used to get pdev_id , why not just save pdev_id here

src/lib/homestore_backend/hs_homeobject.cpp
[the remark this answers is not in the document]
zichanglai: no, I check the code in HS again,  the associated promise will be set in HS ReplServiceImpl::rd_super_blk_found(), which is expected to be called in HS metablk replay thread, then I think it on_pg_meta_blk_found#thenvalue() is also executed in metablk replay thread, so I think it is executed sequentially and no-need lock here.

src/lib/homestore_backend/hs_homeobject.cpp
```diff
@@ -87,16 +86,40 @@ HSHomeObject::~HSHomeObject() {
...
+    pending_recovery_shards_[sb->placement_group].push_back(std::move(sb));
+}
+
+void HSHomeObject::on_shard_meta_blk_recover_completed(bool success) {
+    // Find all shard with opening state and excluede their binding chunks from the HeapChunkSelector;
+    RELEASE_ASSERT(pending_recovery_shards_.empty(), "some shards is still pending on recovery");
+    std::unordered_set< homestore::chunk_num_t > excluding_chunks;
+    std::scoped_lock lock_guard(_pg_lock);
+    for (auto& pair : _pg_map) {
```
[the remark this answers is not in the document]
zichanglai: same as above.

src/lib/homestore_backend/hs_homeobject.hpp
zichanglai: I want to do it before as it is more straightforward. but ShardManager only know the shard to chunk map and not-aware of pdev id at all,  so it cache chunk id here and leave the chunk to pdev map to heap chunk selector.

src/lib/homestore_backend/hs_homeobject.cpp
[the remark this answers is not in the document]
zichanglai: lock is removed

src/lib/homestore_backend/hs_shard_manager.cpp
zichanglai: create_shard is updated and I trend to leave the other apis to the implementation author to make the decision

zichanglai:
All HO UTs are running normally now.

src/lib/homestore_backend/hs_shard_manager.cpp
zichanglai: actually I trend to use replace_pg() method to use a new pg to replace the existing corrupted pg in such case(shard replacement will be better if it is supported)

yamingk (approved):

PR #70, opened 2023-09-25
add interface to get blk alloc hints for chunks on same pg

zichanglai:
HO will try to make all shards on same PG to be located in same HS pdev, but current HO does not know pdev id at all because pdev is managed by HS. In order to do that, we can reuse the HeapChunkSelector logic and its inner data members and add an interface which will return blk_alloc_hints for different shards on same PG.

### DIFF
--- a/src/lib/homestore/heap_chunk_selector.cpp
+++ b/src/lib/homestore/heap_chunk_selector.cpp
@@ -118,10 +118,10 @@ void HeapChunkSelector::build_per_dev_chunk_heap(const std::unordered_set< chunk
     };
 }
 
-homestore::blk_alloc_hints HeapChunkSelector::get_blk_alloc_hints_on_same_pdev(chunk_num_t chunk_id) const {
+homestore::blk_alloc_hints HeapChunkSelector::chunk_to_hints(chunk_num_t chunk_id) const {
     auto iter = m_chunks.find(chunk_id);
     if (iter == m_chunks.end()) {
-        LOGWARN("No chunk found for ChunkID {}, will return default blk alloc hints", chunk_id);
+        LOGWARNMOD(homeobject, "No chunk found for chunk_id {}, will return default blk alloc hints", chunk_id);
         return homestore::blk_alloc_hints();
     }
     homestore::blk_alloc_hints hints;
--- a/src/lib/homestore/heap_chunk_selector.cpp
+++ b/src/lib/homestore/heap_chunk_selector.cpp
@@ -117,4 +117,16 @@ void HeapChunkSelector::build_per_dev_chunk_heap(const std::unordered_set< chunk
         if (excludingChunks.find(p.first) == excludingChunks.end()) { add_chunk_internal(p.first); }
     };
 }
+
+homestore::blk_alloc_hints HeapChunkSelector::get_blk_alloc_hints_on_same_pdev(chunk_num_t chunk_id) const {
+    auto iter = m_chunks.find(chunk_id);
+    if (iter == m_chunks.end()) {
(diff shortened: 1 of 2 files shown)


xiaoxichen (approved):

src/lib/homestore/heap_chunk_selector.cpp
```diff
@@ -117,4 +117,16 @@ void HeapChunkSelector::build_per_dev_chunk_heap(const std::unordered_set< chunk
         if (excludingChunks.find(p.first) == excludingChunks.end()) { add_chunk_internal(p.first); }
     };
 }
-} // namespace homeobject
\ No newline at end of file
+
+homestore::blk_alloc_hints HeapChunkSelector::get_blk_alloc_hints_on_same_pdev(chunk_num_t chunk_id) const {
+    auto iter = m_chunks.find(chunk_id);
+    if (iter == m_chunks.end()) {
+        LOGWARN("No chunk found for ChunkID {}, will return default blk alloc hints", chunk_id);
```
szmyd, line 124: Should we use the `homeobject` logging module?

szmyd:
This makes sense to me, but will let @yamingk provide feedback if he has anything.

src/lib/homestore/heap_chunk_selector.cpp
```diff
@@ -117,4 +117,16 @@ void HeapChunkSelector::build_per_dev_chunk_heap(const std::unordered_set< chunk
         if (excludingChunks.find(p.first) == excludingChunks.end()) { add_chunk_internal(p.first); }
     };
 }
-} // namespace homeobject
\ No newline at end of file
+
+homestore::blk_alloc_hints HeapChunkSelector::get_blk_alloc_hints_on_same_pdev(chunk_num_t chunk_id) const {
```
yamingk, line 121: chunk_id can be const.

yamingk, line 121: nit: how about `chunk_to_hints`?

yamingk:
@zichanglai LGTM except some minor comment

yamingk:
@zichanglai just curious how do you consume this new interface. I am looking at PR-68, and I didn't see a place where we can provide hint to repldev async_alloc_write, or maybe I missed something.

src/lib/homestore/heap_chunk_selector.cpp
hkadayam: I believe we decided not to use `const` for parameter by value as it very verbose especially once you have more parameters and provide very little value if any.

szmyd: yes; i personally can not think of every a time where `const T` would have prevented some bug. (`const T&` obviously i have).

zichanglai: sounds good,  will update this

zichanglai:
> @zichanglai just curious how do we consume this new interface. I am looking at PR-68, and I didn't see a place where we can provide hint to repldev async_alloc_write, or maybe I missed something.

Hi @yamingk ,  this PR will be used in derived class of HS ReplDev Listener get_blk_alloc_hints() function in https://github.com/eBay/HomeStore/blob/319a3d5493a4813038fa7616c1823a26721ab060/src/include/homestore/replication/repl_dev.h#L73,  #PR 68 has not used this interface too as it will cause compile failure before it is merged.  I will add this interface call soon once PR #70 is merged.

yamingk (approved):

src/lib/homestore/heap_chunk_selector.cpp
zichanglai: yes, I just noticed this interface and had updated to use it.

xiaoxichen (approved):

PR #71, opened 2023-09-25
Integration with hs replication service and renamed certain path for in-mem and homestore backends

hkadayam:
* Removed all references of nuraft
* Integrated with homestore repl service and new homestore initialization
* enum and intrinsic type aliases is suffixed with _t
* PGManager creation and restarting with homestore
* Renamed some of the directories and files to avoid name confusion and better reflect the flow
* Disabled mocks


src/include/homeobject/common.hpp
```diff
@@ -25,11 +25,26 @@ constexpr uint64_t Gi = Ki * Mi;
...
+using cshared = const std::shared_ptr< T >;
+
+template < typename T >
+using unique = std::unique_ptr< T >;
+
+template < typename T >
+using intrusive = boost::intrusive_ptr< T >;
+
+template < typename T >
+using cintrusive = const boost::intrusive_ptr< T >;
```
szmyd, line 47: I don't think any of these `pointer` types are used in the API; can we move them to `homeobject_impl.hpp` (or create a new common_impl.hpp there?)

src/lib/homeobject_impl.hpp
```diff
@@ -32,11 +34,19 @@ struct Shard {
 using ShardList = std::list< Shard >;
 using ShardIterator = ShardList::iterator;
 
+struct PGReplDevBase {};
```
szmyd, line 37: ?

src/lib/homeobject_impl.hpp
```diff
@@ -32,11 +34,19 @@ struct Shard {
...
 struct PG {
-    explicit PG(PGInfo info) : pg_info(std::move(info)) {}
-    PGInfo pg_info;
-    uint64_t shard_sequence_num{0};
-    ShardList shards;
+    explicit PG(PGInfo info) : pg_info_(std::move(info)) {}
+    PG(PG const& pg) = delete;
+    PG(PG&& pg) = default;
+    PG& operator=(PG const& pg) = delete;
+    virtual ~PG() = default;
```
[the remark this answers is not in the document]
szmyd: nm...I see it.

src/lib/homeobject_impl.hpp
hkadayam: Tried something else, but forgot to remove. Thanks!

src/lib/homeobject_impl.hpp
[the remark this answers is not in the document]
hkadayam: because HS_PG is overridden and PG base class is used in common layer and we wanted to type cast them inside hs_homeobject

src/lib/homeobject_impl.hpp
```diff
@@ -46,33 +56,34 @@ class HomeObjectImpl : public HomeObject,
...
-    peer_id _our_id;
+    peer_id_t _our_id;
 
     /// Our SvcId retrieval and SvcId->IP mapping
     std::weak_ptr< HomeObjectApplication > _application;
 
     ///
     mutable std::shared_mutex _pg_lock;
-    std::map< pg_id, PG > _pg_map;
+    std::map< pg_id_t, std::shared_ptr< PG > > _pg_map;
```
szmyd, line 82: I'm not following why this has to be a shared_ptr? And shouldn't we use `shared` if we aliased it?

src/lib/homestore_backend/hs_pg_manager.cpp
```diff
@@ -0,0 +1,155 @@
...
+    return pg_info;
+}
+#endif
+
+void HSHomeObject::on_pg_meta_blk_found(sisl::byte_view const& buf, void* meta_cookie) {
+    homestore::superblk< pg_info_superblk > pg_sb;
+    pg_sb.load(buf, meta_cookie);
+
+    hs_repl_service()
+        .open_repl_dev(pg_sb->replica_set_uuid, std::make_unique< ReplicationStateMachine >(this))
```
szmyd, line 115: I'd like to review the cycle of lifetime expectations we're creating here between SM and this lower-layer. I'm not saying it's wrong, but it seems a candidate for incorrect shutdown-order in the future.

src/lib/homestore_backend/hs_shard_manager.cpp
```diff
@@ -1,34 +1,51 @@
...
 #include <homestore/meta_service.hpp>
 
+#include "hs_homeobject.hpp"
+#include "replication_message.hpp"
+#include "replication_state_machine.hpp"
+#include "lib/homeobject_impl.hpp"
+
 namespace homeobject {
 
+static constexpr uint32_t SEQUENCE_BIT_NUM_IN_SHARD{48};
```
szmyd, line 12: This is calculated in homeobject_impl.hpp so it can be dynamic.

src/lib/homestore_backend/hs_shard_manager.cpp
```diff
@@ -1,34 +1,51 @@
...
-shard_id HSHomeObject::generate_new_shard_id(pg_id pg) {
-    std::scoped_lock lock_guard(_pg_lock);
-    auto iter = _pg_map.find(pg);
-    RELEASE_ASSERT(iter != _pg_map.end(), "Missing pg info");
-    auto new_sequence_num = ++(iter->second.shard_sequence_num);
-    RELEASE_ASSERT(new_sequence_num < ShardManager::max_shard_num_in_pg(),
-                   "new shard id must be less than ShardManager::max_shard_num_in_pg()");
-    return make_new_shard_id(pg, new_sequence_num);
+shard_id_t HSHomeObject::make_new_shard_id(pg_id_t pg, uint64_t sequence_num) const {
+    return ((uint64_t)pg << SEQUENCE_BIT_NUM_IN_SHARD) | sequence_num;
```
szmyd, line 38: I think a lot of this was removed; maybe added back in a merge? It's shared by memory through homeobject_impl.hpp.

src/lib/homestore_backend/hs_shard_manager.cpp
```diff
@@ -87,20 +109,15 @@ ShardManager::Result< ShardInfo > HSHomeObject::_create_shard(pg_id pg_owner, ui
...
-    sisl::sg_list value;
     // replicate this create shard message to PG members;
-    auto [p, sf] = folly::makePromiseContract< ShardManager::Result< ShardInfo > >();
-    replica_set->write(sisl::blob(buf->data_begin(), needed_size), sisl::blob(), value, static_cast< void* >(&p));
-    auto info = std::move(sf).get();
-    if (!bool(info)) {
-        LOGWARN("create new shard [{}] on pg [{}] is failed with error:{}", new_shard_id & shard_mask, pg_owner,
-                info.error());
-    }
+    pg->repl_dev_->async_alloc_write(req->hdr_buf_, sisl::blob{}, sisl::sg_list{}, req);
```
xiaoxichen, line 113: The PGID/ShardID needed to be included into header. 
https://github.com/eBay/HomeObject/pull/68/files will do that.

src/include/homeobject/common.hpp
hkadayam: We certainly use it HO, but agree we can move this to homeobject_impl.hpp

hkadayam: But isn't it better be consistent across repo and I don't see anything wrong declaring here either, if homeobject chose to return on of these. So I am very much tempted to keep it here itself.

src/include/homeobject/shard_manager.hpp
```diff
@@ -8,24 +8,24 @@
 
 namespace homeobject {
 
-ENUM(ShardError, uint16_t, UNKNOWN = 1, TIMEOUT, INVALID_ARG, NOT_LEADER, UNKNOWN_PG, UNKNOWN_SHARD);
+ENUM(ShardError, uint16_t, UNKNOWN = 1, TIMEOUT, INVALID_ARG, NOT_LEADER, UNKNOWN_PG, UNKNOWN_SHARD, PG_NOT_READY);
```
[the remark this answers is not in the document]
hkadayam: Lets say PG has been created and we are recovering, but repl_dev has not found that or delayed in getting its replica set, technically UNKNOWN_PG is a wrong message. We need to differentiate to the caller that PG is found, but not ready to receive writes. May be we can have PG_UNAVAILABLE would be better, but thought PG_NOT_READY is also synonymous to that.

src/include/homeobject/common.hpp
szmyd: If we choose to return them we lift them here? I'm not sure I understand the use of providing a `homeobject::shared` to the consumer when they can use `std::shared_ptr` which is arguably shorter and more known?

src/lib/homeobject_impl.hpp
```diff
@@ -32,11 +34,19 @@ struct Shard {
...
+
 struct PG {
-    explicit PG(PGInfo info) : pg_info(std::move(info)) {}
-    PGInfo pg_info;
-    uint64_t shard_sequence_num{0};
-    ShardList shards;
+    explicit PG(PGInfo info) : pg_info_(std::move(info)) {}
+    PG(PG const& pg) = delete;
+    PG(PG&& pg) = default;
+    PG& operator=(PG const& pg) = delete;
```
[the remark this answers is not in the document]
szmyd: Why is this not movable? If all member fields are moveable; we should allow this to move by default too right? We don't have a custom destructor...

src/include/homeobject/common.hpp
hkadayam: Moved to homeobject_impl.hpp

src/lib/homestore_backend/hs_pg_manager.cpp
hkadayam: which one are referring to @szmyd?

src/lib/homeobject_impl.hpp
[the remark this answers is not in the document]
zichanglai: ok,  my bad, I read it wrong.

src/include/homeobject/shard_manager.hpp
```diff
@@ -8,24 +8,24 @@
...
 
 struct ShardInfo {
-    enum class State {
+    enum class State : uint8_t {
```
yamingk, line 14: nit: should we use ENUM/VENUM since we aliased?

yamingk (approved):

src/lib/homestore_backend/hs_pg_manager.cpp
hkadayam: After talking to Brian, we agreed that capturing shared_from_this can cause cyclic dependency, so leaving it as `this` as of now.

src/lib/homestore_backend/hs_pg_manager.cpp
```diff
@@ -0,0 +1,155 @@
...
+        .thenValue([this, pg_sb = std::move(pg_sb)](auto&& v) {
+            if (v.hasError()) {
```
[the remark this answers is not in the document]
hkadayam: Done

src/lib/homestore_backend/hs_shard_manager.cpp
hkadayam: Fixed

src/lib/homeobject_impl.hpp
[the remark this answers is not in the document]
hkadayam: I added move operator to be default.

src/lib/homestore_backend/hs_pg_manager.cpp
```diff
@@ -0,0 +1,155 @@
...
+[[maybe_unused]] static homestore::ReplDev& pg_repl_dev(PG const& pg) {
+    return *(static_cast< HSHomeObject::HS_PG const& >(pg).repl_dev_);
+}
+
+PGManager::NullAsyncResult HSHomeObject::_create_pg(PGInfo&& pg_info, std::set< std::string, std::less<> > peers) {
+    pg_info.replica_set_uuid = boost::uuids::random_generator()();
+    return hs_repl_service()
+        .create_repl_dev(pg_info.replica_set_uuid, std::move(peers), std::make_unique< ReplicationStateMachine >(this))
+        .thenValue([this, pg_info = std::move(pg_info)](auto&& v) -> PGManager::NullResult {
+            if (v.hasError()) { return folly::makeUnexpected(PGError::INVALID_ARG); }
```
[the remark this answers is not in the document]
hkadayam: Done

src/lib/homestore_backend/hs_homeobject.cpp
```diff
@@ -38,45 +40,46 @@ void HSHomeObject::init_homestore() {
...
-        [this]() { register_homestore_metablk_callback(); });
+    bool need_format = HomeStore::instance()
+                           ->with_index_service(nullptr)
+                           .with_repl_data_service(repl_impl_type::solo, std::make_shared< HeapChunkSelector >())
+                           .start(hs_input_params{.devices = device_info, .app_mem_size = app_mem_size},
+                                  [this]() { register_homestore_metablk_callback(); });
 
-    /// TODO how should this work?
-    LOGWARN("Persistence Looks Vacant, Formatting!!");
+    LOGWARN("Seems like we are booting/starting first time, Formatting!!");
```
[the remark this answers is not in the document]
hkadayam: Duh :-)

src/lib/homestore_backend/hs_homeobject.cpp
```diff
@@ -38,45 +40,46 @@ void HSHomeObject::init_homestore() {
...
+            {HS_SERVICE::INDEX, hs_format_params{.size_pct = 5.0}},
         });
     }
 }
 
 void HSHomeObject::register_homestore_metablk_callback() {
     // register some callbacks for metadata recovery;
     using namespace homestore;
-    HomeStore::instance()->meta_service().register_handler(
+    hs()->meta_service().register_handler(
```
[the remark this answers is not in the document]
hkadayam: done

src/lib/homeobject_impl.hpp
hkadayam: Here is the deal: I retained it as shared_ptr, because adding a base class into map and accessing them as derived class doesn't work with or without polymorphism. It is safe to add it as shared_ptr. We can obviously remove the entire concept of this, if we have different map (one for memory and other for homeobject). 

Another `TODO` is that we are taking global lock `_pg_lock` instead of per pg lock. I didn't change that part now, but we must change that.

yamingk (dismissed):

src/lib/homestore_backend/hs_pg_manager.cpp
```diff
@@ -0,0 +1,157 @@
...
+            }
+            add_pg_to_map(std::make_shared< HS_PG >(pg_sb, std::move(v.value())));
+        });
+}
+
+PGInfo HSHomeObject::HS_PG::pg_info_from_sb(homestore::superblk< pg_info_superblk > const& sb) {
+    PGInfo pginfo{sb->id};
+    for (uint32_t i{0}; i < sb->num_members; ++i) {
+        pginfo.members.emplace(sb->members[i].id, std::string(sb->members[i].name), sb->members[i].priority);
+    }
```
zichanglai, line 131: pginfo.replica_set_uuid = sb->replica_set_uuid; 
it is needed in my local env test when HO restarts,  otherwise L66 assert() will be failed.
it is a smaller change, it can be fixed in this pr or use another pr.

yamingk (approved):

PR #73, opened 2023-09-26
File backed HomeObject.

szmyd:
I had this in an old branch; useful maybe to someone or in our back-pocket.

Not-crash resilient and currently not-restartable.


yamingk:
it is definitely useful, maybe we can have a non-crash resilient, nor non-restartable version, and then keep building blocks up on it to make restartable and crash resilient. etc.

szmyd:
> Definitely useful, maybe we can have a non-crash resilient, nor non-restartable version, and then keep building blocks up on it to make restartable and crash resilient. etc.

Eventually arriving @ HomeStore? 😆 
If we can come up with a some other goal maybe...for now just was cleaning up my old branches. Feel free to improve upon and add restart code if you feel intrigued.

xiaoxichen:
Lol this is very interesting if we havent move the replication down to HS :)

PR #75, opened 2023-09-29
Make logging module consistent and remove atomic.

szmyd:


src/lib/homestore_backend/heap_chunk_selector.cpp
```diff
@@ -22,7 +22,7 @@ void HeapChunkSelector::add_chunk(csharedChunk& chunk) { m_chunks.emplace(VChunk
 void HeapChunkSelector::add_chunk_internal(const chunk_num_t chunkID) {
     if (m_chunks.find(chunkID) == m_chunks.end()) {
         // sanity check
-        LOGWARN("No chunk found for ChunkID {}", chunkID);
+        LOGWARNMOD(homeobject, "No chunk found for ChunkID {}", chunkID);
```
szmyd, line 25: HeapChunkSelector currently does not include homeobject headers...so this is left alone,

.clang-format
```diff
@@ -16,7 +16,6 @@ AlignConsecutiveDeclarations: false
 AlignEscapedNewlines: Right
 AlignOperands:   false
 AlignTrailingComments: true
-AllowShortBlocksOnASingleLine: true
```
szmyd, line 19: This was a dupe; it's below as well.

PR #77, opened 2023-10-02
Implement put and get apis

sanebay:
Create index table with key as BlobRoute(shard id, blobid) and value as MultiBlkId.
Do recovery of index table and connect with pg superblk ->index table uuid.
Do put with blob header and metadata. Allow request with a length from a particular offset.
Test create some pg's, shards and do put and get blob. Restart and then again verify with get with random offset and length.

### DIFF
--- a/src/include/homeobject/blob_manager.hpp
+++ b/src/include/homeobject/blob_manager.hpp
@@ -9,8 +9,8 @@
 
 namespace homeobject {
 
-ENUM(BlobError, uint16_t, UNKNOWN = 1, TIMEOUT, INVALID_ARG, NOT_LEADER, UNKNOWN_SHARD, UNKNOWN_BLOB,
-     CHECKSUM_MISMATCH);
+ENUM(BlobError, uint16_t, UNKNOWN = 1, TIMEOUT, INVALID_ARG, NOT_LEADER, UNKNOWN_SHARD, UNKNOWN_BLOB, CHECKSUM_MISMATCH,
+     READ_FAILED, INDEX_ERROR);
 
 struct Blob {
     Blob(sisl::io_blob_safe b, std::string const& u, uint64_t o) : body(std::move(b)), user_key(u), object_off(o) {}
@@ -18,8 +18,8 @@ struct Blob {
     Blob clone() const;
 
     sisl::io_blob_safe body;
-    std::string user_key;
-    uint64_t object_off;
+    std::string user_key{};
+    uint64_t object_off{};
     std::optional< peer_id_t > current_leader{std::nullopt};
 };
 
--- a/src/include/homeobject/common.hpp
+++ b/src/include/homeobject/common.hpp
@@ -11,7 +11,7 @@
 
 SISL_LOGGING_DECL(homeobject);
 
-#define HOMEOBJECT_LOG_MODS grpc_server, HOMESTORE_LOG_MODS, homeobject
+#define HOMEOBJECT_LOG_MODS grpc_server, HOMESTORE_LOG_MODS, homeobject, blobmgr
 
 #ifndef Ki
 constexpr uint64_t Ki = 1024ul;
--- a/src/lib/homestore_backend/hs_blob_manager.cpp
+++ b/src/lib/homestore_backend/hs_blob_manager.cpp
@@ -1,17 +1,308 @@
 #include "hs_homeobject.hpp"
+#include "replication_message.hpp"
+#include "replication_state_machine.hpp"
+#include "lib/homeobject_impl.hpp"
+
+SISL_LOGGING_DECL(blobmgr)
 
 namespace homeobject {
+static constexpr uint64_t io_align{512};
 
-BlobManager::Result< blob_id_t > HSHomeObject::_put_blob(ShardInfo const&, Blob&&) {
+BlobManager::AsyncResult< blob_id_t > HSHomeObject::_put_blob(ShardInfo const& shard, Blob&& blob) {
+    auto& pg_id = shard.placement_group;
+    shared< homestore::ReplDev > repl_dev;
+    {
+        std::shared_lock lock_guard(_pg_lock);
+        auto iter = _pg_map.find(pg_id);
+        RELEASE_ASSERT(iter != _pg_map.end(), "PG not found");
+        repl_dev = static_cast< HS_PG* >(iter->second.get())->repl_dev_;
+    }
+
+    RELEASE_ASSERT(repl_dev != nullptr, "Repl dev instance null");
+
+    const uint32_t needed_size = sizeof(ReplicationMessageHeader);
+    auto req = repl_result_ctx< BlobManager::Result< BlobInfo > >::make(needed_size, io_align);
+
+    uint8_t* raw_ptr = req->hdr_buf_.bytes;
+    ReplicationMessageHeader* header = new (raw_ptr) ReplicationMessageHeader();
+    header->msg_type = ReplicationMessageType::PUT_BLOB_MSG;
+    header->payload_size = 0;
+    header->payload_crc = 0;
+    header->shard_id = shard.id;
+    header->pg_id = pg_id;
+    header->header_crc = header->calculate_crc();
+
+    auto dev_block_size = repl_dev->get_blk_size();
+    bool blob_copied = false, user_key_copied = false;
+    sisl::sg_list sgs;
+    sgs.size = 0;
+
+    // Create blob header.
+    auto blob_header_size = sisl::round_up(sizeof(BlobHeader), io_align);
+    auto blob_header = r_cast< BlobHeader* >(iomanager.iobuf_alloc(io_align, blob_header_size));
+    blob_header->magic = BlobHeader::blob_header_magic;
+    blob_header->version = BlobHeader::blob_header_version;
+    blob_header->shard_id = shard.id;
+    blob_header->hash_algorithm = BlobHeader::HashAlgorithm::CRC32;
+    blob_header->blob_size = blob.body.size;
+    blob_header->user_key_size = blob.user_key.size();
+    sgs.iovs.emplace_back(iovec{.iov_base = blob_header, .iov_len = blob_header_size});
+    sgs.size += blob_header_size;
+
+    // Append blob bytes.
+    auto blob_bytes = blob.body.bytes;
+    auto blob_size = blob.body.size;
+    if ((reinterpret_cast< uintptr_t >(blob.body.bytes) % io_align != 0) || (blob_size % io_align != 0)) {
+        // If address or size is not aligned, align it and create a separate buffer
+        // and do expensive memcpy.
+        blob_size = sisl::round_up(blob_size, io_align);
+        blob_bytes = iomanager.iobuf_alloc(io_align, blob_size);
+        std::memcpy(blob_bytes, blob.body.bytes, blob.body.size);
+        blob_copied = true;
+    }
+
+    sgs.iovs.emplace_back(iovec{.iov_base = blob_bytes, .iov_len = blob_size});
+    sgs.size += blob_size;
+
+    // Append metadata if present and update the offsets and total size.
+    size_t user_key_size = 0;
+    uint8_t* user_key_bytes = nullptr;
+    if (!blob.user_key.empty()) {
+        user_key_size = blob.user_key.size();
+        user_key_bytes = r_cast< uint8_t* >(blob.user_key.data());
+        if ((reinterpret_cast< uintptr_t >(user_key_bytes) % io_align != 0) || (user_key_size % io_align != 0)) {
+            // If address or size is not aligned, create a separate buffer and do expensive memcpy.
+            user_key_size = sisl::round_up(user_key_size, io_align);
+            user_key_bytes = r_cast< uint8_t* >(iomanager.iobuf_alloc(io_align, user_key_size));
+            std::memcpy(user_key_bytes, blob.user_key.data(), blob.user_key.size());
+            user_key_copied = true;
+        }
+
+        sgs.iovs.emplace_back(iovec{.iov_base = user_key_bytes, .iov_len = user_key_size});
+        sgs.size += user_key_size;
+        // Set offset of user meta data is after blob bytes.
+        blob_header->user_key_offset = blob_size;
+    }
+
+    // Check if any padding of zeroes needs to be added to be aligned to device block size.
+    auto pad_len = sisl::round_up(sgs.size, dev_block_size) - sgs.size;
+    uint8_t* pad_zeroes = nullptr;
+    if (pad_len != 0) {
+        // TODO reuse a single pad zero buffer of len dev_block_size
+        pad_zeroes = r_cast< uint8_t* >(iomanager.iobuf_alloc(io_align, pad_len));
+        std::memset(pad_zeroes, 0, pad_len);
+        sgs.iovs.emplace_back(iovec{.iov_base = pad_zeroes, .iov_len = pad_len});
+        sgs.size += pad_len;
+    }
+
+    // Compute the checksum of blob and metadata.
+    compute_blob_payload_hash(blob_header->hash_algorithm, blob.body.bytes, blob.body.size, user_key_bytes,
+                              user_key_size, blob_header->hash, BlobHeader::blob_max_hash_len);
+
+    repl_dev->async_alloc_write(req->hdr_buf_, sisl::blob{}, sgs, req);
+    return req->result().deferValue([this, header, blob_header, blob = std::move(blob), blob_bytes, blob_copied,
+                                     user_key_bytes, user_key_copied,
+                                     pad_zeroes](const auto& result) -> BlobManager::AsyncResult< blob_id_t > {
+        header->~ReplicationMessageHeader();
+        iomanager.iobuf_free(r_cast< uint8_t* >(blob_header));
+        if (blob_copied) { iomanager.iobuf_free(blob_bytes); }
+        if (user_key_copied) { iomanager.iobuf_free(r_cast< uint8_t* >(user_key_bytes)); }
+        if (pad_zeroes) { iomanager.iobuf_free(r_cast< uint8_t* >(pad_zeroes)); }
+
+        if (result.hasError()) { return folly::makeUnexpected(result.error()); }
+        auto blob_info = result.value();
+        LOGTRACEMOD(blobmgr, "Put blob success shard {} blob {} pbas {}", blob_info.shard_id, blob_info.blob_id,
+                    blob_info.pbas.to_string());
+
+        return blob_info.blob_id;
+    });
+}
+
+void HSHomeObject::on_blob_put_commit(int64_t lsn, sisl::blob const& header, sisl::blob const& key,
+                                      const homestore::MultiBlkId& pbas,
+                                      cintrusive< homestore::repl_req_ctx >& hs_ctx) {
+    repl_result_ctx< BlobManager::Result< BlobInfo > >* ctx{nullptr};
+    if (hs_ctx != nullptr) {
+        ctx = boost::static_pointer_cast< repl_result_ctx< BlobManager::Result< BlobInfo > > >(hs_ctx).get();
+    }
+
+    auto msg_header = r_cast< ReplicationMessageHeader* >(header.bytes);
+    if (msg_header->corrupted()) {
+        LOGERROR("replication message header is corrupted with crc error, lsn:{}", lsn);
+        if (ctx) { ctx->promise_.setValue(folly::makeUnexpected(BlobError::CHECKSUM_MISMATCH)); }
+        return;
+    }
+
+    shared< BlobIndexTable > index_table;
+    {
+        std::shared_lock lock_guard(_pg_lock);
+        auto iter = _pg_map.find(msg_header->pg_id);
+        RELEASE_ASSERT(iter != _pg_map.end(), "PG not found");
+        index_table = static_cast< HS_PG* >(iter->second.get())->index_table_;
+        RELEASE_ASSERT(index_table != nullptr, "Index table not intialized");
+    }
+
+    BlobInfo blob_info;
+    blob_info.shard_id = msg_header->shard_id;
+    blob_info.blob_id = lsn;
+    blob_info.pbas = pbas;
+
+    // Write to index table with key {shard id, blob id } and value {pba}.
+    auto r = add_to_index_table(index_table, blob_info);
+    if (r.hasError()) {
+        LOGERROR("Failed to insert into index table for blob {} err {}", lsn, r.error());
+        ctx->promise_.setValue(folly::makeUnexpected(r.error()));
+        return;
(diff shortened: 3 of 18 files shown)


src/include/homeobject/blob_manager.hpp
```diff
@@ -4,13 +4,14 @@
 #include <string>
 
 #include <sisl/fds/buffer.hpp>
+#include <homestore/blk.h>
```
szmyd, line 7: We shouldn't need to add anything to this header except the ENUM added. Can all be moved into hs_homeobject.hpp right?

src/include/homeobject/pg_manager.hpp
```diff
@@ -32,6 +32,7 @@ struct PGInfo {
     pg_id_t id;
     mutable MemberSet members;
     peer_id_t replica_set_uuid;
+    homestore::uuid_t index_table_uuid;
```
szmyd, line 35: Should be part of PG; not PGInfo.

src/include/homeobject/common.hpp
```diff
@@ -11,7 +11,7 @@
 
 SISL_LOGGING_DECL(homeobject);
 
-#define HOMEOBJECT_LOG_MODS grpc_server, HOMESTORE_LOG_MODS, homeobject
+#define HOMEOBJECT_LOG_MODS grpc_server, HOMESTORE_LOG_MODS, homeobject, blobmgr
```
szmyd, line 14: Do we need another logging module; homeobject is not fine-grain enough?

src/lib/homestore_backend/index_kv.hpp
```diff
@@ -0,0 +1,96 @@
...
+#include <homestore/btree/btree_kv.hpp>
+#include <homestore/index/index_table.hpp>
+#include <homestore/index/index_internal.hpp>
+#include <homestore/index_service.hpp>
+#include <homestore/blk.h>
+#include "lib/blob_route.hpp"
+
+namespace homeobject {
+
+class BlobRouteKey : public homestore::BtreeKey {
```
szmyd, line 12: I think `BlobRouteKey` can _be_ both a `BtreeKey` and a `BlobRoute` so you can remove the `friend` yeah?

src/lib/homestore_backend/hs_blob_manager.cpp
```diff
@@ -1,17 +1,220 @@
...
+#include "lib/homeobject_impl.hpp"
+
+SISL_LOGGING_DECL(blobmgr)
 
 namespace homeobject {
 
-BlobManager::Result< blob_id_t > HSHomeObject::_put_blob(ShardInfo const&, Blob&&) {
-    return folly::makeUnexpected(BlobError::UNKNOWN);
+BlobManager::Result< blob_id_t > HSHomeObject::_put_blob(ShardInfo const& shard, Blob&& blob) {
+    auto pg_id = shard.placement_group;
```
szmyd, line 11: `auto&`? do we need a copy?

src/lib/homestore_backend/hs_blob_manager.cpp
```diff
@@ -1,17 +1,220 @@
...
+        std::shared_lock lg{_pg_lock};
+        if (auto it = _pg_map.find(pg_id); it != _pg_map.end()) {
+            pg = s_cast< HS_PG* >(it->second.get());
+        } else {
+            LOGWARN("failed to put blob with non-exist pg [{}]", pg_id);
+            return folly::makeUnexpected(BlobError::UNKNOWN_PG);
+        }
+    }
+
+    if (!pg->repl_dev_) {
```
szmyd, line 23: What is the scenario this can happen? Should we ASSERT instead?

src/lib/homestore_backend/hs_blob_manager.cpp
```diff
@@ -1,17 +1,220 @@
...
+    // Create blob header.
+    sgs.iovs.emplace_back(iovec{.iov_base = buf_header, .iov_len = size});
+    sgs.size += size;
+
+    // Append blob bytes.
+    sgs.iovs.emplace_back(iovec{.iov_base = blob.body.bytes, .iov_len = blob.body.size});
+    sgs.size += blob.body.size;
+
+    pg->repl_dev_->async_alloc_write(req->hdr_buf_, sisl::blob{}, sgs, req);
+    auto result = req->result().get();
```
szmyd, line 63: This should probably read something like:
```
retun req->result().deferValue([this] ...) -> Result<blob_id_t> {
   header->~ReplicationMessage..;
   ...
   return blob_info.blob_id;
  });
}
```
Otherwise this will block on the async call in the current thread which is not what we want (I think?).

src/lib/homestore_backend/hs_blob_manager.cpp
```diff
@@ -1,17 +1,220 @@
...
+
+    auto multi_blkids = r.value();
+    auto block_size = pg->repl_dev_->get_blk_size();
+    auto sgs_ptr = std::make_shared< sisl::sg_list >();
+    auto total_size = multi_blkids.blk_count() * block_size;
+    auto iov_base = iomanager.iobuf_alloc(block_size, total_size);
+    sgs_ptr->size = total_size;
+    sgs_ptr->iovs.emplace_back(iovec{.iov_base = iov_base, .iov_len = total_size});
+
+    auto result = pg->repl_dev_->async_read(multi_blkids, *sgs_ptr, total_size).get();
```
szmyd, line 144: Again, defer everything that follows this.

szmyd (changes requested):
Some initial questions and comments.

src/include/homeobject/pg_manager.hpp
szmyd: or at least hidden somehow from StorageManager.

src/include/homeobject/blob_manager.hpp
sanebay: Ack

src/include/homeobject/common.hpp
sanebay: Since blob put and get have high number of requests, so kept a separate logging trace. Metadata related api's like pg and shard and other can be under homeobject.

src/lib/homestore_backend/hs_blob_manager.cpp
sanebay: Ack

src/lib/homestore_backend/hs_blob_manager.cpp
sanebay: We get api requests while we are booting and pg recovery happened but not repl dev ?

src/lib/homestore_backend/hs_blob_manager.cpp
sanebay: Ack. same as above

src/include/homeobject/pg_manager.hpp
szmyd: Yes; but not memory or StorageManager. Let's look at this some more, it shouldn't need to be here and can be encapsulated below.

src/lib/homestore_backend/hs_blob_manager.cpp
szmyd: No, you cannot get a `shared_ptr<BlobManager>` or any other manager until `init` completes. So we can make this an assertion.

sanebay: Agree

src/include/homeobject/pg_manager.hpp
sanebay: Agree. Removed it from PGInfo.

xiaoxichen:
has this PR ready for review?

sanebay:
> has this PR ready for review?

Yes its ready for review.

src/lib/homestore_backend/index_kv.hpp
sanebay: Discussed with Brian. Both 'has a' or 'is a' is ok in this case.

src/lib/homestore_backend/hs_blob_manager.cpp
```diff
@@ -1,17 +1,271 @@
...
+    auto buf_header = iomanager.iobuf_alloc(block_size, size);
+    std::memcpy(buf_header, &blob_header, sizeof(blob_header));
+
+    // Create blob header.
+    sgs.iovs.emplace_back(iovec{.iov_base = buf_header, .iov_len = size});
+    sgs.size += size;
+
+    // Append blob bytes.
+    sgs.iovs.emplace_back(iovec{.iov_base = blob.body.bytes, .iov_len = blob.body.size});
+    sgs.size += blob.body.size;
```
zichanglai, line 56: sgs.size is needed to be HS datasvc block_size aligned, otherwise the alloc blk id will assert failed.
but there is no guarantee that blob.body.size is aligned with HS blk_size, so if I tried to write a blob with size = 10 in the test case,  repl_dev->async_alloc_write() will crash.

src/lib/homestore_backend/hs_blob_manager.cpp
```diff
@@ -1,17 +1,271 @@
...
+            // whole blob size else copy only the request length.
+            blob_bytes += req_offset;
+            auto res_len = req_len == 0 ? blob_size : req_len;
+            auto body = sisl::io_blob_safe(res_len);
+            std::memcpy(body.bytes, blob_bytes, res_len);
+
+            // Copy the metadata if its present.
+            std::string user_key{};
+            if (header->meta_data_offset != 0) {
+                blob_bytes += header->meta_data_offset;
```
zichanglai, line 204: if req_len is not zero and blob_bytes had been move forward in L196, so we need to rollback before move forward header->meta_data_offset, right?

src/lib/homestore_backend/hs_blob_manager.cpp
```diff
@@ -1,17 +1,271 @@
...
+        .thenValue([block_size, blob_id, req_len, req_offset, shard, multi_blkids, sgs_ptr,
+                    iov_base](auto&& result) mutable -> BlobManager::AsyncResult< Blob > {
+            if (result) {
+                LOGERROR("Failed to read blob {} shard {} err {}", blob_id, shard.id, result.value());
+                iomanager.iobuf_free(iov_base);
+                return folly::makeUnexpected(BlobError::READ_FAILED);
+            }
+
+            BlobHeader* header = (BlobHeader*)iov_base;
+            if (!header->valid() || header->shard_id != shard.id) {
```
xiaoxichen, line 171: maybe differentiate  `!header->valid()`  (data corruption) with `header->shard_id != shard.id` (metadata/index corruption)

src/lib/homestore_backend/hs_homeobject.hpp
```diff
@@ -81,6 +100,43 @@ class HSHomeObject : public HomeObjectImpl {
...
+            MD5 = 2,
+        };
+
+        uint64_t magic{blob_header_magic};
+        uint8_t version{blob_header_version};
+        HashAlgorithm hash_algorithm;
+        uint8_t hash[blob_max_hash_len]{};
+        shard_id_t shard_id;
+        int64_t meta_data_offset{}; // Metadata stored after the blob data.
+        int64_t total_size{};       // Size of data and metadata.
```
xiaoxichen, line 122: as we have implementation limit that sisl::io_blob_safe take uint32 as length as well as the `compute_blob_hash` takes `sisl::blob`  also capping to uint32,  maybe we can use uint32 for all the size and do a capping for put request? 

Though it is fair enough to cap the MAX_BLOB_SIZE to 1GB/2GB, no need to play at the boundary of uint32

src/lib/homestore_backend/hs_blob_manager.cpp
sanebay: Had a discussion on this with hari, yaming, zhichang. Plan is to use have SM align with block_size else we will have to do lot of alloc and copies of data which affects performance and cpu.

sanebay: We dont want to copy the data, so we can reuse the memory of body and user_key for iovec.

src/lib/homestore_backend/hs_blob_manager.cpp
```diff
@@ -1,17 +1,271 @@
...
+void HSHomeObject::on_blob_put_commit(int64_t lsn, sisl::blob const& header, sisl::blob const& key,
+                                      const homestore::MultiBlkId& pbas,
+                                      cintrusive< homestore::repl_req_ctx >& hs_ctx) {
+    repl_result_ctx< BlobManager::Result< BlobInfo > >* ctx{nullptr};
+    if (hs_ctx != nullptr) {
+        ctx = boost::static_pointer_cast< repl_result_ctx< BlobManager::Result< BlobInfo > > >(hs_ctx).get();
+    }
+
+    auto msg_header = r_cast< ReplicationMessageHeader* >(header.bytes);
+    if (msg_header->corrupted()) {
```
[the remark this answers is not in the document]
sanebay: The error is different for blob, pg and shard and we cant check in on_commit. Its safe to skip index update if any sort of memory corruption due to bug.

src/lib/homestore_backend/hs_blob_manager.cpp
sanebay: Ack

src/lib/homestore_backend/hs_blob_manager.cpp
sanebay: Yeah its an issue. Fixed.

src/lib/homestore_backend/hs_homeobject.hpp
sanebay: Ack

src/lib/homestore_backend/hs_blob_manager.cpp
xiaoxichen: How SM knows the block size of HS to do the align ?  How can SM do that w/o copy?

sanebay: If gateways divides blobs of size 16/32MB, then its aligned by 1k block size, then only the last blob will be unaligned. HO can align and do memcpy. But if there are lot of small blobs unaligned, lot of copies will happen.

xiaoxichen: GW for sure will divided blobs by a few MB (lets call it TARGET_BLOB_SIZE). 
However there will be a lot of small objects  whose size < `TARGET_BLOB_SIZE`, then they hit SM as is.

So SM/HO need to do alignment check and memcpy once if the blob is not align,  once and only once.  So I think better HO do that instead of SM.

src/lib/homestore_backend/hs_blob_manager.cpp
```diff
@@ -1,17 +1,271 @@
...
+    // Append blob bytes.
+    sgs.iovs.emplace_back(iovec{.iov_base = blob.body.bytes, .iov_len = blob.body.size});
+    sgs.size += blob.body.size;
+
+    // Append metadata and update the offsets and total size.
+    if (!blob.user_key.empty()) {
+        size_t user_key_size = blob.user_key.size();
+        sgs.iovs.emplace_back(iovec{.iov_base = blob.user_key.data(), .iov_len = user_key_size});
+        sgs.size += user_key_size;
+        blob_header.total_size += user_key_size;
```
JacksonYao287, line 63: the total size in header is updated here, but it was copied to the buffer at line48 with a stale `total_size`.

src/lib/homestore_backend/hs_blob_manager.cpp
```diff
@@ -1,17 +1,271 @@
...
+                iomanager.iobuf_free(iov_base);
+                return folly::makeUnexpected(BlobError::INVALID_ARG);
+            }
+
+            // Copy the blob bytes from the offset. If request len is 0, take the
+            // whole blob size else copy only the request length.
+            blob_bytes += req_offset;
+            auto res_len = req_len == 0 ? blob_size : req_len;
+            auto body = sisl::io_blob_safe(res_len);
+            std::memcpy(body.bytes, blob_bytes, res_len);
```
JacksonYao287, line 199: will there be a user case that req_len == 0 and req_offset != 0?
i know this is a wrong case , but it might happens in some case . so i think we should check this before `memcpy` , otherwise, segment fault will happen

src/lib/homestore_backend/hs_blob_manager.cpp
```diff
@@ -1,17 +1,271 @@
...
+        repl_dev = static_cast< HS_PG* >(iter->second.get())->repl_dev_;
+    }
+
+    RELEASE_ASSERT(repl_dev != nullptr, "Repl dev instance null");
+
+    const uint32_t needed_size = sizeof(ReplicationMessageHeader);
+    auto req = repl_result_ctx< BlobManager::Result< BlobInfo > >::make(needed_size, 512);
+
+    uint8_t* raw_ptr = req->hdr_buf_.bytes;
+    ReplicationMessageHeader* header = new (raw_ptr) ReplicationMessageHeader();
```
zichanglai, line 29: in the repl_result_ctx struct,  a ReplicationMessageHeader is already been defined, can we reuse that to avoid re-create a new one? if we do that, req->hdr_buf_ can be used to hold the(blob_header, body and user_key),the benefits is when req is released, buf will be automatically released too.

src/lib/homestore_backend/hs_blob_manager.cpp
xiaoxichen: nice catch

src/lib/homestore_backend/hs_blob_manager.cpp
xiaoxichen: just make L197 to 
```
auto res_len = req_len == 0 ? blob_size-req_offset : req_len;
```

src/lib/homestore_backend/index_kv.hpp
```diff
@@ -0,0 +1,96 @@
...
+private:
+    BlobRoute key_;
```
hkadayam, line 14: Please pack the BlobRoute structure internally, since it gets directly serialized.

src/lib/homestore_backend/hs_blob_manager.cpp
sanebay: removed the total_size. With alignment we need blob size and user meta size in the header.

src/lib/homestore_backend/hs_blob_manager.cpp
sanebay: Ack

src/lib/homestore_backend/hs_blob_manager.cpp
sanebay: Yeah we could do the first one. But what will we pass in needed_size in that case ? 0 to malloc inside blob ? . I dont prefer option 2 because it will difficult to understand the code. we are anyway trying to reuse the body and user key we got from SM if its aligned and keeping in sg vec.

src/lib/homestore_backend/hs_blob_manager.cpp
sanebay: Yes we need to align the sgs.size to be 1k block size. So we are adding padding now after the metadata. Each iovec also needs to have address and size aligned 512 bytes due to DIRECT IO.

xiaoxichen (dismissed):

JacksonYao287 (commented):
Generally looks good, a few comments left

src/lib/homestore_backend/hs_blob_manager.cpp
```diff
@@ -1,17 +1,325 @@
...
```
JacksonYao287, line 30: in line 109， only the destructor of `ReplicationMessageHeader` is called, when will this allocated buffer be released？

src/lib/homestore_backend/hs_blob_manager.cpp
```diff
@@ -1,17 +1,325 @@
...
+            LOGERROR("failed to put blob with non-exist pg [{}]", pg_id);
+            return folly::makeUnexpected(BlobError::UNKNOWN_PG);
+        }
+        repl_dev = static_cast< HS_PG* >(iter->second.get())->repl_dev_;
+    }
+
+    RELEASE_ASSERT(repl_dev != nullptr, "Repl dev instance null");
+
+    const uint32_t needed_size = sizeof(ReplicationMessageHeader);
+    auto req = repl_result_ctx< BlobManager::Result< BlobInfo > >::make(needed_size, 512);
```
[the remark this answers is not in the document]
sanebay: Ack

JacksonYao287 (approved):
LGTM!

src/include/homeobject/blob_manager.hpp
```diff
@@ -9,17 +9,17 @@
 
 namespace homeobject {
 
-ENUM(BlobError, uint16_t, UNKNOWN = 1, TIMEOUT, INVALID_ARG, NOT_LEADER, UNKNOWN_SHARD, UNKNOWN_BLOB,
-     CHECKSUM_MISMATCH);
+ENUM(BlobError, uint16_t, UNKNOWN = 1, TIMEOUT, INVALID_ARG, NOT_LEADER, UNKNOWN_SHARD, UNKNOWN_BLOB, CHECKSUM_MISMATCH,
+     UNKNOWN_PG, PG_NOT_READY, READ_FAILED, INDEX_ERROR);
```
szmyd, line 13: Remove `PG_NOT_READY`; it's unused but will have to be handled in any `switch` statement on this type.

szmyd (changes requested):
one tweak regarding public API.

src/lib/homestore_backend/hs_blob_manager.cpp
```diff
@@ -1,17 +1,320 @@
...
+        return;
+    }
+
+    shared< BlobIndexTable > index_table;
+    {
+        std::shared_lock lock_guard(_pg_lock);
+        auto iter = _pg_map.find(msg_header->pg_id);
+        if (iter == _pg_map.end()) {
+            LOGERROR("Couldnt find pg {} for blob {}", msg_header->pg_id, lsn);
+            ctx->promise_.setValue(folly::makeUnexpected(BlobError::UNKNOWN_PG));
```
szmyd, line 145: Is this another case of "How can this happen" probably an assert?

src/lib/homestore_backend/hs_blob_manager.cpp
```diff
@@ -1,17 +1,320 @@
...
+                                                         uint64_t req_len) const {
+    auto& pg_id = shard.placement_group;
+    shared< BlobIndexTable > index_table;
+    shared< homestore::ReplDev > repl_dev;
+    {
+        std::shared_lock lock_guard(_pg_lock);
+        auto iter = _pg_map.find(pg_id);
+        if (iter == _pg_map.end()) {
+            LOGERROR("failed to do get blob with non-exist pg [{}]", pg_id);
+            return folly::makeUnexpected(BlobError::UNKNOWN_PG);
```
szmyd, line 179: ditto.

szmyd (approved):
:+1:

PR #81, opened 2023-10-16
Add seal shard api

sanebay:
Similar to create shard. 
Add test to create shard, seal and put blob and restart and verify shard is still sealed and put blob fails.

### DIFF
--- a/src/lib/homestore_backend/hs_shard_manager.cpp
+++ b/src/lib/homestore_backend/hs_shard_manager.cpp
@@ -102,8 +102,54 @@ ShardManager::AsyncResult< ShardInfo > HSHomeObject::_create_shard(pg_id_t pg_ow
     });
 }
 
-ShardManager::Result< ShardInfo > HSHomeObject::_seal_shard(shard_id_t id) {
-    return folly::makeUnexpected(ShardError::UNKNOWN_SHARD);
+ShardManager::AsyncResult< ShardInfo > HSHomeObject::_seal_shard(shard_id_t shard_id) {
+    pg_id_t pg_id;
+    ShardInfo shard_info;
+    {
+        std::scoped_lock lock_guard(_shard_lock);
+        auto shard_iter = _shard_map.find(shard_id);
+        RELEASE_ASSERT(shard_iter != _shard_map.end(), "shard id not found");
+        auto hs_shard = d_cast< HS_Shard* >((*shard_iter->second).get());
+        shard_info = hs_shard->info;
+        pg_id = hs_shard->info.placement_group;
+    }
+
+    shared< homestore::ReplDev > repl_dev;
+    {
+        std::shared_lock lock_guard(_pg_lock);
+        auto iter = _pg_map.find(pg_id);
+        RELEASE_ASSERT(iter != _pg_map.end(), "PG not found");
+        repl_dev = static_cast< HS_PG* >(iter->second.get())->repl_dev_;
+        RELEASE_ASSERT(repl_dev != nullptr, "Repl dev null");
+    }
+
+    shard_info.state = ShardInfo::State::SEALED;
+    auto seal_shard_message = serialize_shard_info(shard_info);
+    const auto msg_size = sisl::round_up(seal_shard_message.size(), repl_dev->get_blk_size());
+    auto req = repl_result_ctx< ShardManager::Result< ShardInfo > >::make(msg_size, 512 /*alignment*/);
+    auto buf_ptr = req->hdr_buf_.bytes;
+    std::memset(buf_ptr, 0, msg_size);
+    std::memcpy(buf_ptr, seal_shard_message.c_str(), seal_shard_message.size());
+
+    req->header_.msg_type = ReplicationMessageType::SEAL_SHARD_MSG;
+    req->header_.pg_id = pg_id;
+    req->header_.shard_id = shard_id;
+    req->header_.payload_size = msg_size;
+    req->header_.payload_crc = crc32_ieee(init_crc32, buf_ptr, msg_size);
+    req->header_.seal();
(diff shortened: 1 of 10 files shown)


zichanglai:
lgtm，it will be more clear once put/get blob pr is merged and rebase on that.

sanebay:
> lgtm，it will be more clear once put/get blob pr is merged and rebase on that.

Rebased to latest. Please review.

src/lib/homestore_backend/tests/test_home_object.cpp
```diff
@@ -278,6 +278,60 @@ TEST_F(HomeObjectFixture, BasicPutGetBlob) {
...
+    EXPECT_EQ(shard_info.id, shard_id);
+    EXPECT_EQ(shard_info.placement_group, pg_id);
+    EXPECT_EQ(shard_info.state, ShardInfo::State::SEALED);
+
+    b = _obj_inst->blob_manager()->put(shard_id, Blob{sisl::io_blob_safe(512u, 512u), "test_blob", 0ul}).get();
+    ASSERT_TRUE(!b);
+    ASSERT_EQ(b.error(), BlobError::SEALED_SHARD);
+    LOGINFO("Put blob {}", b.error());
+}
+
```
szmyd, line 334: Are these different than the basic tests already in `lib/tests/ShardManagerTest.cpp` ? We should be using it for non-HS specific tests like these probably.

src/lib/homestore_backend/hs_shard_manager.cpp
```diff
@@ -102,8 +102,54 @@ ShardManager::AsyncResult< ShardInfo > HSHomeObject::_create_shard(pg_id_t pg_ow
...
+    pg_id_t pg_id;
+    ShardInfo shard_info;
+    {
+        std::scoped_lock lock_guard(_shard_lock);
+        auto shard_iter = _shard_map.find(shard_id);
+        RELEASE_ASSERT(shard_iter != _shard_map.end(), "shard id not found");
+        auto hs_shard = d_cast< HS_Shard* >((*shard_iter->second).get());
+        shard_info = hs_shard->info;
+        pg_id = hs_shard->info.placement_group;
+    }
```
szmyd, line 115: We already called `get_shard` in `lib/shard_manager.cpp`; we should use its result not take the lock all over again. Probably means taking a `shard_it` rather than a `shard_id_t`.

src/lib/homestore_backend/hs_shard_manager.cpp
```diff
@@ -102,8 +102,54 @@ ShardManager::AsyncResult< ShardInfo > HSHomeObject::_create_shard(pg_id_t pg_ow
...
+    sisl::blob header;
+    header.bytes = r_cast< uint8_t* >(&req->header_);
+    header.size = sizeof(req->header_);
+    sisl::sg_list value;
+    value.size = msg_size;
+    value.iovs.push_back(iovec(buf_ptr, msg_size));
+
+    // replicate this seal shard message to PG members;
+    repl_dev->async_alloc_write(header, sisl::blob{}, value, req);
+    return req->result().deferValue([this](auto const& e) -> ShardManager::Result< ShardInfo > {
```
szmyd, line 149: Can we just `return req->result()`?

src/lib/homestore_backend/hs_shard_manager.cpp
sanebay: Will change the internal API to take a const ShardInfo&

src/lib/homestore_backend/hs_shard_manager.cpp
sanebay: Yeah thats nice.

szmyd (approved):
lgtm

src/lib/homestore_backend/hs_shard_manager.cpp
```diff
@@ -102,8 +102,44 @@ ShardManager::AsyncResult< ShardInfo > HSHomeObject::_create_shard(pg_id_t pg_ow
     });
 }
 
-ShardManager::Result< ShardInfo > HSHomeObject::_seal_shard(shard_id_t id) {
-    return folly::makeUnexpected(ShardError::UNKNOWN_SHARD);
+ShardManager::AsyncResult< ShardInfo > HSHomeObject::_seal_shard(ShardInfo const& info) {
```
yamingk, line 105: Once a shard is sealed, the chunk backing this shard should be returned to heapChunkSelector, right?
Can you point me who is doing this part?

src/lib/homestore_backend/hs_shard_manager.cpp
```diff
@@ -102,8 +102,44 @@ ShardManager::AsyncResult< ShardInfo > HSHomeObject::_create_shard(pg_id_t pg_ow
...
+    req->header_.seal();
+    sisl::blob header;
+    header.bytes = r_cast< uint8_t* >(&req->header_);
+    header.size = sizeof(req->header_);
+    sisl::sg_list value;
+    value.size = msg_size;
+    value.iovs.push_back(iovec(buf_ptr, msg_size));
+
+    // replicate this seal shard message to PG members;
+    repl_dev->async_alloc_write(header, sisl::blob{}, value, req);
```
yamingk, line 141: We also need to update the shard's metablk from active to sealed state, right?

sanebay: It was already there in hs_shard_manager.cpp on commit

src/lib/homestore_backend/hs_shard_manager.cpp
sanebay: Yes this needs to be added.

zichanglai (approved):

PR #86, opened 2023-10-18
Implement blob del API

JacksonYao287:
referring the discussion [here](https://ebay-eng.slack.com/archives/CAVQXEAKS/p1698161973016819), we do not need tombstone, and just remove the blob from btree when it is deleted

### DIFF
--- a/src/include/homeobject/blob_manager.hpp
+++ b/src/include/homeobject/blob_manager.hpp
@@ -10,7 +10,7 @@
 namespace homeobject {
 
 ENUM(BlobError, uint16_t, UNKNOWN = 1, TIMEOUT, INVALID_ARG, NOT_LEADER, UNKNOWN_SHARD, UNKNOWN_BLOB, CHECKSUM_MISMATCH,
-     READ_FAILED, INDEX_ERROR, SEALED_SHARD);
+     READ_FAILED, INDEX_ERROR, SEALED_SHARD, TOMBSTONE_BLOB, ALIVE);
 
 struct Blob {
     Blob(sisl::io_blob_safe b, std::string const& u, uint64_t o) : body(std::move(b)), user_key(u), object_off(o) {}
--- a/src/include/homeobject/blob_manager.hpp
+++ b/src/include/homeobject/blob_manager.hpp
@@ -10,7 +10,7 @@
 namespace homeobject {
 
 ENUM(BlobError, uint16_t, UNKNOWN = 1, TIMEOUT, INVALID_ARG, NOT_LEADER, UNKNOWN_SHARD, UNKNOWN_BLOB, CHECKSUM_MISMATCH,
-     READ_FAILED, INDEX_ERROR, SEALED_SHARD, TOMBSTONE_BLOB, ALIVE);
+     READ_FAILED, INDEX_ERROR, SEALED_SHARD);
 
 struct Blob {
     Blob(sisl::io_blob_safe b, std::string const& u, uint64_t o) : body(std::move(b)), user_key(u), object_off(o) {}
--- a/src/lib/homestore_backend/hs_blob_manager.cpp
+++ b/src/lib/homestore_backend/hs_blob_manager.cpp
@@ -351,6 +351,7 @@ void HSHomeObject::on_blob_del_commit(int64_t lsn, sisl::blob const& header, sis
     auto r = delete_from_index_table(index_table, blob_info.shard_id, blob_info.blob_id);
     if (!r) {
         LOGW("fail to delete blob id {} shard {}", blob_info.blob_id, blob_info.shard_id);
+        if (ctx) ctx->promise_.setValue(folly::makeUnexpected(r.error()));
         return;
     }
 
--- a/src/lib/homestore_backend/hs_blob_manager.cpp
+++ b/src/lib/homestore_backend/hs_blob_manager.cpp
@@ -282,8 +282,81 @@ homestore::blk_alloc_hints HSHomeObject::blob_put_get_blk_alloc_hints(sisl::blob
     return hints;
 }
 
-BlobManager::NullAsyncResult HSHomeObject::_del_blob(ShardInfo const&, blob_id_t) {
-    return folly::makeUnexpected(BlobError::UNKNOWN);
+BlobManager::NullAsyncResult HSHomeObject::_del_blob(ShardInfo const& shard, blob_id_t blob_id) {
+    auto& pg_id = shard.placement_group;
+    shared< homestore::ReplDev > repl_dev;
+    {
+        std::shared_lock lock_guard(_pg_lock);
+        auto iter = _pg_map.find(pg_id);
+        RELEASE_ASSERT(iter != _pg_map.end(), "PG not found");
+        repl_dev = static_cast< HS_PG* >(iter->second.get())->repl_dev_;
+    }
+
+    RELEASE_ASSERT(repl_dev != nullptr, "Repl dev instance null");
+
+    auto req = repl_result_ctx< BlobManager::Result< BlobInfo > >::make(sizeof(blob_id_t), io_align);
+
+    req->header_.msg_type = ReplicationMessageType::DEL_BLOB_MSG;
+    req->header_.payload_size = 0;
+    req->header_.payload_crc = 0;
+    req->header_.shard_id = shard.id;
+    req->header_.pg_id = pg_id;
+    req->header_.seal();
+    sisl::blob header;
+    header.bytes = r_cast< uint8_t* >(&req->header_);
+    header.size = sizeof(req->header_);
+
+    memcpy(req->hdr_buf_.bytes, &blob_id, sizeof(blob_id_t));
+
+    repl_dev->async_alloc_write(header, req->hdr_buf_, sisl::sg_list{}, req);
+    return req->result().deferValue([](const auto& result) -> folly::Expected< folly::Unit, BlobError > {
+        if (result.hasError()) { return folly::makeUnexpected(result.error()); }
+        auto blob_info = result.value();
+        LOGTRACEMOD(blobmgr, "Delete blob success,  shard_id {} , blob_id {}", blob_info.shard_id, blob_info.blob_id);
+
+        return folly::Unit();
+    });
+}
+
+void HSHomeObject::on_blob_del_commit(int64_t lsn, sisl::blob const& header, sisl::blob const& key,
+                                      cintrusive< homestore::repl_req_ctx >& hs_ctx) {
+    repl_result_ctx< BlobManager::Result< BlobInfo > >* ctx{nullptr};
+    if (hs_ctx != nullptr) {
+        ctx = boost::static_pointer_cast< repl_result_ctx< BlobManager::Result< BlobInfo > > >(hs_ctx).get();
+    }
+
+    auto msg_header = r_cast< ReplicationMessageHeader* >(header.bytes);
+    if (msg_header->corrupted()) {
+        LOGERROR("replication message header is corrupted with crc error, lsn:{}", lsn);
+        if (ctx) { ctx->promise_.setValue(folly::makeUnexpected(BlobError::CHECKSUM_MISMATCH)); }
+        return;
+    }
+
+    shared< BlobIndexTable > index_table;
+    shared< homestore::ReplDev > repl_dev;
+    {
+        std::shared_lock lock_guard(_pg_lock);
+        auto iter = _pg_map.find(msg_header->pg_id);
+        RELEASE_ASSERT(iter != _pg_map.end(), "PG not found");
+        index_table = static_cast< HS_PG* >(iter->second.get())->index_table_;
+        repl_dev = static_cast< HS_PG* >(iter->second.get())->repl_dev_;
+        RELEASE_ASSERT(index_table != nullptr, "Index table not intialized");
+        RELEASE_ASSERT(repl_dev != nullptr, "Repl dev instance null");
+    }
+
+    BlobInfo blob_info;
+    blob_info.shard_id = msg_header->shard_id;
+    blob_info.blob_id = *r_cast< blob_id_t* >(key.bytes);
+
+    auto r = delete_from_index_table(index_table, blob_info.shard_id, blob_info.blob_id);
+    if (!r) {
+        LOGW("fail to delete blob id {} shard {}", blob_info.blob_id, blob_info.shard_id);
+        return;
+    }
+
+    auto& multiBlks = r.value();
+    repl_dev->async_free_blks(lsn, multiBlks);
+    if (ctx) { ctx->promise_.setValue(BlobManager::Result< BlobInfo >(blob_info)); }
 }
 
 void HSHomeObject::compute_blob_payload_hash(BlobHeader::HashAlgorithm algorithm, const uint8_t* blob_bytes,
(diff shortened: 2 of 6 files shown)


yamingk:
A general question: shouldn't HomeObject issue an async_free_blobk for the blob being deleted to replDev?

JacksonYao287:
> A general question: shouldn't HomeObject issue an async_free_blobk for the blob being deleted to replDev?

GC Thread should scan all the tombstone in indexService, read the multiblks of these deleted blobs and then
1 schedule `async_free_blk` to replDev
2 remove tombstone items from indexservice

for del API , it will only mark the deleted bolbs to tombstone. it will not free blks at all
@yamingk

xiaoxichen:
the async_free_blk does nothing but just book-keeping m_deleted_blks,   I think we need to call the async_free_blocks to update the metrics, otherwise GC cannot pick.

src/lib/blob_route.hpp
```diff
@@ -12,12 +12,12 @@ namespace homeobject {
 // to appear in a different Index should the Blob (Shard) be moved between Pgs.
 #pragma pack(1)
 struct BlobRoute {
+    enum KeyType { kTypeTombStone = 0x0, kTypeNormal = 0x1 };
+    KeyType keyType;
```
zichanglai, line 16: I will suggest to move keyType to behind of blob and use shard_id as the highest bits, as we need to range query all blobs for one shard.

szmyd: Should the Route be where we store the tombstone and not an extension of the Index value (as the memory_backend is doing it)?

yamingk:
@JacksonYao287 yes, you are right about the ask.

src/lib/blob_route.hpp
JacksonYao287: actually, we want to do a rangequery  to find out all the tomestone blob in a given shard. 
in a common KV DB(for example, leveldb), we should provide the key prefix, which should be `shardID+tombstone` in our use case to get the best efficiency , just like @zichanglai said.

so , i prefer to in put the tombstone after shardid in Route for now, since this will get a better efficiency than an extension of the Index value

src/lib/homestore_backend/index_kv.cpp
```diff
@@ -59,9 +59,98 @@ BlobManager::NullResult HSHomeObject::add_to_index_table(shared< BlobIndexTable
...
+    auto status = index_table->get(get_req);
+    if (status != homestore::btree_status_t::success) {
+        LOGERROR("Failed to get from index table {}", index_key.to_string());
+        return folly::makeUnexpected(BlobError::UNKNOWN_BLOB);
+    }
+
+    index_key.change_to_tombstone();
+    // we set a tombstone in key, of which GC can take the benifit to quickly finding the delete blkIDs by using
+    // the range query of key
+    homestore::BtreeSinglePutRequest put_req{&index_key, &index_value,
```
xiaoxichen, line 80: As we already called blkallocator to free the PBAs,  technically the PBA is not belongs to HomeObj anymore.  IMO no point to store the PBA in index_value of tombstone,   the raft LSN makes more sense by providing more info.

src/lib/homestore_backend/index_kv.cpp
```diff
@@ -59,9 +59,98 @@ BlobManager::NullResult HSHomeObject::add_to_index_table(shared< BlobIndexTable
...
+    BlobRouteValue index_value;
+    homestore::BtreeSingleGetRequest get_req{&index_key, &index_value};
+    auto status = index_table->get(get_req);
+    if (status != homestore::btree_status_t::success) {
+        LOGERROR("Failed to get from index table {}", index_key.to_string());
+        return folly::makeUnexpected(BlobError::UNKNOWN_BLOB);
+    }
+
+    index_key.change_to_tombstone();
+    // we set a tombstone in key, of which GC can take the benifit to quickly finding the delete blkIDs by using
```
xiaoxichen, line 78: I fully get the point of this approach, compared to update the value not the key, it provide a fast way to get active/deleted blobs set of a shard.

However I am not convinced by the tradeoff,  with this approach we takes 4(include journal)+( + in the case that append/delete requires b-tree splits) index update compared to 2 if we only change the value, for every blob delete.  
The gain is saving the (maybe one time, only when GC) cost of filtering out deleted blob(identify from index value) in memory, which looks to me like cheaper

src/lib/homestore_backend/index_kv.cpp
JacksonYao287: now, when we free pbas,  appendBlockAllocator only update `m_freeable_nblks` (sometimes , `m_last_append_offset` also), it does not store which pbas are freeed, so we have to store the pbas of tombstone for GC thread to determine which pbas should not be copied to a new chunk.
well, if appendBlockAllocator stores those freeed pbas some day, we can simplifiy the code here.

do not quite understand what infomation can LSN provide. can you explain more deeper?

src/lib/homestore_backend/index_kv.cpp
JacksonYao287: I do not think it is acceptable that we have to scan the whole table and compare the value one by one when GC any shard, especially when the index table is very large . 

i get your point. only updating the value will not change the btree structure and will schedule less index update operations. 

we can discuss this tradeoff offline with more detail.

src/lib/blob_route.hpp
sanebay: How do we know all blobs are deleted as each is independen operation to create a shardID + tombstone key ? Having a key prefix for a blob or shard might help to find if its deleted or not but dont we need to check it for put to know if its deleted or not ? Couple of options
a. Use prefix key tombstone + shard id + blob id in same tree. Cons get needs to check two keys actual and tombstone key.
b. Use a different tree to store all deleted keys (1 active and 1 pending) - Update the value in current index btree to mark it as deleted and insert into active delete tree. GC only needs to look at pending delete tree which is very small compared to the actual index btree. double buffering of active and pending tree.

src/lib/homestore_backend/index_kv.hpp
```diff
@@ -52,6 +52,9 @@ class BlobRouteKey : public homestore::BtreeKey {
         return os;
     }
 
+    void change_to_tombstone() { key_.keyType = BlobRoute::KeyType::kTypeTombStone; }
```
yamingk, line 55: nit: `to_tombstone()`, `to_normal()`

src/lib/homestore_backend/hs_blob_manager.cpp
```diff
@@ -281,8 +281,84 @@ homestore::blk_alloc_hints HSHomeObject::blob_put_get_blk_alloc_hints(sisl::blob
...
+    // set a tombstone for the blob in index table.
+    // it is the GC thread's responsibility to delete the tombstone from index table and reclaim the space.
+    auto r = move_to_tombstone(index_table, blob_info);
+    if (!r) {
+        LOGW("TombStone blob not found in index id {} shard {}", blob_info.blob_id, blob_info.shard_id);
+        return;
+    }
+
+    auto& tombStone_multiBlks = r.value();
+    repl_dev->async_free_blks(lsn, tombStone_multiBlks);
```
yamingk, line 359: as an integration test, can we also verify in the test case that the space is actually freed, e.g. the m_total_freeable is increasing? Just to make sure the call made all the way to underlying chunk and its allocator

src/lib/homestore_backend/index_kv.cpp
zichanglai: in current GC design flow,  when key is marked with tombstone, the value is no used as we will only filter-out the alive index key/value from the index table and only move that parts of data. we do not need to know the pba of one key with tombstone as will not touch that data anymore.

yamingk: no need to store the deleted pba, GC won't consume that pba, but only the state of that blob, if it was tomestoned, skip.

src/lib/homestore_backend/hs_blob_manager.cpp
JacksonYao287: ack

szmyd: We should avoid testing code from other components, those tests should live there yeah, especially if it results in adding a public API *just* to test HS internal logic.

xiaoxichen:
I am with strong opinion to have tombstone.   Trying to avoid a side dicussion here but lets discussion in the slack thread

src/lib/homestore_backend/hs_blob_manager.cpp
yamingk: I agree on the point to avoid adding public API for just testing other component. 
I don't agree on this testing other components. The various underlying component already have their UTs to cover functional. When it is being consumed by external components, we need some level of integration test to verify the the functionality (not in lower component, but in this component this feature is being added) is actually working, right? 

We also need integration test as a full HO functionalities that HO is providing, we can't just say HO supports this, that but only before HomeStore.

szmyd: Simply, if the bug for some issue is found in repo A (or some code is not being tested) that test should live in repo A; not in dependent repo B. I'm not aware of the scenario you're trying to ensure here, it sounds like a validation for the test, but it should not be added in order to validate HomeStore is working correctly, merely that HomeObject's assumptions about the HomeStore API is valid.
In this case the value is not made available in the API (cause it's not normally useful apparently), so we don't need to add a validation here as no API invariants are called into question. If the API did already exist, i agree it's an additional validation needed since the API is what we're testing.

szmyd: Another way to put this is to treat HomeObject is only knowing that HomeStore is an API. Imagine someone implemented a version that had VChunks that did not release storage immediately upon delete (for whatever reason), now this validation is incorrect as the decreasing of this value is not part of the API contract itself.

src/include/homeobject/blob_manager.hpp
```diff
@@ -10,7 +10,7 @@
 namespace homeobject {
 
 ENUM(BlobError, uint16_t, UNKNOWN = 1, TIMEOUT, INVALID_ARG, NOT_LEADER, UNKNOWN_SHARD, UNKNOWN_BLOB, CHECKSUM_MISMATCH,
-     READ_FAILED, INDEX_ERROR, SEALED_SHARD);
+     READ_FAILED, INDEX_ERROR, SEALED_SHARD, TOMBSTONE_BLOB, ALIVE);
```
szmyd, line 13: These are publicly visible but will never be used; we shouldn't add them here or all our consumers are forced to add `case`s for them that are meaningless and can never happen.
From the applications perspective the only things it is concerned with is whether the BLOB did not exist on a READ. Delete on a deleted blob is idempotent and immediate success. So we shouldn't need any new errors in the public API.

szmyd (changes requested):

src/include/homeobject/blob_manager.hpp
szmyd: If we find we do need to add something here (because some external client needs this information) we need to make a corresponding error in the `nuobject_proto` repo so we can return this value from SM; we can bump the version to `0.14.1` here so we don't break SM in the meantime (it's locked to 0.13.x which 0.14.0 is consider part of oddly)

src/lib/homestore_backend/index_kv.cpp
```diff
@@ -70,7 +71,59 @@ HSHomeObject::get_from_index_table(shared< BlobIndexTable > index_table, shard_i
...
+        if (pbas == tombstone_pbas) { return folly::makeUnexpected(BlobError::TOMBSTONE_BLOB); }
+        return pbas;
+    }
+
+    // get tombstone, only used by test for now, can be used by GC thread if needed
+    if (state == BlobState::TOMBSTONE) {
+        if (!(pbas == tombstone_pbas)) { return folly::makeUnexpected(BlobError::ALIVE); }
+    }
+
+    // used for GC thread to range scan index table
```
yamingk, line 87: This comment is not needed, and we use shard_id to do range-query, not pba, right?

src/include/homeobject/blob_manager.hpp
yamingk, line 13: I probably didn't get why TOMBSTONE_BLOB/ALIVE is an BlobError in the first place. 
After reading through the usage of these two error code, both of them can be categorized into INDEX_ERROR, since they all mean we queried some thing from the index with expected state, but index didn't return correctly, right?

src/lib/homestore_backend/index_kv.cpp
JacksonYao287: yea, we use shard_id to do range-query, but we need to check the pba to determine whether this bolb has been deleted, since we put a tombstone pba in value

src/include/homeobject/blob_manager.hpp
JacksonYao287: make sense

JacksonYao287: ack

src/lib/homestore_backend/index_kv.cpp
```diff
@@ -70,7 +71,59 @@ HSHomeObject::get_from_index_table(shared< BlobIndexTable > index_table, shard_i
...
+    if (state == BlobState::TOMBSTONE) {
+        if (!(pbas == tombstone_pbas)) { return folly::makeUnexpected(BlobError::INDEX_ERROR); }
+    }
+
+    // used for GC thread to range scan index table
+    return pbas;
+}
+
+// this is used for GC thread to delete tombstone blobid from btree
+BlobManager::NullResult HSHomeObject::delete_from_index_table(shared< BlobIndexTable > index_table, shard_id_t shard_id,
```
yamingk, line 92: This is not needed, as for GC, it will call index range-update and in the filter callback, it will return remove method for the tombstoned values and btree will remove the tombstoned keys/values atomically.

src/lib/homestore_backend/index_kv.cpp
yamingk: No, we don't need to do that for GC, GC will call range-update to remove tombstoned key/values.

src/lib/homestore_backend/hs_blob_manager.cpp
yamingk: You misunderstood my point. My comment started with "as an integration test", so it is not to test underlying component, which already have their UT to cover its correctness. It is to test HomeObject's deletion blob api, and make sure as a public API, it does what it claims.

JacksonYao287: @yamingk 

>1 delete it from the index. 

this has been tested in the UT of this PR

>2  free the space

this is not what we can guarantee. actually,  what we can only guarantee is that we definitely call `async_free_blks` with the correct parameter. if we call `async_free_blks` with the correct parameter but the space is not freeed as expected, it is the duty of the author of `async_free_blks` to find out the bug, right?

if we go deeper, for any syscall(e.g.  write) , after we call fd#write and sync, do we need to verify the block has been writen to disk and the block content is what we write? for every fd#write, do we need this test in our code?

i believe it is a promise from the API provider that if we call the API with correct parameter, we will get the expected result. the API should be fully tested before it is released, it is not the concern or duty of upper layer which uses it.

xiaoxichen: Technically I believe @JacksonYao287 you only need to ensure by UT(with mock) that `async_free_blks` was called with proper parameters.

Practically as https://github.com/eBay/HomeStore/pull/216/files already expose the `freeable_blks` to vchunk,  having an integration test is fine.

yamingk: If we can have mock inside vchunk it would be good enough I think for this PR.

PR #90, opened 2023-10-19
Support different Executors.

szmyd:
The idea is to add another Executor to `iomanager` and make this one of the enumerations HSHomeObject can use which will wrap `run_on`.


szmyd:
Address issue #91

PR #92, opened 2023-10-19
Fix user key hash calculation.

sanebay:
Fix user key hash calucation by taking the actual user key bytes. 
Test code fix to use resize.


szmyd (dismissed):
lgtm

szmyd (approved):

PR #98, opened 2023-10-30
replay journal log for create/seal shard when HO is crashed before on_commit() is called

zichanglai:
there is a small chance that HO will be crashed before on_commit() is called for create/seal shard when associated journal log is already written(and been replicated to the majority of PG Members when we have real replication), in that case, we need to replay journal log to recover shard info.


src/lib/homestore_backend/hs_shard_manager.cpp
```diff
@@ -183,15 +199,33 @@ void HSHomeObject::do_shard_message_commit(int64_t lsn, ReplicationMessageHeader
...
     switch (header.msg_type) {
     case ReplicationMessageType::CREATE_SHARD_MSG: {
-        add_new_shard_to_map(std::make_unique< HS_Shard >(shard_info, blkids.chunk_num()));
+        bool shard_exist = false;
+        {
+            std::scoped_lock lock_guard(_shard_lock);
+            shard_exist = (_shard_map.find(shard_info.id) != _shard_map.end());
+        }
+
+        if (!shard_exist) { add_new_shard_to_map(std::make_unique< HS_Shard >(shard_info, blkids.chunk_num())); }
```
xiaoxichen, line 208: In this case we need to take the chunk out from `ChunkSelector` right?
There is a racing as we started `ChunkSelector` after metaservice recovered, which make the chunks using by  uncommitted shard looks available. Though it is probably be fine if raft statemachine will not proceed for new on_precommit).

We dont have API to mark the chunk as use

zichanglai: yes, good catch for this case.  I think we need to add such api to make one chunk as using because this api will be needed by GC flow too to pick out the source/destination chunk from the heap chunkselector

xiaoxichen: I dont know how GC flow would create the shadow shard, that part need a design doc.

but chunk_selector need to provide that API @JacksonYao287 . maybe create an issue.

JacksonYao287: sure , we do need this API

src/lib/homestore_backend/hs_shard_manager.cpp
```diff
@@ -183,15 +199,33 @@ void HSHomeObject::do_shard_message_commit(int64_t lsn, ReplicationMessageHeader
...
+            std::scoped_lock lock_guard(_shard_lock);
+            auto iter = _shard_map.find(shard_info.id);
+            RELEASE_ASSERT(iter != _shard_map.end(), "Missing shard info");
+            state = (*iter->second)->info.state;
+        }
+
+        if (state == ShardInfo::State::OPEN) {
+            auto chunk_id = get_shard_chunk(shard_info.id);
+            RELEASE_ASSERT(chunk_id.has_value(), "Chunk id not found");
+            chunk_selector()->release_chunk(chunk_id.value());
```
JacksonYao287, line 225: here is also the  similar problem as mentioned above.

no need to call `release_chunk` , just do not put it in to `excludingChunks` and then `build_per_dev_chunk_heap `

src/lib/homestore_backend/hs_shard_manager.cpp
xiaoxichen: @JacksonYao287  the `build_per_dev_chunk_heap` was called in `on_shard_meta_blk_recover_completed`, however the raft log reply will be after meta-service recovery , all the chunk/shard/PG need to be ready before we can reply IO(like putBlob).

with that said ,the `build_per_dev_chunk_heap` has been called w/o the chunk-id of pending-replay shard.

JacksonYao287: got your point, that is really a problem.

zichanglai: about the implementation of this new API, it need to works with both cases:

1. chunk is already mark busy when write the shard header to the repldev, so when create_shard/seal_shard on_commit() is called, this chunk is already been pop out from the heap chunk selector, this API just does nothing and that is the normal case.
2. chunk is already mark busy when write the shard header to the repldev, before create_shard/seal_shard on_commit() is called, HO is restarted/crashed. after restarts,  the previous chunk will be added back to the chunk selector, so when replay on_commit() will call this new API and this new API will actually remove this chunk from chunk selector.

xiaoxichen: @zichanglai  I am not getting your point. Chunk-selector is stateless,   it rely on shard metablks to remove in-use chunks. So in on_commit() of create_shard, if shard metadata has been updated , no need to touch chunk-selector.  Otherwise need to call the remove_chunk(chunk_id) api.

zichanglai: let us assume this api is HeapChunkSelector::mark_chunk_busy(shard_id_t);   when on_commit()  is called, the associated chunk maybe is already been removed from the selector in here: https://github.com/eBay/HomeStore/blob/ddff069df74d935d10ac8cb9abc5c7ece790c8f0/src/lib/device/virtual_dev.cpp#L192.   but in recover case, this chunk is not removed from selector, so this api needed to handle both cases.

zichanglai: I submit a new PR to add such interface: https://github.com/eBay/HomeObject/pull/103

zichanglai: this issue is addressed in the last commit.

src/lib/homestore_backend/hs_shard_manager.cpp
```diff
@@ -153,10 +153,26 @@ void HSHomeObject::on_shard_message_commit(int64_t lsn, sisl::blob const& header
...
+    repl_dev->async_read(blkids, value, value.size)
+        .thenValue([this, lsn, msg_header = *header_ptr, blkids, value](auto&& err) mutable {
+            if (err) {
+                LOGW("failed to read data from homestore pba, lsn:{}", lsn);
+            } else {
+                sisl::blob value_blob(r_cast< uint8_t* >(value.iovs[0].iov_base), value.size);
+                do_shard_message_commit(lsn, msg_header, blkids, value_blob, nullptr);
+            }
+            iomanager.iobuf_free(r_cast< uint8_t* >(value.iovs[0].iov_base));
+        });
```
szmyd, line 175: I don't understand the need for this addition. When the ReplDev restarts it will replay the operations it thinks are appropriate. In the future this will be driven by `nuraft::raft_server::start()`, but today it will be ReplSoloDev imitating that behavior. Given that, do we need to do anything special?

zichanglai: based on the current behaviors of SoloReplDev, it will call ReplicaSetListener::on_commit() with hs_ctx been set to nullptr when replay the appropriate journal log. Even when we switch to real replication, ReplicaSetListener::on_commit()  will still will be called when replica restarts and replay the log and in that case hs_ctx will still be nullptr, right? 
we do not have the value buf which contains the shard info and has to read from the PBA.

szmyd: @hkadayam I think we've changed this correct? We shouldn't need to read from the Data service to commit a shard creation. The relevant data should be available somewhere in the commit context.

zichanglai: I  revisit  the SoloReplDev restart logic and seems matches my expection here: https://github.com/eBay/HomeStore/blob/1fd188a87b07b776223fb009a59d2238197574b8/src/lib/replication/repl_dev/solo_repl_dev.cpp#L118.   on_commit() will be called with a PBA but leaving ctx to be nullptr.

src/lib/homestore_backend/hs_shard_manager.cpp
```diff
@@ -153,10 +153,26 @@ void HSHomeObject::on_shard_message_commit(int64_t lsn, sisl::blob const& header
...
+    // is already be written into metablk, so we need to do nothing in this case to avoid duplication when replay this
+    // journal log. but there is still a smaller chance that HO is stopped/crashed before writing metablk is called or
+    // completed and we need to recover shard info from journal log in such case.
+    sisl::sg_list value;
+    value.size = blkids.blk_count() * repl_dev->get_blk_size();
+    auto value_buf = iomanager.iobuf_alloc(512, value.size);
+    value.iovs.push_back(iovec{.iov_base = value_buf, .iov_len = value.size});
+    // header will be released when this function returns, but we still need the header when async_read() finished.
+    auto header_ptr = r_cast< const ReplicationMessageHeader* >(header.bytes);
+    repl_dev->async_read(blkids, value, value.size)
```
JacksonYao287, line 166: actually, when restarting,  we want to call `do_shard_message_commit` for the specific shard. But the shard info is in `value` which has been written as a block, so we need to read the block and get the value. this is what here does, right?

i go through the solo_dev, it will first write the block, and then write the log.
https://github.com/eBay/HomeStore/blob/0f460de461b64232d03c235ccfbff97274a0bc60/src/lib/replication/repl_dev/solo_repl_dev.cpp#L45

so if crash happens after the block is written but before the log is written, when restarting , no log for this shard exists, we can not find this issue, and a stale shard info block will be left in the chunk. (not sure will gc handle this?)

if we write log first, then when restart , maybe we cannot read `value` from the disk(the same case as above).
 
so , maybe we should think about what we should do here later?

zichanglai: both mentioned cases is not this PR want to resolve because if the log is not written successfully, the created shard must be failed and regards nothing happens(but may leave a stale data blk and need GC to recycle it). 
what the PR wants to do is when log is already been written, but crashed before on_commit() called or  on_commit() is called but shard metablk is not written yet.  As the log is persisted, so data must be persisted before and we will replay the log to restore the shard info.

JacksonYao287: we can record this case, and no need to solve in this PR. let`s  revisit this issue after raft_repl_dev is merged.
other parts looks good to me!

JacksonYao287 (approved):

PR #100, opened 2023-10-31
Make the different buffers used in the put_blob RAII

raakella1:
The current put blob implementation uses `iobuf_alloc` at the places where it needs to allocate buffers. This results in the user to keep track of these buffers and free them later. This also results in capturing a lot of variables in the completion lambda which is not desirable. 
This PR replaces the iobuf_alloc by using the `sisl::io_blob_safe` which supports RAII.

### DIFF
--- a/src/lib/homestore_backend/hs_blob_manager.cpp
+++ b/src/lib/homestore_backend/hs_blob_manager.cpp
@@ -24,23 +24,24 @@ BlobManager::AsyncResult< blob_id_t > HSHomeObject::_put_blob(ShardInfo const& s
     const uint32_t needed_size = sizeof(ReplicationMessageHeader);
     auto req = repl_result_ctx< BlobManager::Result< BlobInfo > >::make(needed_size, io_align);
 
+    put_blob_ctx put_ctx;
     uint8_t* raw_ptr = req->hdr_buf_.bytes;
-    ReplicationMessageHeader* header = new (raw_ptr) ReplicationMessageHeader();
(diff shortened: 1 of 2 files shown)


sanebay (approved):
LGTM.

src/lib/homestore_backend/hs_blob_manager.cpp
```diff
@@ -24,23 +24,24 @@ BlobManager::AsyncResult< blob_id_t > HSHomeObject::_put_blob(ShardInfo const& s
...
+    put_blob_ctx put_ctx;
     uint8_t* raw_ptr = req->hdr_buf_.bytes;
-    ReplicationMessageHeader* header = new (raw_ptr) ReplicationMessageHeader();
-    header->msg_type = ReplicationMessageType::PUT_BLOB_MSG;
-    header->payload_size = 0;
-    header->payload_crc = 0;
-    header->shard_id = shard.id;
-    header->pg_id = pg_id;
-    header->header_crc = header->calculate_crc();
+    put_ctx.repl_header_ = new (raw_ptr) ReplicationMessageHeader();
```
szmyd, line 29: Where is `delete`?

xiaoxichen: we dont need a delete right? as the `ReplicationMessageHeader` was allocated from `req->hdr_buf_.bytes;` which is `    sisl::io_blob_safe` and will be release as the release of `req`

xiaoxichen (approved):
lgtm

PR #101, opened 2023-11-02
Delete BLOB should be idempotent.

szmyd:
This is not _critical_ as SM treats both errors equally (UNKNOWN_BLOB and INDEX_ERROR). But we should be correct about what we pass back to the client application.

### DIFF
--- a/src/lib/homestore_backend/index_kv.cpp
+++ b/src/lib/homestore_backend/index_kv.cpp
@@ -60,31 +60,20 @@ BlobManager::NullResult HSHomeObject::add_to_index_table(shared< BlobIndexTable 
 }
 
 BlobManager::Result< homestore::MultiBlkId >
-HSHomeObject::get_blob_from_index_table(shared< BlobIndexTable > index_table, shard_id_t shard_id, blob_id_t blob_id,
-                                        BlobState state) const {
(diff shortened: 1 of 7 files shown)


src/lib/homestore_backend/index_kv.cpp
```diff
@@ -75,7 +75,7 @@ HSHomeObject::get_blob_from_index_table(shared< BlobIndexTable > index_table, sh
 
     // blob get API
     if (state == BlobState::ALIVE) {
-        if (pbas == tombstone_pbas) { return folly::makeUnexpected(BlobError::INDEX_ERROR); }
+        if (pbas == tombstone_pbas) { return folly::makeUnexpected(BlobError::UNKNOWN_BLOB); }
```
xiaoxichen, line 78: Cannot make a comment on L63,  but do we really need the `state` in the api call?
It's intent was trying to distinguish between `on purpose reading tombstone` or `trying to read a live blob but hit tombstone`.

can we do 
```
 if (pbas == tombstone_pbas) 
     return folly::makeUnexpected(BlobError::UNKNOWN_BLOB); 
else 
     return pbas;
```

xiaoxichen (approved):

PR #103, opened 2023-11-03
add select specific chunk interface to heap chunk selector for recovery and GC flow

zichanglai:
there are some cases we want to mark one specific chunk as busy in heap chunk selector:

1. GC flow which will mark source chunk and dest chunk as busy to prevent more write traffic to these chunks.
2. recovery flow which may replay some create shard journal logs and need to explicitly make allocated chunk as busy

### DIFF
--- a/src/lib/homestore_backend/heap_chunk_selector.cpp
+++ b/src/lib/homestore_backend/heap_chunk_selector.cpp
@@ -76,15 +76,12 @@ csharedChunk HeapChunkSelector::select_chunk(homestore::blk_count_t count, const
         return nullptr;
     }
 
-    const auto& vchunk = [it]() {
-        auto& heapLock = it->second->mtx;
-        auto& heap = it->second->m_heap;
-        std::lock_guard< std::mutex > l(heapLock);
-        if (heap.empty()) return VChunk(nullptr);
-        VChunk vchunk = heap.top();
+    auto vchunk = VChunk(nullptr);
+    auto& heap = it->second->m_heap;
+    if (auto lock_guard = std::lock_guard< std::mutex >(it->second->mtx); !heap.empty()) {
+        vchunk = heap.top();
         heap.pop();
-        return vchunk;
-    }();
+    }
 
     if (vchunk.get_internal_chunk()) {
         auto& avalableBlkCounter = it->second->available_blk_count;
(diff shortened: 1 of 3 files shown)


src/lib/homestore_backend/heap_chunk_selector.cpp
```diff
@@ -76,7 +76,7 @@ csharedChunk HeapChunkSelector::select_chunk(homestore::blk_count_t count, const
         return nullptr;
     }
 
-    const auto& vchunk = [it = std::move(it)]() {
+    const auto& vchunk = [it]() {
```
szmyd, line 79: Why not just:
```c++
auto vchunk = VChunk(nullptr);
auto& heap = it->second->m_heap;
if (auto heap_guard = std::lock_guard<std::mutex>(it->second->mtx); !heap.empty()) {
   vchunk = heap.top();
   heap.pop();
}
```
which avoids the allocation for the lambda?

src/lib/homestore_backend/heap_chunk_selector.cpp
```diff
@@ -96,6 +96,50 @@ csharedChunk HeapChunkSelector::select_chunk(homestore::blk_count_t count, const
...
 }
 
+csharedChunk HeapChunkSelector::select_specific_chunk(const chunk_num_t chunkID) {
+    if (m_chunks.find(chunkID) == m_chunks.end()) {
+        // sanity check
+        LOGWARNMOD(homeobject, "No chunk found for ChunkID {}", chunkID);
+        return nullptr;
+    }
+
+    auto pdevID = VChunk(m_chunks[chunkID]).get_pdev_id();
```
szmyd, line 106: You've enforced `chunkID` be `const` to this stack, but not this? I would make `chunkID` non-const personally, but suggest consistency either way.

src/lib/homestore_backend/heap_chunk_selector.cpp
```diff
@@ -96,6 +96,50 @@ csharedChunk HeapChunkSelector::select_chunk(homestore::blk_count_t count, const
     return vchunk.get_internal_chunk();
 }
 
+csharedChunk HeapChunkSelector::select_specific_chunk(const chunk_num_t chunkID) {
+    if (m_chunks.find(chunkID) == m_chunks.end()) {
+        // sanity check
+        LOGWARNMOD(homeobject, "No chunk found for ChunkID {}", chunkID);
```
szmyd, line 102: `LOGW(...` macro is same as `LOGWARNMOD(homeobject, ...`

src/lib/homestore_backend/heap_chunk_selector.cpp
```diff
@@ -96,6 +96,50 @@ csharedChunk HeapChunkSelector::select_chunk(homestore::blk_count_t count, const
...
+    }
+
+    auto pdevID = VChunk(m_chunks[chunkID]).get_pdev_id();
+    auto it = m_per_dev_heap.find(pdevID);
+    if (it == m_per_dev_heap.end()) {
+        LOGWARNMOD(homeobject, "No pdev found for pdev {}", pdevID);
+        return nullptr;
+    }
+
+    const auto& vchunk = [it, chunkID]() {
```
szmyd, line 113:
```c++
auto vchunk = VChunk(nullptr);
auto& heap = it->second->m_heap;
while (auto heap_guard = std::lock_guard<std::mutex>(it->second->mtx); !heap.empty()) {
   ...
}
```
will work here too?

src/lib/homestore_backend/heap_chunk_selector.cpp
zichanglai: good suggestion and that make the code more simple.

src/lib/homestore_backend/heap_chunk_selector.cpp
zichanglai: sure, will use auto const pdevID.

src/lib/homestore_backend/heap_chunk_selector.cpp
zichanglai: yes, same as above, will update here too.

src/lib/homestore_backend/heap_chunk_selector.cpp
zichanglai: I noticed this before, but since heap chunk selector is an independent module in HO and macro LOGW(....) is defined in homeobject_impl.hpp and I try to avoid include homeobject_impl.hpp to heap chunk selector to keep it simple and independence.  I think that is why LOGWARNMOD(homeobject,....) is still being widely used in heap chunk selector.

JacksonYao287 (approved):
LGTM

szmyd (approved):
lgtm

PR #104, opened 2023-11-03
Generate blob id from an atomic counter

raakella1:
We maintain a per-PG sequence number and use it to generate a blob id. This sequence number is persisted to pg metablk periodically on an iomgr user reactor by scheduling a global timer. This can be extended to trigger any other timer events in homeobject.

The blob id is generated by the leader during a put blob API and use it as a key for the call `async_alloc_write`

Testing: Added unit test to put blob after restart which will verify the persistence of the blob seq num. 

TODO:

- This PR does not handle the case when we recover from raft snapshot
- As a safety measure, a new raft leader should add a fixed value to his existing sequence number to avoid reusing blob id.

### DIFF
--- a/src/lib/homeobject_impl.hpp
+++ b/src/lib/homeobject_impl.hpp
@@ -63,7 +63,7 @@ struct PG {
 
     PGInfo pg_info_;
     uint64_t shard_sequence_num_{0};
-    std::atomic_uint64_t blob_sequence_num_{0};
+    std::atomic< blob_id_t > blob_sequence_num_{0ull};
     ShardPtrList shards_;
 };
 
--- a/src/lib/homeobject_impl.hpp
+++ b/src/lib/homeobject_impl.hpp
@@ -63,6 +63,7 @@ struct PG {
 
     PGInfo pg_info_;
     uint64_t shard_sequence_num_{0};
+    std::atomic_uint64_t blob_sequence_num_{0};
     ShardPtrList shards_;
 };
 
--- a/src/lib/homestore_backend/hs_blob_manager.cpp
+++ b/src/lib/homestore_backend/hs_blob_manager.cpp
@@ -12,17 +12,21 @@ static constexpr uint64_t io_align{512};
 BlobManager::AsyncResult< blob_id_t > HSHomeObject::_put_blob(ShardInfo const& shard, Blob&& blob) {
     auto& pg_id = shard.placement_group;
     shared< homestore::ReplDev > repl_dev;
+    blob_id_t new_blob_id;
     {
         std::shared_lock lock_guard(_pg_lock);
         auto iter = _pg_map.find(pg_id);
         RELEASE_ASSERT(iter != _pg_map.end(), "PG not found");
         repl_dev = static_cast< HS_PG* >(iter->second.get())->repl_dev_;
+        new_blob_id = iter->second->blob_sequence_num_.fetch_add(1, std::memory_order_relaxed);
+        RELEASE_ASSERT(new_blob_id < std::numeric_limits< decltype(new_blob_id) >::max(),
+                       "exhausted all available blob ids");
     }
 
     RELEASE_ASSERT(repl_dev != nullptr, "Repl dev instance null");
 
     const uint32_t needed_size = sizeof(ReplicationMessageHeader);
-    auto req = repl_result_ctx< BlobManager::Result< BlobInfo > >::make(needed_size, io_align);
+    auto req = put_result_ctx< BlobManager::Result< BlobInfo > >::make(needed_size, io_align, new_blob_id);
 
     uint8_t* raw_ptr = req->hdr_buf_.bytes;
     ReplicationMessageHeader* header = new (raw_ptr) ReplicationMessageHeader();
@@ -44,6 +48,7 @@ BlobManager::AsyncResult< blob_id_t > HSHomeObject::_put_blob(ShardInfo const& s
     blob_header->magic = BlobHeader::blob_header_magic;
     blob_header->version = BlobHeader::blob_header_version;
     blob_header->shard_id = shard.id;
+    blob_header->blob_id = new_blob_id;
     blob_header->hash_algorithm = BlobHeader::HashAlgorithm::CRC32;
     blob_header->blob_size = blob.body.size;
     blob_header->user_key_size = blob.user_key.size();
@@ -124,9 +129,9 @@ BlobManager::AsyncResult< blob_id_t > HSHomeObject::_put_blob(ShardInfo const& s
 void HSHomeObject::on_blob_put_commit(int64_t lsn, sisl::blob const& header, sisl::blob const& key,
                                       const homestore::MultiBlkId& pbas,
                                       cintrusive< homestore::repl_req_ctx >& hs_ctx) {
-    repl_result_ctx< BlobManager::Result< BlobInfo > >* ctx{nullptr};
+    put_result_ctx< BlobManager::Result< BlobInfo > >* ctx{nullptr};
     if (hs_ctx != nullptr) {
-        ctx = boost::static_pointer_cast< repl_result_ctx< BlobManager::Result< BlobInfo > > >(hs_ctx).get();
+        ctx = boost::static_pointer_cast< put_result_ctx< BlobManager::Result< BlobInfo > > >(hs_ctx).get();
     }
 
     auto msg_header = r_cast< ReplicationMessageHeader* >(header.bytes);
@@ -143,11 +148,14 @@ void HSHomeObject::on_blob_put_commit(int64_t lsn, sisl::blob const& header, sis
         RELEASE_ASSERT(iter != _pg_map.end(), "PG not found");
         index_table = static_cast< HS_PG* >(iter->second.get())->index_table_;
         RELEASE_ASSERT(index_table != nullptr, "Index table not intialized");
+        if (iter->second->blob_sequence_num_.load() <= ctx->blob_id_) {
+            iter->second->blob_sequence_num_.store(ctx->blob_id_ + 1);
+        }
     }
 
     BlobInfo blob_info;
     blob_info.shard_id = msg_header->shard_id;
-    blob_info.blob_id = lsn;
+    blob_info.blob_id = ctx->blob_id_;
     blob_info.pbas = pbas;
 
     // Write to index table with key {shard id, blob id } and value {pba}.
--- a/src/lib/homestore_backend/hs_blob_manager.cpp
+++ b/src/lib/homestore_backend/hs_blob_manager.cpp
@@ -162,7 +162,7 @@ void HSHomeObject::on_blob_put_commit(int64_t lsn, sisl::blob const& header, sis
     auto r = add_to_index_table(index_table, blob_info);
     if (r.hasError()) {
         LOGE("Failed to insert into index table for blob {} err {}", lsn, r.error());
-        ctx->promise_.setValue(folly::makeUnexpected(r.error()));
+        if (ctx) { ctx->promise_.setValue(folly::makeUnexpected(r.error())); }
         return;
     }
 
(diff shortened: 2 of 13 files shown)


xiaoxichen (commented):
we dont check `blob_id` in blob_header when doing get_blob,  this was because we dont set it before this PR, please if you can add the verification there

src/lib/homestore_backend/tests/hs_blob_tests.cpp
```diff
@@ -155,39 +178,13 @@ TEST_F(HomeObjectFixture, BasicPutGetDelBlobWRestart) {
...
-        EXPECT_EQ(blob.user_key, result.user_key);
-    }
+    verify_get_blob(blob_map);
 
-    // Verify all get blobs with random offset and length.
-    for (const auto& [id, blob] : blob_map) {
-        std::uniform_int_distribution< uint32_t > rand_off_gen{0u, blob.body.size - 1u};
-        std::uniform_int_distribution< uint32_t > rand_len_gen{1u, blob.body.size};
+    // Put blob after restart to test the persistance of blob sequence number
+    put_blob(blob_map, pg_shard_id_vec, num_blobs_per_shard, max_blob_size);
```
xiaoxichen, line 184: how you verify this ? should we check the blob_id existence in blob_map?

src/lib/homestore_backend/hs_homeobject.cpp
```diff
@@ -89,6 +91,37 @@ void HSHomeObject::init_homestore() {
...
+                ctx->created = true;
+            }
+            ctx->cv.notify_one();
+        }
+    });
+
+    {
+        std::unique_lock< std::mutex > lk{ctx->mtx};
+        ctx->cv.wait(lk, [&ctx] { return ctx->created; });
+    }
```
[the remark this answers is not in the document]
raakella1: The lambda that returns `ctx->created;` is an argument (predicate) to the `cv.wait()` method, it is not a part of the main function
https://en.cppreference.com/w/cpp/thread/condition_variable/wait

src/lib/homestore_backend/hs_pg_manager.cpp
```diff
@@ -193,6 +193,17 @@ HSHomeObject::HS_PG::HS_PG(PGInfo info, shared< homestore::ReplDev > rdev, share
 
 HSHomeObject::HS_PG::HS_PG(homestore::superblk< HSHomeObject::pg_info_superblk > const& sb,
                            shared< homestore::ReplDev > rdev) :
-        PG{pg_info_from_sb(sb)}, pg_sb_{sb}, repl_dev_{std::move(rdev)} {}
+        PG{pg_info_from_sb(sb)}, pg_sb_{sb}, repl_dev_{std::move(rdev)} {
+    blob_sequence_num_ = sb->blob_sequence_num;
+}
+
+void HSHomeObject::persist_pg_sb() {
+    auto lg = std::shared_lock(_pg_lock);
```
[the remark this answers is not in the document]
raakella1: Can you explain a little more? Do you mean we can keep track of the dirty PG and flush only those?

src/lib/homestore_backend/tests/hs_blob_tests.cpp
```diff
@@ -72,6 +72,67 @@ class HomeObjectFixture : public ::testing::Test {
...
+        }
+    }
+
+    void verify_get_blob(blob_map_t const& blob_map, bool const use_random_offset = false) {
+        uint32_t off = 0, len = 0;
+        for (const auto& [id, blob] : blob_map) {
+            int64_t pg_id = std::get< 0 >(id), shard_id = std::get< 1 >(id), blob_id = std::get< 2 >(id);
+            len = blob.body.size;
+            if (use_random_offset) {
+                std::uniform_int_distribution< uint32_t > rand_off_gen{0u, blob.body.size - 1u};
```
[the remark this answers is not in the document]
raakella1: Where do we need the check? If the `use_random_offset` flag is set we generate a value to the len, otherwise we don't use the len.

src/lib/homestore_backend/tests/hs_blob_tests.cpp
raakella1: The index layer in blob manager will return an error if we try to insert a blob id that exists. If the blob id is not persisted, the counter is reset to 0 after restart and we will get the error. I ran this test without my persistence change and it was failing as expected.

src/lib/homestore_backend/tests/hs_blob_tests.cpp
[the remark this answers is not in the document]
xiaoxichen: yes for the `rand_len_gen` you set it to [1, blob_size] which avoid the len to be zero.
I am asking,  by our api definition len=0 means read till end of the blob,   are we on-purpose skipping this case or just overlook?

src/lib/homestore_backend/tests/hs_blob_tests.cpp
xiaoxichen: thx yes we have 
```
    homestore::BtreeSinglePutRequest put_req{&index_key, &index_value,
                                             homestore::btree_put_type::INSERT_ONLY_IF_NOT_EXISTS};
```

src/lib/homestore_backend/hs_pg_manager.cpp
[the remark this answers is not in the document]
xiaoxichen: yes.  But as said might not worth the effort.

src/lib/homestore_backend/hs_homeobject.cpp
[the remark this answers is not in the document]
xiaoxichen: oh thx.

src/lib/homestore_backend/tests/hs_blob_tests.cpp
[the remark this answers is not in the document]
raakella1: ohh yeah, I just modularized the existing code into functions to re-use the code. I am not sure why len is generated from 1 to blob_size.
I think it is because the other case (reading till the end of the blob) is already covered above. (use_random_offset = false case)

src/lib/homestore_backend/hs_blob_manager.cpp
```diff
@@ -143,11 +148,14 @@ void HSHomeObject::on_blob_put_commit(int64_t lsn, sisl::blob const& header, sis
...
         RELEASE_ASSERT(index_table != nullptr, "Index table not intialized");
+        if (iter->second->blob_sequence_num_.load() <= ctx->blob_id_) {
+            iter->second->blob_sequence_num_.store(ctx->blob_id_ + 1);
+        }
     }
 
     BlobInfo blob_info;
     blob_info.shard_id = msg_header->shard_id;
-    blob_info.blob_id = lsn;
+    blob_info.blob_id = ctx->blob_id_;
```
zichanglai, line 158: ctx can be nullptr when HO restarted and replay the journal log, right?
I think in that case may need to read the blob header from the pba and decode it from the blob header. or put the copy of blob id to the header too.

src/lib/homestore_backend/hs_blob_manager.cpp
```diff
@@ -12,17 +12,21 @@ static constexpr uint64_t io_align{512};
...
     auto& pg_id = shard.placement_group;
     shared< homestore::ReplDev > repl_dev;
+    blob_id_t new_blob_id;
     {
         std::shared_lock lock_guard(_pg_lock);
         auto iter = _pg_map.find(pg_id);
         RELEASE_ASSERT(iter != _pg_map.end(), "PG not found");
         repl_dev = static_cast< HS_PG* >(iter->second.get())->repl_dev_;
+        new_blob_id = iter->second->blob_sequence_num_.fetch_add(1, std::memory_order_relaxed);
+        RELEASE_ASSERT(new_blob_id < std::numeric_limits< decltype(new_blob_id) >::max(),
```
szmyd, line 22: How can `new_blob_id` hold a value larger than it itself can hold? Was this meant to be `decltype(blob_sequence_num_) > ::max()`?

src/lib/homeobject_impl.hpp
```diff
@@ -63,6 +63,7 @@ struct PG {
 
     PGInfo pg_info_;
     uint64_t shard_sequence_num_{0};
+    std::atomic_uint64_t blob_sequence_num_{0};
```
szmyd, line 66: Can we do:
```c++
std::atomic<blob_id_t> blob_sequence_num_{0};
```
So it's clear what it's value is representing?

src/lib/homestore_backend/hs_homeobject.hpp
```diff
@@ -65,6 +65,7 @@ class HSHomeObject : public HomeObjectImpl {
         uint32_t num_members;
         peer_id_t replica_set_uuid;
         homestore::uuid_t index_table_uuid;
+        uint64_t blob_sequence_num;
```
szmyd, line 68: `blob_id_t`?

src/lib/homeobject_impl.hpp
szmyd: We can also remove the `shard_seq_num_` from the `ShardIndex` in memory implementation if we have a base counter now right?

src/lib/homestore_backend/replication_state_machine.hpp
```diff
@@ -33,6 +33,19 @@ struct repl_result_ctx : public ho_repl_ctx {
     folly::SemiFuture< T > result() { return promise_.getSemiFuture(); }
 };
 
+template < typename T >
+struct put_result_ctx : public repl_result_ctx< T > {
```
szmyd, line 37: Can we specialize `put_result_ctx` on `repl_result_ctx<BlobManager::Result< BlobInfo >>`? Is there a put that would be a different result?

src/lib/homestore_backend/hs_blob_manager.cpp
szmyd: There should be no difference in the commit path when replaying logs versus playing them in real-time. If there is an issue replaying from the log, that issue would also exist on any follower appending/committing in the normal runtime.

src/lib/homestore_backend/hs_homeobject.cpp
```diff
@@ -89,6 +91,37 @@ void HSHomeObject::init_homestore() {
     LOGI("Initialize and start HomeStore is successfully");
 }
 
+void HSHomeObject::init_timer_thread() {
+    struct Context {
+        std::condition_variable cv;
+        std::mutex mtx;
+        bool created{false};
+    };
+    auto ctx = std::make_shared< Context >();
```
szmyd, line 100: This can just live on the stack right? We ensure it will out-live the lambda with the `wait` call immediately following. It's already being captured by reference.

src/lib/homestore_backend/hs_homeobject.cpp
```diff
@@ -89,6 +91,37 @@ void HSHomeObject::init_homestore() {
     LOGI("Initialize and start HomeStore is successfully");
 }
 
+void HSHomeObject::init_timer_thread() {
+    struct Context {
+        std::condition_variable cv;
+        std::mutex mtx;
+        bool created{false};
+    };
```
szmyd, line 99: How about:
```c++
#include <latch>
...
auto ctx = std::latch{1};
iomanger.create_reactory(..., [&ctx] ...) {
   ...
   ctx.count_down();
});
ctx.wait();
```

src/lib/homestore_backend/hs_blob_manager.cpp
zichanglai: I am not sure if ctx will be nullptr when follower apply/replay the commited log as we do not have real replication until now. but for HS SoloReplDev,  when it restarts, it will replay all commited log  from the checkpoint LSN in here with ctx is nullptr(https://github.com/eBay/HomeStore/blob/391168223297f672a03fd95e1ddef810535b005a/src/lib/replication/repl_dev/solo_repl_dev.cpp#L117)

raakella1: I am using the attribute `key` to serialize blob id now. This problem should be resolved

src/lib/homestore_backend/replication_state_machine.hpp
raakella1: Now that I am using the `key` field for blob_id, I dont need the `put_result_ctx` anymore. I am removing this struct.

szmyd (approved):

PR #105, opened 2023-11-08
Adopt recovery order of HS metaservice

JacksonYao287:
since we add recover dependency logic in HS, this PR adopt this functionality and simplify the recover logic in homeobject

### DIFF
--- a/src/lib/homestore_backend/hs_homeobject.cpp
+++ b/src/lib/homestore_backend/hs_homeobject.cpp
@@ -105,14 +105,16 @@ void HSHomeObject::register_homestore_metablk_callback() {
     HomeStore::instance()->meta_service().register_handler(
         _shard_meta_name,
         [this](homestore::meta_blk* mblk, sisl::byte_view buf, size_t size) { on_shard_meta_blk_found(mblk, buf); },
-        [this](bool success) { on_shard_meta_blk_recover_completed(success); }, true);
+        [this](bool success) { on_shard_meta_blk_recover_completed(success); }, true,
+        std::optional< meta_subtype_vec_t >(meta_subtype_vec_t{_pg_meta_name}));
 
     HomeStore::instance()->meta_service().register_handler(
         _pg_meta_name,
         [this](homestore::meta_blk* mblk, sisl::byte_view buf, size_t size) {
             on_pg_meta_blk_found(std::move(buf), voidptr_cast(mblk));
         },
-        nullptr, true);
+        // TODO: move "repl_dev" to homestore::repl_dev and "index" to homestore::index
+        nullptr, true, std::optional< meta_subtype_vec_t >(meta_subtype_vec_t{"repl_dev", "index"}));
 }
 
 HSHomeObject::~HSHomeObject() {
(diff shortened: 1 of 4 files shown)


src/lib/homestore_backend/hs_homeobject.cpp
```diff
@@ -135,14 +137,9 @@ void HSHomeObject::on_shard_meta_blk_found(homestore::meta_blk* mblk, sisl::byte
         add_new_shard_to_map(std::make_unique< HS_Shard >(sb));
         return;
     }
-
```
zichanglai, line 138: Line 128 ~ 137 is no need anymore, but just keep L135

JacksonYao287: ack

src/lib/homestore_backend/hs_homeobject.cpp
```diff
@@ -105,14 +105,16 @@ void HSHomeObject::register_homestore_metablk_callback() {
     HomeStore::instance()->meta_service().register_handler(
         _shard_meta_name,
         [this](homestore::meta_blk* mblk, sisl::byte_view buf, size_t size) { on_shard_meta_blk_found(mblk, buf); },
-        [this](bool success) { on_shard_meta_blk_recover_completed(success); }, true);
+        [this](bool success) { on_shard_meta_blk_recover_completed(success); }, true,
+        std::optional< meta_subtype_vec_t >(meta_subtype_vec_t{_pg_meta_name}));
```
szmyd, line 109: Do we really have to wrap all these arguments with `std::optional<...>`, won't the compiler deduce this type automatically from the function signature?

src/lib/homestore_backend/hs_homeobject.cpp
```diff
@@ -105,14 +105,16 @@ void HSHomeObject::register_homestore_metablk_callback() {
...
 
     HomeStore::instance()->meta_service().register_handler(
         _pg_meta_name,
         [this](homestore::meta_blk* mblk, sisl::byte_view buf, size_t size) {
             on_pg_meta_blk_found(std::move(buf), voidptr_cast(mblk));
         },
-        nullptr, true);
+        // TODO: move "repl_dev" to homestore::repl_dev and "index" to homestore::index
+        nullptr, true, std::optional< meta_subtype_vec_t >(meta_subtype_vec_t{"repl_dev", "index"}));
```
yamingk, line 117: apply dependency to index is actually unnecessary and will slow down the recovery time.

src/lib/homestore_backend/hs_pg_manager.cpp
```diff
@@ -137,27 +137,14 @@ void HSHomeObject::on_pg_meta_blk_found(sisl::byte_view const& buf, void* meta_c
...
-                }
-                pending_recovery_shards_.erase(iter);
-            }
-
             // During PG recovery check if index is already recoverd else
             // add entry in map, so that index recovery can update the PG.
             std::scoped_lock lg(index_lock_);
             auto uuid_str = boost::uuids::to_string(pg_sb->index_table_uuid);
             auto it = index_table_pg_map_.find(uuid_str);
-            if (it != index_table_pg_map_.end()) {
```
yamingk, line 155: This should be kept, as there is no need for it->second.index_table to be there at this point. It will be set from index service callback.

PR #112, opened 2023-11-14
correct PG/Shard metablk client name during recovery

zichanglai:


zichanglai:
This PR depends on @JacksonYao287 's latest PR to fix the HO build failure: https://github.com/eBay/HomeObject/pull/111

xiaoxichen (approved):
lgtm

Can we disable the default constructor of  `homestore::superblk<T>` ? This seems risky.

zichanglai:
> lgtm
> 
> Can we disable the default constructor of `homestore::superblk<T>` ? This seems risky.

actually, when I take a look on the implementation of HS superblock,  it will not make any mistakes even we have not correct the name of the metablk because sb.write() function will not use the metablk name again when sb has a cache metablk handler in here: https://github.com/eBay/HomeStore/blob/6897d3eca84506c0cbb36ec77fd38da03dc6aa80/src/include/homestore/superblk_handler.hpp#L76. 
The name is used only when the sb is first time to write. when we recover, it will use the cache metablk as the destination to write. but it will make coder more perfect if we had set it correctly.

JacksonYao287 (approved):
LGTM！

src/lib/homestore_backend/tests/hs_shard_tests.cpp
```diff
@@ -169,35 +169,28 @@ TEST_F(ShardManagerTestingRecovery, ShardManagerRecovery) {
...
     ASSERT_TRUE(!!e);
-
     ShardInfo shard_info = e.value();
     EXPECT_EQ(ShardInfo::State::OPEN, shard_info.state);
     EXPECT_EQ(Mi, shard_info.total_capacity_bytes);
     EXPECT_EQ(Mi, shard_info.available_capacity_bytes);
     EXPECT_EQ(0ul, shard_info.deleted_capacity_bytes);
     EXPECT_EQ(_pg_id, shard_info.placement_group);
 
+    // restart homeobject and check if pg/shard info will be recovery
```
yamingk, line 179: nit: recovery-> recovered successfully

yamingk (approved):

yamingk:
@zichanglai I forget to mention there is another sb in HomeObject  svc_info_superblk_t that is also using the similar logic during recovery, can you also check whether it needs to be fixied?

zichanglai:
> similar

Hi @yamingk , I had check it before, svc_info_superblock do not have this issue as it construct with correct client name: https://github.com/eBay/HomeObject/blob/a85345a29ef2dc0395503bc5db1e990cc1e4f430/src/lib/homestore_backend/hs_homeobject.cpp#L120C53-L120C53

Issue #102, opened 2023-11-02
Shard Sequence Number recovery


raakella1:
During recovery, the shard sequence number is reset to 0. We need to persist the number in metablk and recover on restart
Improve unit tests

zichanglai:
do we really has this issue? when shard is recovered from metalk, pg shard sequence num will be updated in here: https://github.com/eBay/HomeObject/blob/3bf233a1cb0d5bb475ca78700f1f4c75056abb23/src/lib/homestore_backend/hs_shard_manager.cpp#L219

zichanglai:
I enhanced the shard recover unit test to cover this: https://github.com/eBay/HomeObject/pull/114. And from the unit test, we knew PG shard_sequence_num can be recovered after restarted.

PR #114, opened 2023-11-17
enhance shard recovery test case

zichanglai:
add unit test  to verify shard sequence num will be recovered after HO restarted.


szmyd:
This looks good generally to me, but will let @raakella1 approve since he wrote the original issue #102

raakella1 (approved):
LG!

JacksonYao287 (changes requested):
LGTM

JacksonYao287 (approved):

PR #115, opened 2023-11-18
Issue: 109: Checkpointing for HomeObject 

yamingk:
Changes.
1. Allow HomeObject to participant checkpointing. 
3. Created dirty list for PG SB, which currently is empty. Any dirty candidate that want to join CP can be added to the same place or any place that makes sense for the author. 
4. Some refactor change to test files 
5. convert existing blob_id maintenance to CP (replace periodic timer, but not a must)

Testing:
======
1. Added a basic CP test for create a PG/Shard and put some random data in the dirty PG 
3. trigger cp
4. restart homestore and assert the PG recovered should have the same dirtied value set before restart.
6. Test passed.


src/lib/homestore_backend/hs_pg_manager.cpp
```diff
@@ -193,4 +193,17 @@ void HSHomeObject::persist_pg_sb() {
...
+void HSHomeObject::persist_pg_sb(pg_id_t id) {
+    {
+        std::scoped_lock lock_guard(_pg_lock);
+        auto const pg_it = _pg_map.find(id);
+        if (pg_it == _pg_map.end()) {
+            // relax this assert if PG remove is supported;
+            DEBUG_ASSERT(false, "persist_pg_sb: pg {} not found", id);
+        } else {
+            auto hs_pg = static_cast< HS_PG* >(pg_it->second.get());
+            hs_pg->pg_sb_.write();
```
JacksonYao287, line 205: hs_pg->pg_sb_->blob_sequence_num should be the same withe current hs_pg->blob_sequence_num_;

src/lib/homestore_backend/hs_hmobj_cp.hpp
```diff
@@ -0,0 +1,70 @@
...
+// This is a per_cp context for home object.
+// When a new CP is created, a new HomeObjCPContext is created by CP Manager;
+// CP consumer doesn't need to free the dirty list inside this context as it will be automatically freed when this cp is
+// completed (goes out of life cycle) which is controlled by cp manager;
+//
+class HomeObjCPContext : public CPContext {
+public:
+    HomeObjCPContext(CP* cp);
+    virtual ~HomeObjCPContext() = default;
+    void add_pg_to_dirty_list(pg_id_t pg_id);
```
JacksonYao287, line 63: where this is used in HO , except testing

JacksonYao287 (commented):
@yamingk 
1 we already have a timer for flushing pg super block periodically, if we have CP for homeobject,  we could remove that timer. 

2 the dirty list can not guarantee that it will be flushed to disk(crash before trigger_cp_flush), so we need a separate log  to make sure all the updates will be recovered when restart.  if we use cp for homeobject, we should implement that log, right?

yamingk:
@JacksonYao287 
| >>>>1.  we already have a timer for flushing pg super block periodically, if we have CP for homeobject, we could remove that timer.
As was discussed in HS meeting, @raakella1 is going to combine the timer with CP. 

| >>>>> 2 the dirty list can not guarantee that it will be flushed to disk(crash before trigger_cp_flush), so we need a separate log to make sure all the updates will be recovered when restart. if we use cp for homeobject, we should implement that log, right?
yes. Everything writing to CP has to be protected by log unless the consumer are aware of it and want to garunteen the idompotency itself.

src/lib/homestore_backend/hs_pg_manager.cpp
yamingk: it is pointing to the same instance of pg_sb_, there is no difference.

src/lib/homestore_backend/hs_hmobj_cp.hpp
yamingk: explained in other comment.

src/lib/homestore_backend/hs_hmobj_cp.hpp
```diff
@@ -0,0 +1,70 @@
...
+    virtual ~HomeObjCPCallbacks() = default;
+
+public:
+    std::unique_ptr< CPContext > on_switchover_cp(CP* cur_cp, CP* new_cp) override;
+    folly::Future< bool > cp_flush(CP* cp) override;
+    void cp_cleanup(CP* cp) override;
+    int cp_progress_percent() override;
+
+private:
+    HSHomeObject* home_obj_{nullptr}; // it is a raw pointer because HSHomeObject triggers shutdown in its destructor,
```
raakella1, line 49: Do we know that the lifecycle of HSHomeObject is greater than HomeObjCPCallbacks?

raakella1 (commented):
Minor comment, just for knowledge purpose, LG otherwise

src/lib/homestore_backend/hs_hmobj_cp.cpp
```diff
@@ -0,0 +1,52 @@
...
+
+// when cp_flush is called, it means that all the dirty candidates are already in the dirty list.
+// new dirty candidates will arrive on next cp's context.
+folly::Future< bool > HomeObjCPCallbacks::cp_flush(CP* cp) {
+    auto cp_ctx = s_cast< HomeObjCPContext* >(cp->context(homestore::cp_consumer_t::HS_CLIENT));
+
+    // start to flush all dirty candidates.
+    // no need to take the lock as no more dirty candidate is going to be added to this context when we are here;
+    for (auto it = cp_ctx->pg_dirty_list_.begin(); it != cp_ctx->pg_dirty_list_.end(); ++it) {
+        home_obj_->persist_pg_sb(*it);
```
raakella1, line 33: We take the pg lock for each pg id in the dirty list. can we optimize it?

src/lib/homestore_backend/hs_hmobj_cp.hpp
yamingk: yes. HomeObject in its destruct will call will call homestore shutdown which will do a sync trigger cp which will in return access this homeobj instance. During homestore shutdown, cp manager will be stopped and destroyed which in return destroy all the registered customer of cp manager. After homestore shutdown completes, homeobj destructor continues its destruction.

src/lib/homestore_backend/hs_homeobject.cpp
```diff
@@ -27,6 +29,7 @@ extern std::shared_ptr< HomeObject > init_homeobject(std::weak_ptr< HomeObjectAp
     auto instance = std::make_shared< HSHomeObject >(std::move(application));
     instance->init_homestore();
     instance->init_timer_thread();
+    instance->init_cp();
```
zichanglai, line 32: when we had checkpoint for homeobject,  there is no need to add the extra timer thread, right?

src/lib/homestore_backend/hs_homeobject.hpp
```diff
@@ -72,6 +73,24 @@ class HSHomeObject : public HomeObjectImpl {
...
+        pg_info_superblk() = default;
+        pg_info_superblk(pg_info_superblk const& rhs) { *this = rhs; }
+
+        pg_info_superblk& operator=(pg_info_superblk const& rhs) {
+            id = rhs.id;
+            num_members = rhs.num_members;
+            replica_set_uuid = rhs.replica_set_uuid;
+            index_table_uuid = rhs.index_table_uuid;
+            blob_sequence_num = rhs.blob_sequence_num;
+            memcpy(members, rhs.members, sizeof(pg_members) * num_members);
```
JacksonYao287, line 89: is this safe to copy `sizeof(pg_members) * num_members` bytes to `members`, which can hold only one `pg_members`?

src/lib/homestore_backend/hs_pg_manager.cpp
```diff
@@ -174,14 +176,23 @@ HSHomeObject::HS_PG::HS_PG(PGInfo info, shared< homestore::ReplDev > rdev, share
         pg_sb_->members[i].priority = m.priority;
         ++i;
     }
-
     pg_sb_.write();
```
JacksonYao287, line 179: since we use cp_flush to write pg_sb, do we need this write here?

src/lib/homestore_backend/hs_pg_manager.cpp
```diff
@@ -174,14 +176,23 @@ HSHomeObject::HS_PG::HS_PG(PGInfo info, shared< homestore::ReplDev > rdev, share
...
+    init_cp();
+}
+
+void HSHomeObject::HS_PG::init_cp() {
+    cache_pg_sb_ = (pg_info_superblk*)malloc(pg_sb_->size());
+    cache_pg_sb_->copy(*(pg_sb_.get()));
```
JacksonYao287, line 185: i prefer to remove `copy` , and overwrite operator new for `pg_info_superblk`. i think it will be more readable

src/lib/homestore_backend/hs_hmobj_cp.hpp
```diff
@@ -0,0 +1,95 @@
...
+     * Memory:
+     * This function will allocate memory for the pg sb if it is the first time the pg sb is added;
+     * otherwise, it will reuse the memory allocated before.
+     *
+     * @param pg_sb A pointer to the page superblock to be added to the dirty list.
+     */
+    void add_pg_to_dirty_list(HSHomeObject::pg_info_superblk* pg_sb);
+
+    static void init_pg_sb(homestore::superblk< HSHomeObject::pg_info_superblk >&& sb) {
+        pg_sb_[sb->id] = std::move(sb); // move the sb to the map;
```
JacksonYao287, line 85: a lock to protect this in case two or more pg are created simutanously

src/lib/homestore_backend/hs_pg_manager.cpp
yamingk: The idea is for PG creation (or via recovery), it will always be written in PG's constructor. After a particular PG is created, pg sb write always go through cp_flush, which is log protected.

src/lib/homestore_backend/hs_homeobject.hpp
yamingk: I think the idea behind the original author is that members is always come from the heap, and its pointed size is always the size of num_members.

src/lib/homestore_backend/hs_hmobj_cp.hpp
yamingk: done

raakella1 (approved):
LG!

PR #117, opened 2023-11-23
Implement HeapChunkSelector#select_chunk_for_GC()

JacksonYao287:


src/lib/homestore_backend/heap_chunk_selector.cpp
```diff
@@ -130,11 +135,44 @@ csharedChunk HeapChunkSelector::select_specific_chunk(const chunk_num_t chunkID)
...
+        }
+        auto chunk = select_specific_chunk(chunkID);
+        if (chunk) return chunk;
+    }
+    return nullptr;
+}
+
+void HeapChunkSelector::remove_chunk_from_defrag_heap(const chunk_num_t chunkID) {
+    std::vector< VChunk > chunks;
+    chunks.reserve(m_defrag_heap.size());
```
zichanglai, line 163: size() should be called when lock is hold, right?  The rest part is fine to me.

JacksonYao287: ack

zichanglai (approved):
LG

src/lib/homestore_backend/heap_chunk_selector.cpp
```diff
@@ -130,11 +135,44 @@ csharedChunk HeapChunkSelector::select_specific_chunk(const chunk_num_t chunkID)
...
         avalableBlkCounter.fetch_sub(vchunk.available_blks());
+        remove_chunk_from_defrag_heap(vchunk.get_chunk_id());
     }
 
     return vchunk.get_internal_chunk();
 }
 
+// select_chunk_for_GC will only be called when GC is triggered, and will return the chunk with the most
+// defrag blocks
+csharedChunk HeapChunkSelector::select_chunk_for_GC() {
```
yamingk, line 146: nit: Suggest to rename to most_defrag_chunk(), so that we don't keep GC knowledge inside heapchunkselector if possible?

yamingk (approved):

PR #118, opened 2023-11-28
Make delete and put api idempotent in recovery case.

sanebay:
If put fails and returns existing valid pba, then ignore. If key not found for delete, then ignore. If key exists and already tomebestone, then ignore. If key exists and valid pba then do the free blks.

### DIFF
--- a/src/lib/homestore_backend/hs_blob_manager.cpp
+++ b/src/lib/homestore_backend/hs_blob_manager.cpp
@@ -363,14 +363,22 @@ void HSHomeObject::on_blob_del_commit(int64_t lsn, sisl::blob const& header, sis
 
     auto r = move_to_tombstone(index_table, blob_info);
     if (!r) {
-        LOGW("fail to move blob to tombstone,  blob_id {}, shard_id {}, {}", blob_info.blob_id, blob_info.shard_id,
-             r.error());
-        if (ctx) ctx->promise_.setValue(folly::makeUnexpected(r.error()));
-        return;
+        if (recovery_done_) {
+            LOGE("fail to move blob to tombstone,  blob_id {}, shard_id {}, {}", blob_info.blob_id, blob_info.shard_id,
+                 r.error());
+            if (ctx) ctx->promise_.setValue(folly::makeUnexpected(r.error()));
+            return;
+        } else {
+            if (ctx) { ctx->promise_.setValue(BlobManager::Result< BlobInfo >(blob_info)); }
+            return;
+        }
     }
 
     auto& multiBlks = r.value();
-    repl_dev->async_free_blks(lsn, multiBlks);
+    if (multiBlks != tombstone_pbas) {
+        repl_dev->async_free_blks(lsn, multiBlks);
+    }
+
     if (ctx) { ctx->promise_.setValue(BlobManager::Result< BlobInfo >(blob_info)); }
 }
 
--- a/src/lib/homestore_backend/hs_homeobject.cpp
+++ b/src/lib/homestore_backend/hs_homeobject.cpp
@@ -1,4 +1,3 @@
-
 #include <latch>
 #include <optional>
 #include <spdlog/fmt/bin_to_hex.h>
@@ -142,6 +141,7 @@ void HSHomeObject::init_homestore() {
         RELEASE_ASSERT(new_id == _our_id, "Received new SvcId [{}] AFTER recovery of [{}]?!", to_string(new_id),
                        to_string(_our_id));
     }
+    recovery_done_ = true;
     LOGI("Initialize and start HomeStore is successfully");
 }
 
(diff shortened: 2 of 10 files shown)


szmyd (commented):
is there a unit test we can add to cover these conditions?

sanebay:
> is there a unit test we can add to cover these conditions?

We need to replay the logs to simulate this. We can do it after raft repldev is done

src/lib/homestore_backend/index_kv.cpp
```diff
@@ -74,7 +81,7 @@ BlobManager::Result< homestore::MultiBlkId > HSHomeObject::move_to_tombstone(sha
     BlobRouteValue index_value_get;
     homestore::BtreeSingleGetRequest get_req{&index_key, &index_value_get};
     auto status = index_table->get(get_req);
-    if (status != homestore::btree_status_t::success) {
+    if (status != homestore::btree_status_t::success && recovery_done_) {
         LOGDEBUG("Failed to get from index table [route={}]", index_key);
```
yamingk, line 85: Is there a reason this is not a LOGERROR? 

Since the caller is also checking recovery_done_, can we keep this function out of the knowledge from recovery? maybe I didn't get the reason why both the caller and this function needs to check it.

src/lib/homestore_backend/hs_blob_manager.cpp
```diff
@@ -362,15 +362,18 @@ void HSHomeObject::on_blob_del_commit(int64_t lsn, sisl::blob const& header, sis
     blob_info.blob_id = *r_cast< blob_id_t* >(key.bytes);
 
     auto r = move_to_tombstone(index_table, blob_info);
-    if (!r) {
+    if (!r && recovery_done_) {
```
yamingk, line 365: Is it possible to verify it with solorepldev?

yamingk:
please update conan version

src/lib/homestore_backend/hs_homeobject.cpp
```diff
@@ -89,6 +88,7 @@ void HSHomeObject::init_homestore() {
         RELEASE_ASSERT(new_id == _our_id, "Received new SvcId [{}] AFTER recovery of [{}]?!", to_string(new_id),
                        to_string(_our_id));
     }
+    recovery_done_ = true;
```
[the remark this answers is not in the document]
sanebay: Confirmed with Hari. by the time start homestore is done. All logs are replayed.

src/lib/homestore_backend/hs_blob_manager.cpp
```diff
@@ -362,15 +362,18 @@ void HSHomeObject::on_blob_del_commit(int64_t lsn, sisl::blob const& header, sis
...
         LOGW("fail to move blob to tombstone,  blob_id {}, shard_id {}, {}", blob_info.blob_id, blob_info.shard_id,
              r.error());
         if (ctx) ctx->promise_.setValue(folly::makeUnexpected(r.error()));
         return;
     }
 
     auto& multiBlks = r.value();
```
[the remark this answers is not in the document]
sanebay: Ack. We shouldnt assert , just ignore the error.

src/lib/homestore_backend/index_kv.cpp
sanebay: We didnt want to print those errors during recovery. as we are ignoring them.

src/lib/homestore_backend/hs_blob_manager.cpp
sanebay: Currently we dont have a way to inject error.

sanebay:
> please update conan version

done.

src/lib/homestore_backend/hs_blob_manager.cpp
```diff
@@ -363,14 +363,22 @@ void HSHomeObject::on_blob_del_commit(int64_t lsn, sisl::blob const& header, sis
 
     auto r = move_to_tombstone(index_table, blob_info);
     if (!r) {
-        LOGW("fail to move blob to tombstone,  blob_id {}, shard_id {}, {}", blob_info.blob_id, blob_info.shard_id,
-             r.error());
-        if (ctx) ctx->promise_.setValue(folly::makeUnexpected(r.error()));
-        return;
+        if (recovery_done_) {
```
szmyd, line 366: Why is it an error if we are not in recovery, but success if we are? Shouldn't it just be success regardless? Maybe I don't understand exactly when `move_to_tombstone` could fail...

sanebay: We return error if delete failed to mark it as tombstone in usual rpc request. Only for recovery we can ignore the error. move_to_tombstone if there is any issue in accessing index or delete came again for a GCed blob etc. rare though.

szmyd: I guess I just want to make sure that deleting a BLOB in *any* case where a read would not be able to read the blob should return success. This includes when the blob never existed, or previously existed and no does not. As long as both those would be *success* on blob delete I'm okay with this...we should only fail when a subsequent read would return a blob.

raakella1 (approved):
LG!

src/lib/homestore_backend/hs_blob_manager.cpp
sanebay: If i understood you correctly, yeah a duplicate delete should be success

PR #120, opened 2023-12-03
SealShard state change in pre-commit

JacksonYao287:
in this PR, we change the shard state from open to sealed when pre_committing the seal_shard message, and change the state back to open if roll_back happens. 

but this can not fundamentally solve the conflict between seal_shard and put_blob, since when handling put_blob request, it will only check the shard state at the very beginning, and will not check again before entering raft phase(propose_to_raft). so we need a callback before appending the log where we could check the shard state again.

this should be done in homestore to provide the interfact accepting this callback

for some corner case , like we have a stale blob for shard-1 in shard-2, it will behandled by GC.

### DIFF
--- a/src/lib/homestore_backend/hs_shard_manager.cpp
+++ b/src/lib/homestore_backend/hs_shard_manager.cpp
@@ -191,7 +191,7 @@ ShardManager::AsyncResult< ShardInfo > HSHomeObject::_seal_shard(ShardInfo const
 
 // move seal_shard to pre_commit can not fundamentally solve the conflict between seal_shard and put_blob, since
 // put_blob handler will only check the shard state at the very beginning and will not check again before proposing to
-// raft, so we need a callback to check whether we can handle this requeest before appending log, which is previous to
+// raft, so we need a callback to check whether we can handle this request before appending log, which is previous to
 // pre_commit.
 
 // FIXME after we have the callback, which is coming in homestore.
--- a/src/lib/homestore_backend/hs_shard_manager.cpp
+++ b/src/lib/homestore_backend/hs_shard_manager.cpp
@@ -189,9 +189,25 @@ ShardManager::AsyncResult< ShardInfo > HSHomeObject::_seal_shard(ShardInfo const
     return req->result();
 }
 
+// move seal_shard to pre_commit can not fundamentally solve the conflict between seal_shard and put_blob, since
+// put_blob handler will only check the shard state at the very beginning and will not check again before proposing to
+// raft, so we need a callback to check whether we can handle this requeest before appending log, which is previous to
+// pre_commit.
+
+// FIXME after we have the callback, which is coming in homestore.
+
 bool HSHomeObject::on_shard_message_pre_commit(int64_t lsn, sisl::blob const& header, sisl::blob const& key,
-                                               cintrusive< homestore::repl_req_ctx >& ctx) {
+                                               cintrusive< homestore::repl_req_ctx >& hs_ctx) {
+    repl_result_ctx< ShardManager::Result< ShardInfo > >* ctx{nullptr};
+    if (hs_ctx && hs_ctx->is_proposer) {
+        ctx = boost::static_pointer_cast< repl_result_ctx< ShardManager::Result< ShardInfo > > >(hs_ctx).get();
+    }
     const ReplicationMessageHeader* msg_header = r_cast< const ReplicationMessageHeader* >(header.cbytes());
+    if (msg_header->corrupted()) {
+        LOGW("replication message header is corrupted with crc error, lsn:{}", lsn);
+        if (ctx) { ctx->promise_.setValue(folly::makeUnexpected(ShardError::CRC_MISMATCH)); }
+        return false;
+    }
     switch (msg_header->msg_type) {
     case ReplicationMessageType::SEAL_SHARD_MSG: {
         auto sb = r_cast< shard_info_superblk const* >(header.cbytes() + sizeof(ReplicationMessageHeader));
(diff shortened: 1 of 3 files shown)


src/lib/homestore_backend/heap_chunk_selector.cpp
```diff
@@ -26,6 +26,14 @@ void HeapChunkSelector::add_chunk_internal(const chunk_num_t chunkID) {
         return;
     }
     const auto& chunk = m_chunks[chunkID];
+
+    {
+        std::lock_guard< std::mutex > l(m_defrag_mtx);
+        if (m_selected_chunks.find(chunkID) != m_selected_chunks.end()) return;
+        m_defrag_heap.emplace(chunk);
+        m_selected_chunks.erase(chunkID);
```
zichanglai, line 34: we do not need L34 if we reach here.

src/lib/homestore_backend/heap_chunk_selector.h
```diff
@@ -70,6 +70,7 @@ class HeapChunkSelector : public homestore::ChunkSelector {
     std::unordered_map< chunk_num_t, csharedChunk > m_chunks;
 
     VChunkDefragHeap m_defrag_heap;
+    std::unordered_set< chunk_num_t > m_selected_chunks;
```
zichanglai, line 73: m_selected_chunks means chunk with  opening shard?  I do not see anyway to put element into it.

src/lib/homestore_backend/hs_shard_manager.cpp
```diff
@@ -142,6 +142,100 @@ ShardManager::AsyncResult< ShardInfo > HSHomeObject::_seal_shard(ShardInfo const
...
+
+    auto shard_info = deserialize_shard_info(r_cast< const char* >(value.bytes), value.size);
+    switch (header.msg_type) {
+    case ReplicationMessageType::SEAL_SHARD_MSG: {
+        // we can not release chunk here, since if rollback happens, we can not make sure we can get the same chunk.
+        // chunk selector will always return the a chunk of least used, and GC will happen at the time window
+        // we can not make sure we can get the same chunk as before, so we need to wait for the commit phase to release
+        // chunk;
+        update_shard_in_map(shard_info);
+        if (ctx) { ctx->promise_.setValue(ShardManager::Result< ShardInfo >(shard_info)); }
```
zichanglai, line 187: we can not set promise here, right? otherwise it will response client that seal shard is successfully, but maybe seal shard message is not replicated to quorum later.

JacksonYao287: I also considered this point earlier.
according to the design of nuraft , if we use pre_commit, we need to reply to client in pre_commit. pls refer to https://github.com/eBay/NuRaft/blob/9de4522beccaa1901d49488fcbb01447a9335ea9/docs/async_replication.md?plain=1#L17

it mentioned in nuraft `However, there are some loosened cases where we want to achieve low latency by sacrificing consistency and resolving conflicts manually`.

so if we use pre_commit to take the real action, there might be a case that client get a success response, but then the action is rollbacked. that means , if we adopt pre_commit, we need accept this conflict. that is the contract.

but,  move it to `on_commit` is also i want to do earlier. i write the code here to see if there is any discussion  about.

src/lib/homestore_backend/heap_chunk_selector.h
JacksonYao287: at  heap_chunk_selector.cpp line 178 , we should m_selected_chunks.emplace(chunkID), not erase, this is a mistake, will fix it.

src/lib/homestore_backend/hs_shard_manager.cpp
JacksonYao287: let`s move it to on_commit

src/lib/homestore_backend/heap_chunk_selector.cpp
JacksonYao287: we need this

src/lib/homestore_backend/hs_shard_manager.cpp
```diff
@@ -189,6 +189,90 @@ ShardManager::AsyncResult< ShardInfo > HSHomeObject::_seal_shard(ShardInfo const
     return req->result();
 }
 
+// move seal_shard to pre_commit can not fundamentally solve the conflict between seal_shard and put_blob, since
+// put_blob handler will only check the shard state at the very beginning and will not check again before proposing to
+// raft, so we need a callback to check whether we can handle this request before appending log, which is previous to
+// pre_commit.
```
xiaoxichen, line 195: prior to the pre_commit of put_blob

src/lib/homestore_backend/hs_shard_manager.cpp
```diff
@@ -189,6 +189,90 @@ ShardManager::AsyncResult< ShardInfo > HSHomeObject::_seal_shard(ShardInfo const
...
+
+// FIXME after we have the callback, which is coming in homestore.
```
xiaoxichen, line 197: an issue in HS and HO respectively, pls

xiaoxichen (approved):
lgtm

src/lib/homestore_backend/hs_shard_manager.cpp
JacksonYao287: sure

src/lib/homestore_backend/hs_shard_manager.cpp
JacksonYao287: thanks for pointing out this misusing , will be careful!

PR #126, opened 2023-12-19
Adapt raft replication dev  in homestore

zichanglai:
This PR will utilize the latest raft replication devs inside homestore to  construct a PG group which provides the capability to replicate blobs to other memers.


JacksonYao287:
reassign to jyao3

PR #129, opened 2024-01-04
major version upgrades to homestore and sisl. 

raakella1:
This PR introduces homestore ver 5 which has the raft based data replication


src/lib/tests/fixture_app.cpp
```diff
@@ -12,7 +12,7 @@ SISL_OPTION_GROUP(
     (num_shards, "", "num_shards", "number of shards", ::cxxopts::value< uint64_t >()->default_value("20"), "number"),
     (num_blobs, "", "num_blobs", "number of blobs", ::cxxopts::value< uint64_t >()->default_value("50"), "number"));
 
-SISL_LOGGING_INIT(logging, HOMEOBJECT_LOG_MODS)
+SISL_LOGGING_DEF(logging, HOMEOBJECT_LOG_MODS)
```
szmyd, line 15: `_DEF` is meant to be in a library translation unit, otherwise SM will have to do this too. I suggest `homobject_impl.cpp` (likely the heap test too since it doesn't link homeobject i think?)

raakella1: ACK, makes sense. I will change this to a library.

src/lib/homestore_backend/hs_homeobject.cpp
```diff
@@ -31,6 +33,54 @@ extern std::shared_ptr< HomeObject > init_homeobject(std::weak_ptr< HomeObjectAp
...
+class HSReplApplication : public homestore::ReplApplication {
+public:
+    HSReplApplication(homestore::repl_impl_type impl_type, bool need_timeline_consistency, HSHomeObject* home_object,
+                      std::shared_ptr< HomeObjectApplication > ho_application) :
+            _impl_type(impl_type),
+            _need_timeline_consistency(need_timeline_consistency),
+            _home_object(home_object),
+            _ho_application(ho_application) {}
+
+    virtual ~HSReplApplication() = default;
```
szmyd, line 46: This should be `override` correct?

szmyd (changes requested):

src/lib/tests/fixture_app.cpp
szmyd: Just for clarity:
`..._DECL` : Just to make `LOGMOD...(mod,` compile, can appear any number of times
`..._DEF` : Provide symbol to linker (once per module)
`..._INIT` : Once per executable; loads modules into the global map to make them settable.

.github/workflows/build_dependencies.yml
```diff
@@ -197,11 +190,11 @@ jobs:
         key_prefix: HomeObjectDeps-${{ inputs.platform }}-${{ inputs.build-type }}-${{ inputs.malloc-impl }}-${{ inputs.prerelease }}
       if: ${{ github.event_name != 'pull_request' && steps.restore-cache.outputs.cache-hit != 'true' }}
 
-    - name: Reload Sisl10 Cache
+    - name: Reload Sisl11 Cache
       uses: eBay/sisl/.github/actions/load_conan@stable/v8.x
       with:
         load_any: 'True'
-        key_prefix: Sisl10Deps-${{ inputs.platform }}-${{ inputs.build-type }}-${{ inputs.malloc-impl }}-${{ inputs.prerelease }}
+        key_prefix: Sisl11Deps-${{ inputs.platform }}-${{ inputs.build-type }}-${{ inputs.malloc-impl }}-${{ inputs.prerelease }}
```
szmyd, line 197: I think this needs to be just `SislDeps`

.github/workflows/build_commit.yml
```diff
@@ -55,7 +55,7 @@ jobs:
     if: ${{ github.event_name != 'pull_request' }}
 
   IOMgrDeps:
-    needs: [Sisl10Deps, Sisl11Deps]
```
szmyd, line 58: Can remove the block for this job above too.

PR #131, opened 2024-01-09
fix multipile definitions of logging modules in test package

raakella1:


xiaoxichen:
@raakella1  do we still need this?

PR #133, opened 2024-01-16
implement mem pg manager apis for SM unit tests

raakella1:


szmyd (changes requested):
We should cover this code from the tests right? Supposedly we are not for Homestore impl either; or the tests were not made in the generic `src/lib/tests` location?

raakella1:
> We should cover this code from the tests right? Supposedly we are not for Homestore impl either; or the tests were not made in the generic `src/lib/tests` location?

It is the later. I will add the get_stats test to the generic tests folder

szmyd (approved):
image

PR #136, opened 2024-01-25
replicate createpg message across raft group

JacksonYao287:
we need to replicate create PG message across the rafte group, so that the follower can create pginfo and index table.

### DIFF
--- a/src/include/homeobject/pg_manager.hpp
+++ b/src/include/homeobject/pg_manager.hpp
@@ -9,7 +9,8 @@
 
 namespace homeobject {
 
-ENUM(PGError, uint16_t, UNKNOWN = 1, INVALID_ARG, TIMEOUT, UNKNOWN_PG, UNKNOWN_PEER, UNSUPPORTED_OP);
+ENUM(PGError, uint16_t, UNKNOWN = 1, INVALID_ARG, TIMEOUT, UNKNOWN_PG, UNKNOWN_PEER, UNSUPPORTED_OP, PG_ALREADY_EXISTS,
+     CRC_MISMATCH);
 
 struct PGMember {
     explicit PGMember(peer_id_t _id) : id(_id) {}
--- a/src/include/homeobject/pg_manager.hpp
+++ b/src/include/homeobject/pg_manager.hpp
@@ -9,7 +9,8 @@
 
 namespace homeobject {
 
-ENUM(PGError, uint16_t, UNKNOWN = 1, INVALID_ARG, TIMEOUT, UNKNOWN_PG, UNKNOWN_PEER, UNSUPPORTED_OP, CRC_MISMATCH);
+ENUM(PGError, uint16_t, UNKNOWN = 1, INVALID_ARG, TIMEOUT, UNKNOWN_PG, UNKNOWN_PEER, UNSUPPORTED_OP, CRC_MISMATCH,
+     NO_SPACE_LEFT, DRIVE_WRITE_ERROR);
 
 struct PGMember {
     explicit PGMember(peer_id_t _id) : id(_id) {}
--- a/src/include/homeobject/pg_manager.hpp
+++ b/src/include/homeobject/pg_manager.hpp
@@ -9,8 +9,7 @@
 
 namespace homeobject {
 
-ENUM(PGError, uint16_t, UNKNOWN = 1, INVALID_ARG, TIMEOUT, UNKNOWN_PG, UNKNOWN_PEER, UNSUPPORTED_OP, PG_ALREADY_EXISTS,
-     CRC_MISMATCH);
+ENUM(PGError, uint16_t, UNKNOWN = 1, INVALID_ARG, TIMEOUT, UNKNOWN_PG, UNKNOWN_PEER, UNSUPPORTED_OP, CRC_MISMATCH);
 
 struct PGMember {
     explicit PGMember(peer_id_t _id) : id(_id) {}
--- a/src/lib/homestore_backend/hs_pg_manager.cpp
+++ b/src/lib/homestore_backend/hs_pg_manager.cpp
@@ -1,4 +1,5 @@
 #include <boost/uuid/random_generator.hpp>
+#include <boost/uuid/string_generator.hpp>
 #include <homestore/replication_service.hpp>
 #include "hs_homeobject.hpp"
 #include "replication_state_machine.hpp"
@@ -34,6 +35,12 @@ PGError toPgError(ReplServiceError const& e) {
         return PGError::TIMEOUT;
     case ReplServiceError::SERVER_NOT_FOUND:
         return PGError::UNKNOWN_PG;
+    /* TODO: enable this after add erro type to homestore
+    case ReplServiceError::ALREADY_EXISTS:
+        return PGError::PG_ALREADY_EXISTS;
+    case ReplServiceError::CRC_MISMATCH:
+        return PGError::CRC_MISMATCH;
+     */
     case ReplServiceError::OK:
         DEBUG_ASSERT(false, "Should not process OK!");
         [[fallthrough]];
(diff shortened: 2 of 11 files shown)


src/include/homeobject/pg_manager.hpp
```diff
@@ -9,7 +9,8 @@
 
 namespace homeobject {
 
-ENUM(PGError, uint16_t, UNKNOWN = 1, INVALID_ARG, TIMEOUT, UNKNOWN_PG, UNKNOWN_PEER, UNSUPPORTED_OP);
+ENUM(PGError, uint16_t, UNKNOWN = 1, INVALID_ARG, TIMEOUT, UNKNOWN_PG, UNKNOWN_PEER, UNSUPPORTED_OP, PG_ALREADY_EXISTS,
```
szmyd, line 12: We should not introduce an error for something that is intended to be idempotent. If a PG exists; just return SUCCESS? There may be instances where the CM sends this to one SM; it times out, but actually works, then sends it again either to the same SM or one of the other members. Should just be instant success.

src/lib/homestore_backend/hs_pg_manager.cpp
```diff
@@ -48,32 +55,138 @@ PGError toPgError(ReplServiceError const& e) {
 }
 
 PGManager::NullAsyncResult HSHomeObject::_create_pg(PGInfo&& pg_info, std::set< peer_id_t > const& peers) {
+    auto pg_id = pg_info.id;
+    {
```
szmyd, line 59: You can condense this to:
```c++
if (auto lg = std::shared_lock(_pg_lock); _pg_map.end() != _pg_map.find(pg_id))
    return folly::makeUnexpected(PGError::PG_ALREADY_EXISTS);
```

src/lib/homestore_backend/hs_pg_manager.cpp
```diff
@@ -48,32 +55,138 @@ PGError toPgError(ReplServiceError const& e) {
...
     pg_info.replica_set_uuid = boost::uuids::random_generator()();
     return hs_repl_service()
         .create_repl_dev(pg_info.replica_set_uuid, peers)
         .via(executor_)
-        .thenValue([this, pg_info = std::move(pg_info)](auto&& v) mutable -> PGManager::NullResult {
+        .thenValue([this, pg_info = std::move(pg_info)](auto&& v) mutable -> PGManager::NullAsyncResult {
             if (v.hasError()) { return folly::makeUnexpected(toPgError(v.error())); }
+            // we will write a PGHeader to the raft group and commit it when PGHeader is commited on raft
+            // group. so that all raft members will create the PGinfo and index table for this PG.
+            return do_create_pg(v.value(), std::move(pg_info));
```
szmyd, line 73: What do we do if this fails? We need to handle that and either offload it somewhere to retry until it works; or unwind this group optimistically so we don't leak a repl_dev.

src/lib/homestore_backend/hs_pg_manager.cpp
```diff
@@ -48,32 +55,138 @@ PGError toPgError(ReplServiceError const& e) {
...
+
+    req->header_.msg_type = ReplicationMessageType::CREATE_PG_MSG;
+    req->header_.pg_id = pg_info.id;
+    req->header_.shard_id = 0;
+    req->header_.payload_size = msg_size;
+    req->header_.payload_crc = crc32_ieee(init_crc32, buf_ptr, msg_size);
+    req->header_.seal();
+    sisl::blob header;
+    header.set_bytes(r_cast< uint8_t* >(&req->header_));
+    header.set_size(sizeof(req->header_));
```
szmyd, line 94: I know this may be the pattern so far; but I would suggest writing some free (non-member) functions in this source that just do type conversion for building requests and their headers; that way we can write unit tests around them without having to exercise the raft code.

src/lib/homestore_backend/hs_pg_manager.cpp
```diff
@@ -48,32 +55,138 @@ PGError toPgError(ReplServiceError const& e) {
...
+                                               homestore::MultiBlkId const& blkids, homestore::ReplDev* repl_dev,
+                                               cintrusive< homestore::repl_req_ctx >& hs_ctx) {
+    if (hs_ctx != nullptr) {
+        auto ctx = boost::static_pointer_cast< repl_result_ctx< PGManager::NullResult > >(hs_ctx);
+        do_create_pg_message_commit(lsn, *r_cast< ReplicationMessageHeader* >(const_cast< uint8_t* >(header.cbytes())),
+                                    blkids, ctx->hdr_buf_, hs_ctx);
+        return;
+    }
+
+    // when homeobject restarts and replay the committed log, hs_ctx will be nullptr
```
szmyd, line 117: I think this block of code is not exercised in our current tests according to codecov; we should add something for this and possibly the two corruption/crc cases below?

src/include/homeobject/pg_manager.hpp
JacksonYao287: make sense, will remove this error type

src/lib/homestore_backend/hs_pg_manager.cpp
JacksonYao287: got it ,will change the code style

src/lib/homestore_backend/hs_pg_manager.cpp
```diff
@@ -48,32 +55,138 @@ PGError toPgError(ReplServiceError const& e) {
...
+    sisl::sg_list value;
+    value.size = blkids.blk_count() * repl_dev->get_blk_size();
+    auto value_buf = iomanager.iobuf_alloc(512, value.size);
+    value.iovs.push_back(iovec{.iov_base = value_buf, .iov_len = value.size});
+    auto header_ptr = r_cast< const ReplicationMessageHeader* >(header.cbytes());
+    repl_dev->async_read(blkids, value, value.size)
+        .thenValue([this, lsn, msg_header = *header_ptr, blkids, value](auto&& err) mutable {
+            if (err) {
+                LOGW("failed to read data from homestore pba, lsn:{}", lsn);
```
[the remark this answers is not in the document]
JacksonYao287: no, if we come here, it is a log reapplicaition in recovery. so hs_ctx is null, and we can not get ctx

src/lib/homestore_backend/replication_state_machine.cpp
```diff
@@ -7,6 +7,10 @@ void ReplicationStateMachine::on_commit(int64_t lsn, const sisl::blob& header, c
     LOGI("applying raft log commit with lsn:{}", lsn);
     const ReplicationMessageHeader* msg_header = r_cast< const ReplicationMessageHeader* >(header.cbytes());
     switch (msg_header->msg_type) {
+    case ReplicationMessageType::CREATE_PG_MSG: {
```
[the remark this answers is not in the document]
JacksonYao287: will do

src/lib/homestore_backend/hs_pg_manager.cpp
JacksonYao287: good question!

i do not think we should retry, since if it crashes after `create_repl_dev` and before sending the createPG message, when restarting, the pginfo is lost, the repl_dev does not know which PG it should belongs.

my solution is 
1 if `do_create_pg` fails, we just delete the repl_dev(unwind this group). create_pg is not frequnent and not in io path, so even if this will be time-consuming, it is acceptable!

2 when restarting and after applying all the logs, get the repl_dev list from Homestore and iterate all the pgs to to find out if there is any repl_dev does not belong to any pg and delete this repl_dev.

if we do this , we need to add two apis to Homestore:
1 get_repl_dev_list(repl_dev_type t);
2 delete_repl_dev(uuid )

I thinks this can make sure we do not leak any repl_dev

cc @hkadayam

src/lib/homestore_backend/hs_pg_manager.cpp
JacksonYao287: yes, it will be exercised only when recovery. we need to wait until recovery is supported by homesotre

src/lib/homestore_backend/hs_pg_manager.cpp
JacksonYao287: will creat separate PR to move this part to a free function and add the corresponding UT

src/lib/homestore_backend/hs_pg_manager.cpp
xiaoxichen: <pls correct me if i am wrong>
seems like we put group_id in the repl_dev superblock, can that be used to retrieve the pginfo from HomeObject?

JacksonYao287: @xiaoxichen 
the `gorup_id` is a randomly generated uuid, and the `pg_id` is a uint_16.
but yes, we can  generate `uuid` from `uint_16`, and get `unit_16` from previously generated `uuid`, so that we get a natural map between `uuid` and `uint_16`.

the key point is that the pginfo originally exists in the `create_pg_message`, which need to be replicated across the repl_dev(raft group). if this message is not replicated to the majority of this group, then it is not committed. when restarts, the uncommitted message will probabaly be dropped，and repl_dev can not  get pginfo from that dropped message.

JacksonYao287:
I have manually  test this PR with tree replicas. it works as expected!

src/lib/homestore_backend/hs_pg_manager.cpp
xiaoxichen: I am trying to say if we can avoid replicating `pginfo` at all to avoid the 2PC in this PG. If we created the repl_dev we are all set, followers can construct `pginfo` from repl_dev. During recovery the `pginfo` also can be recovered by the sb of repl_dev/raft_config;

the 3 required fields we already have 
`members` which can get from raft cluster config,  the `replica_set_uuid` is `group_id`,  the only thing needed is `pg_id`.

As you said , we can 
1. encode pg_id into `replica_set_uuid` so we get `pg_id` out from `replicat_set_uuid`
2. put the pg_id into `_user_ctx` field of `libnuraft::cluster_config`, c.f [link](https://github.com/eBay/NuRaft/blob/master/include/libnuraft/cluster_config.hxx#L113-L114)
3. put the entire `serailized_pg_info` into `_user_ctx`

2/3 can be implemented as the cluster_config was loaded by calling `RaftReplDev::load_config()`.
```
struct PGInfo {
    explicit PGInfo(pg_id_t _id) : id(_id) {}
    pg_id_t id;
    mutable MemberSet members;
    peer_id_t replica_set_uuid;

    auto operator<=>(PGInfo const& rhs) const { return id <=> rhs.id; }
    auto operator==(PGInfo const& rhs) const { return id == rhs.id; }
};
```

JacksonYao287: if we do not do 2PC for creating PG, `on_commit` will not be called. how to trigger building pg superblk and index table? it is easy for leader since it has the future to check the completion of creating repl_dev. but how about follower?

we can add some callback in raft_repl_service(like `on_repl_dev_created`?)? but , when should this callback be triggered?  i think it is a little complex.  we need to handle a lot of corner case.
for example, if the first follower is added to the raft group , but fail to add the second follower. at this case, we should not create pg on the repl dev. but the callback(if we have) is also called at the first follower, since it is not aware of this.

xiaoxichen: yes you are right.  
The corner case is more on panic-on-rollback that when we trying to delete the repl dev.  The good thing in your approach , the linking from repl-dev to PG happened through raft proposal in second phase, so that partial created repl-dev will not be visible. So client(CM) can easily retry

szmyd: >1 if do_create_pg fails, we just delete the repl_dev(unwind this group). create_repl_dev is not frequnent and not in io path, so even if this will be time-consuming, it is acceptable!

If deleting the repl_dev means properly committing a `DROP_GROUP` (REMOVE_PG) message to the journal, then that can fail too. If it doesn't, you'll be stuck with some members that formed a raft group, were abandoned by their leader and now don't know what to do and are still in the group.

The only way to guarantee PGs are not leaked is to ask the CM "What PGs should I belong to?". There are many corner-cases when considering pg-migration, old SMs coming up that used to belong to still alive or deleted PGs but not in the group...etc.

xiaoxichen: @szmyd I think there are two cases, 

the more general one is like you said recycle a PG replica from a SM/HS, in the case of pg-migration etc.  In this case we need to do the graceful exit (drop_group) to ensure the cluster config is right. As you said the only way is checking with SM via API call (or, consume the PGMap from SM, similar).

The later one that @JacksonYao287 targeted in this PR, is more on the orphan rep-dev that doesnt linked to PG,  that rep-dev can have K (1<=K<=N) members but got failure in the first(4th actually) proposal with PGInfo. In this case as as the rep-dev has no data is written and the raft-group can be discard,  the tear down may not necessary go through graceful way but just every replica recycle on its own by calling `msg_service::leave_group` to shutdown raft and deleting the rep-dev? (needs to deletes the metablks from the raft group as well).

JacksonYao287: as a summary, there are two problems:
**1 HS side: how to handle create repl_dev failure(for example , fail to add any member to the raft group)?**
this is the responsibility of HS, we should not consider this in HO side.

**2 HO side: if we get a good repl_dev from HS, but we fail to create PG(for example, createPG message fail to be replicated to the majority of the raft group and is not committed, or crash happens at some moment ), how to do cleanup to avoid repl_dev leak?**

according to the meeting tomorrow, this will be done in phrase three and by that time we will have my new features in homestore and homeobject, which are probably  very helpful for solving this problem. so what about just leave the problem here for now (only happy path) and revisit it in phase 3.
@xiaoxichen @szmyd @sanebay

xiaoxichen: Yes lets create an issue and put a #FIXME  in the code

JacksonYao287: sure , done

src/lib/homestore_backend/hs_pg_manager.cpp
```diff
@@ -48,32 +53,131 @@ PGError toPgError(ReplServiceError const& e) {
...
+    req->header_.payload_size = msg_size;
+    req->header_.payload_crc = crc32_ieee(init_crc32, buf_ptr, msg_size);
+    req->header_.seal();
+    sisl::blob header;
+    header.set_bytes(r_cast< uint8_t* >(&req->header_));
+    header.set_size(sizeof(req->header_));
+
+    sisl::sg_list value;
+    value.size = msg_size;
+    value.iovs.push_back(iovec(buf_ptr, msg_size));
```
hkadayam, line 94: I am not sure this needs to be value and be written into BlkId etc.. It can be inline data with value being empty and write everything in the sisl::blob header itself. That avoids more code during recovery to read from local and handle its failure. The replay simply have all information about the value in the header itself.

src/lib/homestore_backend/hs_pg_manager.cpp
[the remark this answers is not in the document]
hkadayam: This is the code I referred to earlier, we don't need async_read(), if we write as much information as needed in header. This is true for other use cases to replicate (except shard create and blob put)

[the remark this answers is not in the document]
xiaoxichen: @hkadayam yes you said make sense.
is that makes sense if we can let small IO (< pagesize ? or even <16K etc) to through raft? most of the metadata ops like createpg/createshard etc can go through this .

```
 // If it is header only entry, directly propose to the raft
    if (rreq->value.size > MIN_DATACHANNEL_VALUESIZE) {
         // go through data channel
      } else {
         // propose through raft.
        RD_LOG(INFO, "Skipping data channel send since value size is 0");
        rreq->state.fetch_or(uint32_cast(repl_req_state_t::DATA_WRITTEN));
        m_state_machine->propose_to_raft(std::move(rreq));
    }
```

src/lib/homestore_backend/hs_pg_manager.cpp
JacksonYao287: when recovery, what we can get from the log are header, key and blkid.
key is used for consistent read. blkid is for the blk which the value is written to.

so your idea is we put the header and serailized_pg_info together into header blob, which will be writen to log and can be got back when recovery?

implementation wis , i think it`s good and simplify the process of recovery. let`s do this.

src/lib/homestore_backend/hs_pg_manager.cpp
szmyd: @xiaoxichen yes i know; i brought this up initially. Actually; there is one problem that is the generalized form of all problems:

- I (SM) belong to a raft group (repl_dev) that I am not managing a PG for according to truth (CM).

All issues regarding repl_dev failure, pg_creation failure, migration during offline member (orphaned pg)...they generalize to this. The solution is what I suggest, anything else is an optimization for a specialization of this issue, and need not take the risk of complicating machinery just to correct something that is not in the immediate necessary to correct.

This is from years of fighting the raft membership issues with AM; we do _try_ to cleanup in the cases that we can, but don't worry ourselves about it as the global cleanup routine will correct things. We should just do that here to start so things are _correct_ and we can asses if there are timing issues that warrant an optimization beyond fire-and-forget cleanup.

xiaoxichen: @szmyd yes you said makes perfect sense. The raft membership change in SM would be less than AM but still reconcile based on CM's PGMap is the right path. There are some timing issue we need to be careful though.

PR #143, opened 2024-02-13
Buffer management of all shard create, put blob, get blob calls

hkadayam:
* Improved buffer management which fixes the issue of failed writes as well as more efficient allocation.

* Buffer safety by using io_blob_safe instead of raw alloc and free.

* Improved module logging to make it easy to visualize on blob put/get


src/lib/homestore_backend/hs_homeobject.hpp
```diff
@@ -93,15 +96,19 @@ class HSHomeObject : public HomeObjectImpl {
...
+    struct DataHeader {
+        static constexpr uint8_t data_header_version = 0x01;
+        static constexpr uint64_t data_header_magic = 0x21fdffdba8d68fc6; // echo "BlobHeader" | md5sum
+
+        enum class data_type_t : uint32_t { SHARD_INFO = 1, BLOB_INFO = 2 };
+
+        uint64_t magic{data_header_magic};
+        uint8_t version{data_header_version};
+        data_type_t type{data_type_t::BLOB_INFO};
+    };
```
hkadayam, line 108: Created the DataHeader common, so that both blobheader and shardheader have common DataHeader information, which is essential while doing GC.

src/lib/homestore_backend/hs_shard_manager.cpp
```diff
@@ -116,66 +127,120 @@ ShardManager::AsyncResult< ShardInfo > HSHomeObject::_seal_shard(ShardInfo const
...
-    if (hs_ctx != nullptr) {
-        do_shard_message_commit(lsn, *r_cast< ReplicationMessageHeader* >(const_cast< uint8_t* >(header.cbytes())),
-                                blkids, hs_ctx->key, hs_ctx);
+    if (crc32_ieee(init_crc32, value.cbytes(), value.size()) != header->payload_crc) {
+        // header & value is inconsistent;
+        LOGW("replication message header is inconsistent with value, lsn:{}", lsn);
+        if (ctx) { ctx->promise_.setValue(folly::makeUnexpected(ShardError::CRC_MISMATCH)); }
         return;
     }
+#endif
```
hkadayam, line 193: Have this option to do an optional validation of every commit to see if actual data written is same as whats there in journal. We should perhaps make this as run-time option enabled in may be debug build?

src/lib/homestore_backend/hs_homeobject.hpp
```diff
@@ -93,15 +96,19 @@ class HSHomeObject : public HomeObjectImpl {
...
+
+    struct shard_info_superblk : public DataHeader {
+        ShardInfo info;
```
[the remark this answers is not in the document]
szmyd: Should we encode some kind of versioning to these serialized types to ease upgradability post-GA?

src/lib/homestore_backend/hs_shard_manager.cpp
szmyd: The normal use case would be to only enable it _after_ some inconsistency was detected right? Do we have an analogous "safe mode" yet?

src/lib/homestore_backend/replication_state_machine.hpp
```diff
@@ -12,14 +12,57 @@ using homestore::repl_req_ctx;
 using homestore::ReplServiceError;
 
 struct ho_repl_ctx : public homestore::repl_req_ctx {
-    ReplicationMessageHeader header_;
     sisl::io_blob_safe hdr_buf_;
+    sisl::io_blob_safe key_buf_;
```
sanebay, line 16: comment will be helpful that data_bufs_ are kept so data_sgs_ buffers dont get destroyed.

src/lib/homestore_backend/hs_blob_manager.cpp
```diff
@@ -34,122 +67,82 @@ BlobManager::AsyncResult< blob_id_t > HSHomeObject::_put_blob(ShardInfo const& s
...
-    blob_header->shard_id = shard.id;
-    blob_header->blob_id = new_blob_id;
-    blob_header->hash_algorithm = BlobHeader::HashAlgorithm::CRC32;
-    blob_header->blob_size = blob.body.size();
-    blob_header->user_key_size = blob.user_key.size();
-    blob_header->object_offset = blob.object_off;
-    sgs.iovs.emplace_back(iovec{.iov_base = blob_header, .iov_len = blob_header_size});
-    sgs.size += blob_header_size;
+    // Create a put_blob request which allocates for header, key and blob_header. Data sgs are added later
+    auto req = put_blob_req_ctx::make(sizeof(BlobHeader) + blob.user_key.size());
```
sanebay, line 71: comment that memory for blob user key is stored as part of req_ctx.

src/lib/homestore_backend/hs_homeobject.hpp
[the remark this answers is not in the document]
hkadayam: Its base class `DataHeader` has versioning and magic encoded. Is that what you were looking for?

src/lib/homestore_backend/hs_homeobject.hpp
```diff
@@ -158,15 +165,13 @@ class HSHomeObject : public HomeObjectImpl {
...
-        static ShardInfo shard_info_from_sb(homestore::superblk< shard_info_superblk > const& sb);
     };
 
 #pragma pack(1)
     // Every blob payload stored in disk as blob header | blob data | blob metadata(optional) | padding.
     // Padding of zeroes is added to make sure the whole payload be aligned to device block size.
-    struct BlobHeader {
+    struct BlobHeader : public DataHeader {
         static constexpr uint64_t blob_max_hash_len = 32;
         static constexpr uint8_t blob_header_version = 0x01;
```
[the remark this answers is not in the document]
hkadayam: I can remove that.

src/lib/homestore_backend/tests/hs_shard_tests.cpp
```diff
@@ -72,6 +72,7 @@ TEST_F(TestFixture, CreateMultiShardsOnMultiPG) {
     }
 }
 
+#if 0
```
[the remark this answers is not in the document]
hkadayam: Couple of reasons, which I can get clarified here:
1. What is this test case doing which is not done by actual seal_shard?
2. It has internals of SealShard implementation duplicated here. Is that what we want for "HS" based tests?

src/lib/homestore_backend/hs_shard_manager.cpp
hkadayam: Yes we will enable this dynamic config after some inconsistency. Its also ok to do the same for all debug builds as well. Its good question to ask '_should we tie this config to check as part of "safe mode"_'. Perhaps I can create an issue to track that.

raakella1 (approved):
LG! Lets run the SM level tests to find bugs

PR #145, opened 2024-02-20
enable restartability for homeobject

JacksonYao287:
since we have restartability in homestore, enable the restartability in homeobject now

in the current implementation , when restarting, repl_dev will not be loaded when restarting meta service, it will be postponed to the the start of raft_repl_service. but when restarting, Pg is recovered when restarting metaservice , which will get the repl_dev from raft_repl_service, this causes an error.

### DIFF
--- a/conanfile.py
+++ b/conanfile.py
@@ -9,7 +9,7 @@ required_conan_version = ">=1.50.0"
 
 class HomeObjectConan(ConanFile):
     name = "homeobject"
-    version = "1.0.6"
+    version = "1.0.7"
     homepage = "https://github.com/eBay/HomeObject"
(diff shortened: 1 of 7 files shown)


JacksonYao287:
UT will succeed after https://github.com/eBay/HomeStore/pull/333 is merged

xiaoxichen:
This PR is right in general ,  but I have concern regarding caching all the shard info in memory during recovery as the # of shard is technically unbound, unless GC will recycle empty shard.

xiaoxichen:
based on the discussion today, we will merge this PR to unblock the test.  In the meantime we will rework this to defer the recover of PG/Shard after HS recover.

xiaoxichen (approved):

PR #152, opened 2024-03-06
rework recovery

JacksonYao287:
recovery pg and shard after homestore is fully restarted

this io_uring problem is tracked [in this issue](https://github.com/eBay/HomeObject/issues/153)

### DIFF
--- a/src/lib/homestore_backend/tests/hs_shard_tests.cpp
+++ b/src/lib/homestore_backend/tests/hs_shard_tests.cpp
@@ -125,41 +125,16 @@ TEST_F(TestFixture, MockSealShard) {
 }
 #endif
 
-class FixtureAppWithRecovery : public FixtureApp {
-    std::string fpath_{"/tmp/test_shard_manager.data." + std::to_string(rand())};
-
-public:
-    std::list< std::filesystem::path > devices() const override {
-        auto device_info = std::list< std::filesystem::path >();
-        device_info.emplace_back(std::filesystem::canonical(fpath_));
-        return device_info;
-    }
-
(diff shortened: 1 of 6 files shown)


xiaoxichen (dismissed):
lgtm

src/lib/homestore_backend/tests/hs_shard_tests.cpp
```diff
@@ -162,6 +162,7 @@ TEST_F(ShardManagerTestingRecovery, ShardManagerRecovery) {
     _home_object.reset();
     LOGI("restart home_object");
     _home_object = homeobject::init_homeobject(std::weak_ptr< homeobject::HomeObjectApplication >(app));
+    std::this_thread::sleep_for(std::chrono::seconds{5});
```
xiaoxichen, line 165: could we use `pg->repl_dev_->get_leader_id()` to check in L190 and sleep if not the leader?

JacksonYao287: I use 5s because the election_time_out is 3.2s. https://github.com/eBay/nuraft_mesg/blob/7e8cdeb261270642dac186902351da2174d7f077/src/lib/manager_impl.cpp#L25C1-L25C72

i want to refactor this two test case here , since they are much similiar and  have a lot of duplicated logic as
https://github.com/eBay/HomeObject/blob/a0df650e0cfb6391cd0dd6816175bcc8b284e944/src/lib/homestore_backend/tests/hs_blob_tests.cpp#L114

can we do this in a separate refactor PR?

xiaoxichen (approved):
sure

PR #154, opened 2024-03-11
Implement on_error() callback for statemachine.

xiaoxichen:
on_error() will only be called on where it proposed, i.e is_proposer == true.  Set the error back to future to notifiy caller.


src/lib/homestore_backend/replication_state_machine.cpp
```diff
@@ -46,8 +46,50 @@ void ReplicationStateMachine::on_rollback(int64_t lsn, sisl::blob const&, sisl::
...
                                        cintrusive< repl_req_ctx >& ctx) {
-    LOGE("on_error  with :{}, lsn {}", error, ctx->lsn);
-    // TODO:: block is already freeed at homestore side, handle error if necessay.
+
+    RELEASE_ASSERT(ctx->is_proposer, "on_error should only be called from proposer");
+    const ReplicationMessageHeader* msg_header = r_cast< const ReplicationMessageHeader* >(header.cbytes());
+    LOGE("on_error, message type {} with lsn {}, error {}", msg_header->msg_type, ctx->lsn, error);
+    switch (msg_header->msg_type) {
+    case ReplicationMessageType::CREATE_PG_MSG: {
+        if (ctx) {
```
JacksonYao287, line 55: we do not need `if (ctx)` here, since on_error will never be call in recovery mode and will always be called in leader.
on the other hand,  before `ctx->is_proposer` , we should make sure ctx is not nullptr

so , remove all the `if(ctx)` here and add a line before line 50:
`RELEASE_ASSERT(ctx , "ctx should not be nullptr in on_error");`

src/lib/homestore_backend/replication_state_machine.cpp
```diff
@@ -46,8 +46,50 @@ void ReplicationStateMachine::on_rollback(int64_t lsn, sisl::blob const&, sisl::
...
+    case ReplicationMessageType::DEL_BLOB_MSG:
+        if (ctx) {
+            auto result_ctx =
+                boost::static_pointer_cast< repl_result_ctx< BlobManager::Result< HSHomeObject::BlobInfo > > >(ctx)
+                    .get();
+            result_ctx->promise_.setValue(folly::makeUnexpected(toBlobError(error)));
+        }
+        break;
+    default: {
+        LOGE("Unknown message type, error unhandled , error :{}, lsn {}", error, ctx->lsn);
```
JacksonYao287, line 89: can we add a `UnknowMessageTypeError` for this case?
if we do nothing here, it will still block, right?  because `result_ctx->promise_.setValue` is not called

JacksonYao287 (commented):
I have two small comments, PTAL.

other parts look fine to me

src/lib/homestore_backend/replication_state_machine.cpp
xiaoxichen: for unknown message you dont know the type of the result_ctx, not to say setValue

src/lib/homestore_backend/replication_state_machine.cpp
xiaoxichen: thanks, addressed

src/lib/homestore_backend/replication_state_machine.cpp
JacksonYao287: IMO , if we can not handle this,  it is better to terminate the process here rather than leaving a stuck one , since the crash incident can be captured by our monitor or something else , which can at least give us an alert to look into what happens.

PR #159, opened 2024-03-27
Use homestore 6.x

hkadayam:
This PR provides features
* To use homestore 6.x which supports vetoing the shard create or blob put when PG or shard is not ready respectively.

* HomeObject Metrics and also durable entities to ensure that blob_id, count etc are all made durable and loaded after restart

* Start HttpServer and provide metrics, objlife counter, settings related endpoints


src/include/homeobject/blob_manager.hpp
```diff
@@ -10,7 +10,7 @@
 namespace homeobject {
 
 ENUM(BlobError, uint16_t, UNKNOWN = 1, TIMEOUT, INVALID_ARG, UNSUPPORTED_OP, NOT_LEADER, REPLICATION_ERROR,
-     UNKNOWN_SHARD, UNKNOWN_BLOB, CHECKSUM_MISMATCH, READ_FAILED, INDEX_ERROR, SEALED_SHARD);
+     UNKNOWN_SHARD, UNKNOWN_BLOB, CHECKSUM_MISMATCH, READ_FAILED, INDEX_ERROR, SEALED_SHARD, DATA_EXIST_ALREADY);
```
szmyd, line 13: It's unfortunate that we need to update the public API for an ec that will never leave the library itself... Not sure how to avoid this other than re-use an existing ec incorrectly.../shrug

src/lib/homestore_backend/hs_cp_callbacks.cpp
```diff
@@ -0,0 +1,59 @@
...
+folly::Future< bool > HSHomeObject::MyCPCallbacks::cp_flush(CP* cp) {
+    std::vector< HSHomeObject::HS_PG* > dirty_pg_list;
+    dirty_pg_list.reserve(home_obj_._pg_map.size());
+    {
+        std::shared_lock lock_guard(home_obj_._pg_lock);
+        for (auto const& [id, pg] : home_obj_._pg_map) {
+            auto hs_pg = static_cast< HSHomeObject::HS_PG* >(pg.get());
+
+            // All dirty durable entries are updated in the superblk. We persist outside the pg_lock
+            if (hs_pg->is_dirty_.load()) {
```
szmyd, line 39:
```c++
if (hs_pg->is_dirty_.exchange(false)) {
   hs_pg->pg_sb_->blob_sequence_num = hs_pg->durable_entities_.blob_sequence_num.load();
   ...
```

src/lib/homestore_backend/tests/homeobj_fixture.hpp
```diff
@@ -138,6 +137,19 @@ class HomeObjectFixture : public ::testing::Test {
         }
     }
 
+    void verify_obj_count(uint32_t num_pgs, uint32_t shards_per_pg, uint32_t blobs_per_shard,
+                          bool deleted_all = false) {
+        uint32_t exp_active_blobs = deleted_all ? 0 : shards_per_pg * blobs_per_shard;
+        uint32_t exp_tombstone_blobs = deleted_all ? shards_per_pg * blobs_per_shard : 0;
+
+        for (uint32_t i = 1; i <= num_pgs; i++) {
```
szmyd, line 145: `++i`? probably no-op...

szmyd (changes requested):

src/lib/homestore_backend/tests/homeobj_fixture.hpp
hkadayam: I use ++i always, wondered why I did i++ this time, but remembered used github copilot to auto fill which for some reason uses i++ (may be seen previous lines somewhere) :-)

src/lib/homestore_backend/hs_cp_callbacks.cpp
hkadayam: sure will do that

src/lib/homestore_backend/hs_cp_callbacks.cpp
```diff
@@ -0,0 +1,59 @@
...
+
+std::unique_ptr< CPContext > HSHomeObject::MyCPCallbacks::on_switchover_cp(CP* cur_cp, CP* new_cp) { return nullptr; }
+
+// when cp_flush is called, it means that all the dirty candidates are already in the dirty list.
+// new dirty candidates will arrive on next cp's context.
+folly::Future< bool > HSHomeObject::MyCPCallbacks::cp_flush(CP* cp) {
+    std::vector< HSHomeObject::HS_PG* > dirty_pg_list;
+    dirty_pg_list.reserve(home_obj_._pg_map.size());
+    {
+        std::shared_lock lock_guard(home_obj_._pg_lock);
```
yamingk, line 34: Is there a reason you choose to hold this _pg_lock? The original idea is cp_flush won't need to take this lock (because BlobManager is going to take the lock for every blob put) so that no to block I/O during cp_flush. Or do you intend not to take that lock during blob put?

src/lib/homestore_backend/hs_hmobj_cp.cpp
```diff
@@ -1,84 +0,0 @@
...
-    return folly::makeFuture< bool >(true);
-}
-
-void HomeObjCPCallbacks::cp_cleanup(CP* cp) {}
-
-int HomeObjCPCallbacks::cp_progress_percent() { return 0; }
-
-HomeObjCPContext::HomeObjCPContext(CP* cp) : CPContext(cp) { pg_dirty_list_.clear(); }
-
-void HomeObjCPContext::add_pg_to_dirty_list(HSHomeObject::pg_info_superblk* pg_sb) {
```
yamingk, line 67: to assist above comment, I probably didn't get why we remove this, the dirty buffer is there because we don't want to hold the _pg_lock during cp_flush.

src/lib/homestore_backend/hs_cp_callbacks.cpp
hkadayam: We are not taking lock while during /O during cp_flush, Under the lock, it only takes the durables, update the in-mem version of the pg_sb and then unlocks and persist them. In the critical data path, we could avoid putting the pg into an unordered_map which we need to put under the lock. At present, we just update each PG's atomics and then gather all PGs metadata during cp_flush.

src/lib/homestore_backend/hs_hmobj_cp.cpp
hkadayam: As mentioned we are not maintaining the changes as a separate buffer, but we are maintaining in the PG structure itself and while CP, we copy that to the superblk and persist. We are not persisting under any lock.

src/lib/homestore_backend/hs_cp_callbacks.cpp
szmyd: c++20 has initialize for ranged-for loops
```c++
for (auto lock_guard = shared_lock(home_obj_._pg_lock); auto const& [id, pg] ...) {
   ...
}
```

xiaoxichen (commented):
lgtm other than the discussion regarding http service, which can be done later

src/lib/homestore_backend/tests/homeobj_cp_tests.cpp
```diff
@@ -1,5 +1,6 @@
 #include "homeobj_fixture.hpp"
 
+#if 0
```
yamingk, line 3: Why this is disabled? is it because same scenario is already covered by other test cases already?

src/lib/homestore_backend/hs_blob_manager.cpp
```diff
@@ -409,7 +415,12 @@ void HSHomeObject::on_blob_del_commit(int64_t lsn, sisl::blob const& header, sis
     }
 
     auto& multiBlks = r.value();
-    if (multiBlks != tombstone_pbas) { repl_dev->async_free_blks(lsn, multiBlks); }
+    if (multiBlks != tombstone_pbas) {
+        repl_dev->async_free_blks(lsn, multiBlks);
+        hs_pg->durable_entities_.active_blob_count.fetch_sub(1, std::memory_order_relaxed);
```
yamingk, line 420: Does it make sense to make these three lines in a single function, say update_durable_ent() and inside this funciton hs_pg is set to dirty? The point is to give api to give a hint to others that this should not be accidentally updated without updating the dirty flag.

src/lib/homestore_backend/hs_cp_callbacks.cpp
yamingk: That lock can be removed actually. Personally I think we should keep cp out of the hosting consumers as much as possible. People who works on PG (HS_PG) doesn't need to look into what is meaning of hs_pg->pg_sb_, to understand it is actually only touched by cp routine. So why not keep anything need to be flushed in a dedicated class and let that class do the dirty buff flush? 

With your changes here, we can move hs_pg->pg_sb_ into HomeObjectCPContext, right?

src/lib/homestore_backend/hs_homeobject.hpp
```diff
@@ -23,6 +23,7 @@ class BlobRouteKey;
 class BlobRouteValue;
 using BlobIndexTable = homestore::IndexTable< BlobRouteKey, BlobRouteValue >;
 class HomeObjCPContext;
```
yamingk, line 25: This can be removed (though I suggest to keep it and move pg_sb_ into here)

src/lib/homestore_backend/tests/homeobj_cp_tests.cpp
hkadayam: Yes we are handling this check as part of other test.

src/lib/homestore_backend/hs_shard_manager.cpp
```diff
@@ -325,22 +325,22 @@ std::optional< homestore::chunk_num_t > HSHomeObject::get_shard_chunk(shard_id_t
...
-        return pg->any_allocated_chunk_id_;
+    if (pg->any_allocated_chunk_id_.has_value()) { // it is already cached and use it;
+        return {true /* pg_found */, true /* shards_found */, *pg->any_allocated_chunk_id_};
     }
 
     auto& shards = pg->shards_;
-    if (shards.empty()) { return std::nullopt; }
+    if (shards.empty()) { return {true /* pg_found */, false /* shards_found */, 0 /* chunk_id */}; }
+
     auto hs_shard = d_cast< HS_Shard* >(shards.front().get());
```
JacksonYao287, line 341: this function is used to select a chunk to put shard_info_super_blk when creating a shard. not quite understand why to select the chunk of the front shard in the pg shard list. should we select a chunk through chunkselector for a newly created shard?

src/lib/homestore_backend/hs_blob_manager.cpp
hkadayam: I have protected the access in a separate method and allow updates to happen later.

src/include/homeobject/blob_manager.hpp
hkadayam: I see your point, let me change the signature of add_index_table() to return internal status and return the actual error as INDEX_ERROR and remove the DATA_EXIST_ALREADY.

src/lib/homestore_backend/hs_cp_callbacks.cpp
yamingk: this is a minor comment above, I am okay to merge as it is. 
If you still are considering doing this, a requirement from HomeStore (as we discussed yesterday) is we need to allow this pg_sb_ to be persistent outside of cp_flush as well.

yamingk:
LGTM

szmyd (approved):

src/lib/homestore_backend/hs_shard_manager.cpp
JacksonYao287: ok, i think we need to revisit the current logic here. i doubt this probably casued the io_uring error we met before when creating shard.

src/lib/homestore_backend/hs_cp_callbacks.cpp
```diff
@@ -0,0 +1,58 @@
...
+            if (!hs_pg->is_dirty_.exchange(false)) { continue; }
+
+            hs_pg->pg_sb_->blob_sequence_num = hs_pg->durable_entities().blob_sequence_num.load();
+            hs_pg->pg_sb_->active_blob_count = hs_pg->durable_entities().active_blob_count.load();
+            hs_pg->pg_sb_->tombstone_blob_count = hs_pg->durable_entities().tombstone_blob_count.load();
+            dirty_pg_list.push_back(hs_pg);
+        }
+    }
+
+    for (auto& hs_pg : dirty_pg_list) {
```
JacksonYao287, line 48: soryy, a stupid question, if sm crashes before dirty_pg_list is flushed, how to ensure we can recover the correct durable_entities, espectially blob_sequence_num

hkadayam: If SM crashes, without CP, we also don't commit last_commit_idx in Raft which also part of CP. So those entries are provided by the leader again and go through commit again. That time, entities are updated including blob_sequence_number.

PR #164, opened 2024-04-08
Remove http_server start logic

yuwmao:
Http_server should be started in SM, otherwise sm will crash due to setup twice.

### DIFF
--- a/conanfile.py
+++ b/conanfile.py
@@ -9,7 +9,7 @@ required_conan_version = ">=1.50.0"
 
 class HomeObjectConan(ConanFile):
     name = "homeobject"
-    version = "1.0.13"
+    version = "1.0.14"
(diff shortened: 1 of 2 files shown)


yuwmao:
@xiaoxichen @hkadayam

zhiteng:
The start of HTTP server in HO is redundant/unnecessary as SM already did that.

hkadayam (dismissed):
Please update the conanfile.py version on every PR.

yuwmao:
> Please update the conanfile.py version on every PR.

Sure, updated.

xiaoxichen:
I am not sure about why the  regression in `HomeObjectFixture.BasicPutGetDelBlobWRestart`  get on and off, but as far as it has nothing to do with this pr, lets merge it

xiaoxichen (approved):

PR #165, opened 2024-04-09
Reject put_blob with zero length.

xiaoxichen:
Put_blob takes `Blob&& blob` which will make the provided blob into an empty one.

In previous code we introduced retry logic for put_blob however didnt properly recreate the blob buffer which mistakenly create an empty blob with size=0.

As we only retry on not_leader and this error is not local-retryable, we dont have a case that need local retry now. It still make sense to take right value ref for blob.  Caller (now only the UT) should take responsibility to copy before call, if retry is intended.

Rejecting zero-length put will better help user to identify the error.

### DIFF
--- a/conanfile.py
+++ b/conanfile.py
@@ -9,7 +9,7 @@ required_conan_version = ">=1.50.0"
 
 class HomeObjectConan(ConanFile):
     name = "homeobject"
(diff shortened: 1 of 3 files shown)


JacksonYao287 (approved):
LGTM

PR #168, opened 2024-04-12
Use pg_id as metrics instance name

yuwmao:


JacksonYao287 (approved):
LGTM. we can use the only label to show pg id for now, and add uuid back after the support of sisl

PR #170, opened 2024-04-19
flatbuffer for baseline resync

yamingk:
1. define flatbuffer structure for pg/shard creation during base line resync
2. define blob data flatbuffer structure during blob data resync
3. add a mock api to pick up homestore 6.3


src/lib/homestore_backend/CMakeLists.txt
```diff
@@ -4,6 +4,13 @@ if(NOT ${CMAKE_CURRENT_BINARY_DIR} STREQUAL ${CMAKE_CURRENT_SOURCE_DIR})
   include_directories(BEFORE ${CMAKE_CURRENT_BINARY_DIR})
 endif()
 
+#list(APPEND SCHEMA_FLAGS "--scoped-enums" "--gen-name-strings" "--cpp-std=c++17" "--cpp-static-reflection" "--reflect-names")
```
yamingk, line 7: this will be removed

src/lib/homestore_backend/replication_state_machine.hpp
```diff
@@ -166,6 +166,12 @@ class ReplicationStateMachine : public homestore::ReplDevListener {
     /// @brief Called when the replica set is being stopped
     void on_replica_stop() override;
 
+    /// @brief Called when the snapshot is being created.
+    /// @param snapshot - The snapshot object which contains the lsn and term;
+    void create_snapshot(homestore::repl_snapshot&) override {
```
yamingk, line 171: This is a place holder to allow HomeObject to be able to compile with HomeStore 6.3

src/lib/homestore_backend/resync_blob_data.fbs
```diff
@@ -0,0 +1,14 @@
+namespace homeobject;
+
+table BlobData {
+    shard_id : uint64;
+    key : [ubyte];
```
sanebay, line 5: Is this the user_key ? Then can rename it s/key/user_key/

src/lib/homestore_backend/resync_blob_data.fbs
```diff
@@ -0,0 +1,14 @@
...
+    size : uint32;
+    data : [ubyte];
```
sanebay, line 7: We also need blob id ?

src/lib/homestore_backend/resync_pg_shard.fbs
```diff
@@ -0,0 +1,25 @@
+namespace homeobject;
+
+table Member {
+   uuid : [ubyte];
+   name : [ubyte];
+}
+
+table PGInfoEntry {
```
sanebay, line 8: Should we reuse PG.descriptor and shard from nuobject proto ? Also the root table if it has a type whether its a blob or shardpg info ? Or we can assume that if obj_id = 0 its PGShardInfo and others will be BlobDataBatch

yamingk: yes, that is the design we discussed by reading the obj_id as you mentioned above.

src/lib/homestore_backend/resync_blob_data.fbs
yamingk: done

src/lib/homestore_backend/resync_pg_shard.fbs
```diff
@@ -0,0 +1,25 @@
+namespace homeobject;
+
+table Member {
+   uuid : [ubyte];
+   name : [ubyte];
+}
```
sanebay, line 6: priority too.

src/lib/homestore_backend/resync_pg_shard.fbs
```diff
@@ -0,0 +1,25 @@
...
+
+table PGInfoEntry {
+    pg_id : uint32;                     // only low 16 bit is used for pg_id;
+    replica_set_uuid : [ubyte];         // uuid of replica set
+    members : [Member];                 // peers;
```
sanebay, line 11: we need to add blob sequence number here or home object has to update , probably its on commit in its incremental resync.

src/lib/homestore_backend/resync_blob_data.fbs
```diff
@@ -0,0 +1,16 @@
+namespace homeobject;
+
+table BlobData {
+    shard_id : uint64;
```
sanebay, line 4: nit: may be uint32 can be together so that they are packed. shard_id, blob_id, data_size, user_key_size, data, user_key

src/lib/homestore_backend/resync_pg_shard.fbs
```diff
@@ -0,0 +1,25 @@
...
+}
+
+table ShardInfoEntry {
+    pg_id : uint32;                     // pg id which this shard belongs to;
```
sanebay, line 15: Do we need all these whats there in shard_info_superblk ?
   State state;
    uint64_t created_time;
    uint64_t last_modified_time;
    uint64_t available_capacity_bytes;
    uint64_t total_capacity_bytes;
    uint64_t deleted_capacity_bytes;
    std::optional< peer_id_t > current_leader{std::nullopt};

src/lib/homestore_backend/resync_pg_shard.fbs
yamingk: Looks like no one is populating the priority and it is just a dummy field. Is it still needed? BTW, priority is not part of create-pg input param.

src/lib/homestore_backend/resync_pg_shard.fbs
yamingk: Isn't blob id (same as blob_seq_num) already included in BlobData already, it will update receiver's PG sb for every blob put? 
Why do we need another blob seq number, here? It will be a moving target for PG's blob seq id on leader side, rigth?

src/lib/homestore_backend/resync_blob_data.fbs
yamingk: sure. will do

src/lib/homestore_backend/resync_pg_shard.fbs
sanebay: It could be for nuraft leader election priority

src/lib/homestore_backend/resync_pg_shard.fbs
sanebay: But the create and last modify time and state, we may not get log entries for these. We may have to calculate total bytes for each shard too. It might be easy to use whatever leader gives.

src/lib/homestore_backend/resync_pg_shard.fbs
sanebay: yeah we have to update it in increment resync.

src/lib/homestore_backend/resync_pg_shard.fbs
yamingk: but no one is actually setting them in HO, I can include them here for future use.

src/lib/homestore_backend/resync_pg_shard.fbs
yamingk: it is wrong to sync create/modify time to follower, it should be updated when this receiver member got the message and create the shard on it. Other fields like available/total/deleted, they are not actually used anywhere and even though we want them in the future, they need to be updated incrementally based on this member's put/del blob requests. Syncing from leader will cause wrong result because it is a moving target.

xiaoxichen (approved):
lgtm

sanebay (approved):

src/lib/homestore_backend/resync_blob_data.fbs
```diff
@@ -0,0 +1,16 @@
...
+    blob_id : uint64;
+    user_key_size : uint32;
+    data_size : uint32;
+    user_key : [ubyte];
+    data : [ubyte];
+}
+
+table ResyncBlobDataBatch {
+    data_array : [BlobData];
```
sanebay, line 13: We also need a bool end_of_batch to tell there is no more blobs in this batch.

sanebay (approved):

src/lib/homestore_backend/resync_pg_shard.fbs
```diff
@@ -0,0 +1,29 @@
...
+    members : [Member];                 // peers;
+}
+
+table ShardInfoEntry {
+    state : ubyte;                      // shard state;
+    pg_id : uint32;                     // pg id which this shard belongs to;
+    shard_id : uint64;                  // shard id to be created with;
+    shard_size : uint64;                // shard size;
+    created_time : uint64;              // shard creation time
+    last_modified_time : uint64;        // shard last modify time
```
xiaoxichen, line 21: as we agree to add created_lsn on shard, do we want to sync this ? 
or actually setting the created_lsn to snapshot_lsn is also right,  just sort of `inconsistency` we need to aware when we doing the scrubber.

yamingk: there should be no harm adding that, let me add it.

sanebay (commented):
native_include "sisl/utility/non_null_ptr.hpp";
this is needed on both .fbs files to compile with c++.

sanebay (approved):

PR #167, opened 2024-04-12
Adopt device type

xiaoxichen:

PR #171, opened 2024-04-20
compatible with cxxopt.

xiaoxichen:
in https://github.com/jarro2783/cxxopts/blob/v3.2.1/include/cxxopts.hpp#L1011-L1018

it checks the (!in),  if we read after reaching EOF,  the !in will be true which results a incorrect_argument_type exception thrown.

we should be very careful to avoid read after reaching eof.

### DIFF
--- a/src/lib/homestore_backend/hs_homeobject.cpp
+++ b/src/lib/homestore_backend/hs_homeobject.cpp
@@ -170,7 +170,7 @@ void HSHomeObject::init_homestore() {
                 {HS_SERVICE::REPLICATION,
                  hs_format_params{.dev_type = HSDevType::Data,
                                   .size_pct = 99.0,
-                                  .num_chunks = 60000,
+                                  .num_chunks = 32768,
                                   .block_size = _data_block_size,
                                   .alloc_type = blk_allocator_type_t::append,
                                   .chunk_sel_type = chunk_selector_type_t::CUSTOM}},


xiaoxichen:
we are expecting 
`/dev/sdb:NVME` can be parsed as `/dev/sdb` as path , `NVME` as type.
also expecting
`dev/sdc` can be parsed as`/dev/sdc` as path, `AUTO_DETECT` as type.

However the later one doesnt work prior to this change ,  as  we attempt to read post EOF for the type.

Now SM has to pass `AUTO_DETECT` explicitly, c.f https://github.corp.ebay.com/SDS/storage_mgr/pull/106/files#diff-77713e57d4672c7f5d23883b612967d7984474a2cc0222ac802717228501b50fR161


With this PR merged,  SM do not need pass-in the AUTO_DETECT

src/lib/homestore_backend/hs_homeobject.cpp
```diff
@@ -170,7 +170,7 @@ void HSHomeObject::init_homestore() {
                 {HS_SERVICE::REPLICATION,
                  hs_format_params{.dev_type = HSDevType::Data,
                                   .size_pct = 99.0,
-                                  .num_chunks = 60000,
```
JacksonYao287, line 173: what is the intention here? why 32768?
can we specify the chunk_size and let homestore itself determine how many chunks will be created according to the disk space?

xiaoxichen: https://github.com/eBay/HomeObject/pull/167/files#r1571091449
pls refer this discussion why specified chunk_size is less prefered. 

Regarding the intent of move down to 32K , it is to work around the bug we are discussing in HS channel
https://ebay-eng.slack.com/archives/CAVQXEAKS/p1713889924523089

Issue #172, opened 2024-04-22
calculate Pg used_bytes according to sum of the blob in this PG


JacksonYao287:
now, we calculate the used_bytes of a PG using total_bytes - avaliable_bytes, this is not accurate, since each pg might have different amount of blobs. 

we need to calculate Pg used_bytes according to sum of the blob in this PG

JacksonYao287:
fixed in this PR

PR #173, opened 2024-04-22
fix calculation of pg used_bytes

JacksonYao287:
#172

### DIFF
--- a/src/lib/homestore_backend/hs_homeobject.hpp
+++ b/src/lib/homestore_backend/hs_homeobject.hpp
@@ -78,9 +78,10 @@ public:
         peer_id_t replica_set_uuid;
         homestore::uuid_t index_table_uuid;
         blob_id_t blob_sequence_num;
-        uint64_t active_blob_count;    // Total number of active blobs
-        uint64_t tombstone_blob_count; // Total number of tombstones
-        pg_members members[1];         // ISO C++ forbids zero-size array
+        uint64_t active_blob_count;        // Total number of active blobs
+        uint64_t tombstone_blob_count;     // Total number of tombstones
+        uint64_t total_occupied_blk_count; // Total number of occupied blocks
+        pg_members members[1];             // ISO C++ forbids zero-size array
 
         uint32_t size() const { return sizeof(pg_info_superblk) + ((num_members - 1) * sizeof(pg_members)); }
         static std::string name() { return _pg_meta_name; }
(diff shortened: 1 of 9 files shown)


src/lib/homestore_backend/hs_homeobject.hpp
```diff
@@ -175,10 +175,14 @@ class HSHomeObject : public HomeObjectImpl {
                              pg_.durable_entities().active_blob_count.load(std::memory_order_relaxed));
                 GAUGE_UPDATE(*this, tombstone_blob_count,
                              pg_.durable_entities().tombstone_blob_count.load(std::memory_order_relaxed));
+                GAUGE_UPDATE(*this, total_occupied_space,
+                             pg_.durable_entities().total_occupied_blk_count.load(std::memory_order_relaxed) *
+                                 blk_size);
```
yuwmao, line 180: When will blk_size be updated?

JacksonYao287: blk_size is a fixed value, it will not be updated after initialized

yuwmao: Sorry, I read it wrong.

src/lib/homestore_backend/hs_shard_manager.cpp
```diff
@@ -328,6 +328,22 @@ void HSHomeObject::on_shard_message_commit(int64_t lsn, sisl::blob const& h, hom
             chunk_selector_->select_specific_chunk(blkids.chunk_num());
         }
         if (ctx) { ctx->promise_.setValue(ShardManager::Result< ShardInfo >(shard_info)); }
+
+        // TODO: add shard header blk count to total_occupied_blk_count and handle the case of replaying journal after
```
xiaoxichen, line 332: if this is the case, it also involve  in on_blob_put_commit?

JacksonYao287: yes, all the action in `on_commit` will be involved. we need to revisit this later.

xiaoxichen: how other metrics handle this case? IIRC we restart from a checkpoint ?

JacksonYao287: i believe other metrics do not handle this case. as you said,  we definitely restart from a checkpoint. but in homestore, checkpoint is not a real checkpoint, partial persistence might happen. actually, when we do cp_flush, we just start several future to flush those cp of every component. the key point here is that no atomicity of the cp of every component is promised. this is the potential risk of inconsistence。

JacksonYao287: I have post this question in homestore channel. hair will fix it.
https://ebay-eng.slack.com/archives/CAVQXEAKS/p1714099698282209

xiaoxichen (approved):

PR #180, opened 2024-06-05
add pg recovery test

JacksonYao287:
when need to implement the recovery test for all the persistent structures to achieve code complete. in this PR:
1 refined the existing shard test case
2 enabled the disabled test case before adopting homestore 6.4
3 add pg recover test, since other recovery cases are already covered by the existing test code. shard recovery is in shard test case, and blob/index recovery is in blob test case.


xiaoxichen (approved):
nit: verify_XXX -> deep_equal_XXX

PR #188, opened 2024-08-02
recover pg before repl dev is started

raakella1:
When the raft_repl_dev recovers in homestore, it replays the nuraft logs and calls on_commit for the lsn values greater than checkpoint lsn. PG recovery should be complete in Homeobject to commit the lsn. This PR moves the PG and shard recovery before raft_repl_dev is created while metablk recovery is happening.

### DIFF
--- a/src/lib/homestore_backend/hs_homeobject.cpp
+++ b/src/lib/homestore_backend/hs_homeobject.cpp
@@ -205,21 +205,8 @@ void HSHomeObject::init_homestore() {
                        to_string(_our_id));
     }
 
-    // recover PG
-    HomeStore::instance()->meta_service().register_handler(
-        _pg_meta_name,
-        [this](homestore::meta_blk* mblk, sisl::byte_view buf, size_t size) {
-            on_pg_meta_blk_found(std::move(buf), voidptr_cast(mblk));
-        },
-        nullptr, true);
-    HomeStore::instance()->meta_service().read_sub_sb(_pg_meta_name);
-
-    // recover shard
-    HomeStore::instance()->meta_service().register_handler(
-        _shard_meta_name,
-        [this](homestore::meta_blk* mblk, sisl::byte_view buf, size_t size) { on_shard_meta_blk_found(mblk, buf); },
-        [this](bool success) { on_shard_meta_blk_recover_completed(success); }, true);
-    HomeStore::instance()->meta_service().read_sub_sb(_shard_meta_name);
+    // Replication service is started after PG sb is recovered
+    set_pg_repl_dev_after_recovery();
 
     recovery_done_ = true;
     LOGI("Initialize and start HomeStore is successfully");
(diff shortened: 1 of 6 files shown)


JacksonYao287:
in `on_pg_meta_blk_found`, we have this `auto v = hs_repl_service().get_repl_dev(pg_sb->replica_set_uuid);`
so , when we recover PG, we need the corresponding repl_dev to recovered first.

however, we do not register any handler to recover repl_dev when meta service start. what we do is just keeping the repl service metablk  in a buffer.
```
GenericReplService::GenericReplService(cshared< ReplApplication >& repl_app) : m_repl_app{repl_app} {
    m_sb_bufs.reserve(100);
    meta_service().register_handler(
        get_meta_blk_name(),
        [this](meta_blk* mblk, sisl::byte_view buf, size_t) {
            m_sb_bufs.emplace_back(std::pair(std::move(buf), voidptr_cast(mblk)));
        },
        nullptr);
}

RaftReplService::RaftReplService(cshared< ReplApplication >& repl_app) : GenericReplService{repl_app} {
    m_config_sb_bufs.reserve(100);
    meta_service().register_handler(
        get_meta_blk_name() + "_raft_config",
        [this](meta_blk* mblk, sisl::byte_view buf, size_t) {
            m_config_sb_bufs.emplace_back(std::pair(std::move(buf), voidptr_cast(mblk)));
        },
        nullptr, false, std::optional< meta_subtype_vec_t >({get_meta_blk_name()}));
}
```

only when `RaftReplService::start()` , the repl_dev is actually recoverd.
```
    for (auto const& [buf, mblk] : m_sb_bufs) {
        load_repl_dev(buf, voidptr_cast(mblk));
    }
```
however , `RaftReplService::start()` is after mete service starts.  only at this time, we can make sure that all the repl_dev is recovered and pg can be recovered too. so , we need a extra step to recover pg and shard. this is why PG recovery  is  moved after homestore start.

src/lib/homestore_backend/hs_homeobject.cpp
```diff
@@ -272,6 +259,28 @@ void HSHomeObject::register_homestore_metablk_callback() {
...
         nullptr, true);
+
+    // recover PG
+    HomeStore::instance()->meta_service().register_handler(
+        _pg_meta_name,
+        [this](homestore::meta_blk* mblk, sisl::byte_view buf, size_t size) {
+            on_pg_meta_blk_found(std::move(buf), voidptr_cast(mblk));
+        },
+        nullptr, true);
+
```
JacksonYao287, line 270: `HomeStore::instance()->meta_service().read_sub_sb(_shard_meta_name);`
`HomeStore::instance()->meta_service().read_sub_sb(_pg_meta_name);`

don`t we need this ?

xiaoxichen: yeah I also dont under stand how this works.  The initiative probably want to move these line before HS.start(), which should be put into `register_homestore_metablk_callback`

src/lib/homestore_backend/hs_pg_manager.cpp
```diff
@@ -195,22 +197,19 @@ void HSHomeObject::on_pg_meta_blk_found(sisl::byte_view const& buf, void* meta_c
...
     // During PG recovery check if index is already recoverd else
     // add entry in map, so that index recovery can update the PG.
     std::scoped_lock lg(index_lock_);
-    auto it = index_table_pg_map_.find(uuid_str);
-    RELEASE_ASSERT(it != index_table_pg_map_.end(), "IndexTable should be recovered before PG");
-    hs_pg->index_table_ = it->second.index_table;
-    it->second.pg_id = pg_id;
+
+    if (auto it = index_table_pg_map_.find(uuid_str); it != index_table_pg_map_.end()) {
+        hs_pg->index_table_ = it->second.index_table;
```
JacksonYao287, line 208: this is  called when meta service start , which is the first of all services. so this will be prior to index_service start.  so I think we will never reach here.

src/lib/homestore_backend/index_kv.cpp
```diff
@@ -32,9 +32,15 @@ HSHomeObject::recover_index_table(homestore::superblk< homestore::index_table_sb
 
     // Check if PG is already recovered.
     std::scoped_lock lock_guard(index_lock_);
-    auto it = index_table_pg_map_.find(uuid_str);
-    RELEASE_ASSERT(it == index_table_pg_map_.end(), "PG should be recovered after IndexTable");
-    index_table_pg_map_.emplace(uuid_str, PgIndexTable{0, index_table});
+    if (auto it = index_table_pg_map_.find(uuid_str); it == index_table_pg_map_.end()) {
+        index_table_pg_map_.emplace(uuid_str, PgIndexTable{0, index_table});
```
JacksonYao287, line 36: the same as above

JacksonYao287 (commented):
some minor comments. other parts looks good

src/lib/homestore_backend/hs_homeobject.cpp
raakella1: I think we need this if we register callback after metablk init. If we move the register to before metablk init, the cb gets called

PR #196, opened 2024-08-07
add on_restart callback to listener

raakella1:


xiaoxichen (commented):
lgtm

src/lib/homestore_backend/replication_state_machine.cpp
```diff
@@ -71,6 +71,8 @@ void ReplicationStateMachine::on_rollback(int64_t lsn, sisl::blob const& header,
     }
 }
 
+void ReplicationStateMachine::on_restart() { home_object_->on_replica_restart(); }
```
xiaoxichen, line 74: should we add `std::call_once` ?  since the on_restart will be called for each repl_dev but on_replica_restart only need to run one time.

raakella1: Yes, you are right. We need call_once

xiaoxichen (approved):

src/lib/homestore_backend/hs_homeobject.cpp
```diff
@@ -198,29 +198,14 @@ void HSHomeObject::init_homestore() {
         svc_sb.create(sizeof(svc_info_superblk_t));
         svc_sb->svc_id_ = _our_id;
         svc_sb.write();
+        on_replica_restart();
```
JacksonYao287, line 201: when we format homestore, it means this is the first-time start and no pg and shard should exist. so do we need this here?

src/lib/homestore_backend/hs_homeobject.cpp
```diff
@@ -232,6 +217,28 @@ void HSHomeObject::init_homestore() {
     }
 }
 
+void HSHomeObject::on_replica_restart() {
+    std::call_once(replica_restart_flag_, [this]() {
```
JacksonYao287, line 221: according to the logic here, when the first repl_dev is recovered , all the pg and shard will be recovered also.
however, pg2 , which depends on repl_dev2, can not be recovered since when repl_dev1 is recoverd, repl_dev2 is not recovered, right?

xiaoxichen: can you explain why pg2 cannot recovered?  the callback get called after all the repl_dev metablk recovered.

src/lib/homestore_backend/hs_homeobject.cpp
xiaoxichen: otherwise when we persist PG metadata we will get exception
```
[08/12/24 18:09:40-07:00] [C] [192263] [type=PGManager] not registered yet!
```

xiaoxichen (approved):

src/lib/homestore_backend/hs_homeobject.cpp
JacksonYao287: so, can we only register pg and shard here., not do `read_sub_sb`?

xiaoxichen: yes we can , but is that worth to saving 2 IOs?

JacksonYao287: this is not nitpicking here and I do not want to save any io since no IO will happen(first-time start, metaservice will have not data for pg and shard).

 my point here is it seems a little weird to me that we have a `on_replica_restart` in the first-time start path, which should be in the recovery path.

PR #204, opened 2024-08-26
Add baseline resync read part in homeobject.

sanebay:
Add read_snapshot_data to go over all shards and blobs of a PG. If obj_id is zero, send all shards. obj_id is concatenation of blob sequence number and batch number. For all other values of obj_id, we send batch of blobs for a shard. Once all blobs are finished in a shard, we move to next shard_id, and batch_num is reset to 0. Add LSN in shard metadata to ignore all reads of shards which are created later that the snapshot LSN.

Added temporary code to write blobs and metadata, tested with SM long running test to create a baseline resync with follower. 
Tested with UT.

### DIFF
--- a/src/include/homeobject/shard_manager.hpp
+++ b/src/include/homeobject/shard_manager.hpp
@@ -22,6 +22,7 @@ struct ShardInfo {
     shard_id_t id;
     pg_id_t placement_group;
     State state;
+    uint64_t lsn;
     uint64_t created_time;
     uint64_t last_modified_time;
     uint64_t available_capacity_bytes;
--- a/src/lib/homestore_backend/hs_blob_manager.cpp
+++ b/src/lib/homestore_backend/hs_blob_manager.cpp
@@ -4,6 +4,7 @@
 #include "lib/homeobject_impl.hpp"
 #include "lib/blob_route.hpp"
 #include <homestore/homestore.hpp>
+#include <homestore/blkdata_service.hpp>
 
 SISL_LOGGING_DECL(blobmgr)
 
@@ -162,9 +163,7 @@ BlobManager::AsyncResult< blob_id_t > HSHomeObject::_put_blob(ShardInfo const& s
     return req->result().deferValue([this, req, repl_dev](const auto& result) -> BlobManager::AsyncResult< blob_id_t > {
         if (result.hasError()) {
             auto err = result.error();
-            if (err.getCode() == BlobErrorCode::NOT_LEADER) {
-                err.current_leader = repl_dev->get_leader_id();
-            }
+            if (err.getCode() == BlobErrorCode::NOT_LEADER) { err.current_leader = repl_dev->get_leader_id(); }
             return folly::makeUnexpected(err);
         }
         auto blob_info = result.value();
@@ -259,7 +258,13 @@ BlobManager::AsyncResult< Blob > HSHomeObject::_get_blob(ShardInfo const& shard,
         return folly::makeUnexpected(r.error());
     }
 
-    auto const blkid = r.value();
+    return _get_blob_data(repl_dev, shard.id, blob_id, req_offset, req_len, r.value() /* blkid*/);
+}
+
+BlobManager::AsyncResult< Blob > HSHomeObject::_get_blob_data(const shared< homestore::ReplDev >& repl_dev,
+                                                              shard_id_t shard_id, blob_id_t blob_id,
+                                                              uint64_t req_offset, uint64_t req_len,
+                                                              const homestore::MultiBlkId& blkid) const {
     auto const total_size = blkid.blk_count() * repl_dev->get_blk_size();
     sisl::io_blob_safe read_buf{total_size, io_align};
 
@@ -267,10 +272,10 @@ BlobManager::AsyncResult< Blob > HSHomeObject::_get_blob(ShardInfo const& shard,
     sgs.size = total_size;
     sgs.iovs.emplace_back(iovec{.iov_base = read_buf.bytes(), .iov_len = read_buf.size()});
 
-    BLOGT(shard.id, blob_id, "Blob get request: blkid={}, buf={}", blkid.to_string(), (void*)read_buf.bytes());
+    BLOGT(shard_id, blob_id, "Blob get request: blkid={}, buf={}", blkid.to_string(), (void*)read_buf.bytes());
     return repl_dev->async_read(blkid, sgs, total_size)
-        .thenValue([this, blob_id, shard_id = shard.id, req_len, req_offset, blkid,
-                    read_buf = std::move(read_buf), repl_dev](auto&& result) mutable -> BlobManager::AsyncResult< Blob > {
+        .thenValue([this, blob_id, shard_id, req_len, req_offset, blkid, repl_dev,
+                    read_buf = std::move(read_buf)](auto&& result) mutable -> BlobManager::AsyncResult< Blob > {
             if (result) {
                 BLOGE(shard_id, blob_id, "Failed to get blob: err={}", blob_id, shard_id, result.value());
                 return folly::makeUnexpected(BlobError(BlobErrorCode::READ_FAILED));
@@ -379,9 +384,7 @@ BlobManager::NullAsyncResult HSHomeObject::_del_blob(ShardInfo const& shard, blo
     return req->result().deferValue([repl_dev](const auto& result) -> folly::Expected< folly::Unit, BlobError > {
         if (result.hasError()) {
             auto err = result.error();
-            if (err.getCode() == BlobErrorCode::NOT_LEADER) {
-                err.current_leader = repl_dev->get_leader_id();
-            }
+            if (err.getCode() == BlobErrorCode::NOT_LEADER) { err.current_leader = repl_dev->get_leader_id(); }
             return folly::makeUnexpected(err);
         }
         auto blob_info = result.value();
--- a/src/lib/homestore_backend/hs_shard_manager.cpp
+++ b/src/lib/homestore_backend/hs_shard_manager.cpp
@@ -67,6 +67,7 @@ std::string HSHomeObject::serialize_shard_info(const ShardInfo& info) {
     j["shard_info"]["shard_id_t"] = info.id;
     j["shard_info"]["pg_id_t"] = info.placement_group;
     j["shard_info"]["state"] = info.state;
+    j["shard_info"]["lsn"] = info.lsn;
     j["shard_info"]["created_time"] = info.created_time;
     j["shard_info"]["modified_time"] = info.last_modified_time;
     j["shard_info"]["total_capacity"] = info.total_capacity_bytes;
@@ -81,6 +82,7 @@ ShardInfo HSHomeObject::deserialize_shard_info(const char* json_str, size_t str_
     shard_info.id = shard_json["shard_info"]["shard_id_t"].get< shard_id_t >();
     shard_info.placement_group = shard_json["shard_info"]["pg_id_t"].get< pg_id_t >();
     shard_info.state = static_cast< ShardInfo::State >(shard_json["shard_info"]["state"].get< int >());
+    shard_info.lsn = shard_json["shard_info"]["lsn"].get< uint64_t >();
     shard_info.created_time = shard_json["shard_info"]["created_time"].get< uint64_t >();
     shard_info.last_modified_time = shard_json["shard_info"]["modified_time"].get< uint64_t >();
     shard_info.available_capacity_bytes = shard_json["shard_info"]["available_capacity"].get< uint64_t >();
@@ -116,6 +118,7 @@ ShardManager::AsyncResult< ShardInfo > HSHomeObject::_create_shard(pg_id_t pg_ow
     sb->info = ShardInfo{.id = new_shard_id,
                          .placement_group = pg_owner,
                          .state = ShardInfo::State::OPEN,
+                         .lsn = 0,
                          .created_time = create_time,
                          .last_modified_time = create_time,
                          .available_capacity_bytes = size_bytes,
@@ -313,7 +316,8 @@ void HSHomeObject::on_shard_message_commit(int64_t lsn, sisl::blob const& h, hom
     switch (header->msg_type) {
     case ReplicationMessageType::CREATE_SHARD_MSG: {
         auto sb = r_cast< shard_info_superblk const* >(h.cbytes() + sizeof(ReplicationMessageHeader));
-        auto const shard_info = sb->info;
+        auto shard_info = sb->info;
+        shard_info.lsn = lsn;
 
         bool shard_exist = false;
         {
--- a/src/lib/homestore_backend/pg_blob_iterator.cpp
+++ b/src/lib/homestore_backend/pg_blob_iterator.cpp
@@ -1,0 +1,129 @@
+#include "hs_homeobject.hpp"
+#include <sisl/logging/logging.h>
+#include <sisl/options/options.h>
+#include <sisl/settings/settings.hpp>
+#include "generated/resync_pg_shard_generated.h"
+#include "generated/resync_blob_data_generated.h"
+
+namespace homeobject {
+
+HSHomeObject::PGBlobIterator::PGBlobIterator(HSHomeObject& home_obj, homestore::group_id_t group_id,
+                                             uint64_t upto_lsn) :
+        home_obj_(home_obj), group_id_(group_id) {
+    auto pg = get_pg_metadata();
+    pg_id_ = pg->pg_info_.id;
+    repl_dev_ = static_cast< HS_PG* >(pg)->repl_dev_;
+    if (upto_lsn != 0) {
+        // Iterate all shards and its blob which have lsn <= upto_lsn
+        for (auto& shard : pg->shards_) {
+            auto sequence_num = home_obj_.get_sequence_num_from_shard_id(shard->info.id);
+            if (shard->info.lsn <= upto_lsn) { max_shard_seq_num_ = std::max(max_shard_seq_num_, sequence_num); }
+        }
+    } else {
+        max_shard_seq_num_ = pg->shard_sequence_num_;
+    }
+}
+
+PG* HSHomeObject::PGBlobIterator::get_pg_metadata() {
+    std::scoped_lock lock_guard(home_obj_._pg_lock);
+    auto iter = home_obj_._pg_map.begin();
+    for (; iter != home_obj_._pg_map.end(); iter++) {
+        if (iter->second->pg_info_.replica_set_uuid == group_id_) { break; }
+    }
+
+    RELEASE_ASSERT(iter != home_obj_._pg_map.end(), "PG not found replica_set_uuid={}",
+                   boost::uuids::to_string(group_id_));
+    return iter->second.get();
+}
+
+void HSHomeObject::PGBlobIterator::create_pg_shard_snapshot_data(sisl::io_blob_safe& meta_blob) {
+    auto pg = get_pg_metadata();
+    auto& pg_info = pg->pg_info_;
+    auto& pg_shards = pg->shards_;
+
+    flatbuffers::FlatBufferBuilder builder;
+    std::vector< std::uint8_t > uuid(pg_info.replica_set_uuid.size());
+    std::copy(pg_info.replica_set_uuid.begin(), pg_info.replica_set_uuid.end(), uuid.begin());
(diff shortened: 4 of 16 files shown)


src/include/homeobject/shard_manager.hpp
```diff
@@ -22,6 +22,7 @@ struct ShardInfo {
     shard_id_t id;
     pg_id_t placement_group;
     State state;
+    uint64_t lsn;
```
szmyd, line 25: Does this need to be exposed to SM or other? If not, can we hide it somewhere that's not part of the public API?

sanebay: Yeah we can make it private.

szmyd: it's okay to leave public; i was trying to not include it in the public API as moving/changing it will impact the layout/size of this structure for code that depends on homeobject (SM), requiring a re-compile rather than a simple re-link.

src/lib/homestore_backend/hs_blob_manager.cpp
```diff
@@ -250,17 +251,23 @@ BlobManager::AsyncResult< Blob > HSHomeObject::_get_blob(ShardInfo const& shard,
         return folly::makeUnexpected(r.error());
     }
 
-    auto const blkid = r.value();
+    return _get_blob_data(repl_dev, shard.id, blob_id, req_offset, req_len, r.value() /* blkid*/);
+}
+
+BlobManager::AsyncResult< Blob > HSHomeObject::_get_blob_data(shared< homestore::ReplDev > repl_dev,
```
szmyd, line 257: This can be by `const` reference yeah? No need to copy the shared_ptr.

src/lib/homestore_backend/replication_state_machine.cpp
```diff
@@ -159,30 +164,139 @@ ReplicationStateMachine::create_snapshot(std::shared_ptr< homestore::snapshot_co
...
+    if (obj_id == 0) {
+        // obj_id = 0 means its the first message and we send the pg and its shards metadata.
+        cur_snapshot_batch_num = 0;
+        pg_iter->create_pg_shard_snapshot_data(snp_data->blob);
+        RELEASE_ASSERT(snp_data->blob.size() > 0, "Empty metadata snapshot data");
+        LOGD("Read snapshot data first message {}", log_str);
+        return 0;
+    }
+
+    if (shard_seq_num != pg_iter->cur_shard_seq_num_ || batch_number != cur_snapshot_batch_num) {
```
xiaoxichen, line 218: the implementation right now is not friendly for retry.   writer side cannot re-try on shard level if it failed to apply some batch.

for example,   if {shard x, batch 1} is the last batch of shard x,  the reader(leader) side will set 
```
cur_shard_seq_num_ = x+1
cur_snapshot_batch_num = 0
```
writer (follower) send back {x, 1} will result a failure.

IIRC we agreed in the design if failure happens in {x, y} , we rewind to {x, 0}

src/lib/homestore_backend/replication_state_machine.cpp
```diff
@@ -159,30 +164,139 @@ ReplicationStateMachine::create_snapshot(std::shared_ptr< homestore::snapshot_co
...
+        pg_iter = new HSHomeObject::PGBlobIterator(*home_object_, repl_dev()->group_id());
+        snp_data->user_ctx = (void*)pg_iter;
+    } else {
+        pg_iter = r_cast< HSHomeObject::PGBlobIterator* >(snp_data->user_ctx);
+    }
+
+    // Nuraft uses obj_id as a way to track the state of the snapshot read and write.
+    // Nuraft starts with obj_id == 0 as first message always, leader send all the shards and
+    // PG metadata as response. Follower responds with next obj_id it expects. obj_id's are
+    // encoded in the form of obj_id = shard_seq_num(6 bytes) | batch_number(2 bytes)
```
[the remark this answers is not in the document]
xiaoxichen: https://github.com/eBay/HomeObject/blob/acb04e8e55186734769ab9236de087d4226ef244/src/lib/homeobject_impl.hpp#L68

yes the shardID started with 0

src/lib/homestore_backend/replication_state_machine.cpp
xiaoxichen: my general suggestion is changing pg_iter to shard_iter, and let the writer controls the pace of which shard it is requesting. 

we can maintain a bitmap `is_shard_completed[shardID]` to track if a shard are fully transferred.  and based on that we know we are fully done (set the `is_last_obj = true`).

some enhancement can be :
1.  If writer side can confirm a shard done by sending next_objid = {shardid,  0} if our batch started with 1.
2.  if writer side can tell reader it has all shards (based on the list it got from first step) ,  leader can check if it does reach to the end,  if not , we can better error out.

src/lib/homestore_backend/replication_state_machine.cpp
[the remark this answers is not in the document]
sanebay: shard_id is incremented first, so it starts with 1. 
https://github.com/eBay/HomeObject/blob/main/src/lib/homestore_backend/hs_shard_manager.cpp#L55

src/lib/homestore_backend/hs_blob_manager.cpp
sanebay: Ack

src/lib/homestore_backend/pg_blob_iterator.cpp
```diff
@@ -0,0 +1,139 @@
...
+}
+
+void HSHomeObject::PGBlobIterator::create_pg_shard_snapshot_data(sisl::io_blob_safe& meta_blob) {
+    auto pg = get_pg_metadata();
+    auto& pg_info = pg->pg_info_;
+    auto& pg_shards = pg->shards_;
+
+    flatbuffers::FlatBufferBuilder builder;
+    std::vector< ::flatbuffers::Offset< Member > > member_entries;
+    member_entries.reserve(pg_info.members.size());
```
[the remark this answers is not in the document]
sanebay: Yeah right. not needed to copy the member list.

src/lib/homestore_backend/replication_state_machine.cpp
sanebay: Yeah agree. we need to support for retry for errors or if it crashed and came back. There is a chance it can go back.  If we see a difference in shard_sequence or batch number and is less than current, then we know follower is re-requesting the data and we can accept it as our current. We can continue from what the follower has provided.

xiaoxichen: yeah just do not trust batch number... the batch will change by IO, 
for example, with batchsize = 3
```
B1{lsn 1}, B2{lsn 2}, B3{{lsn 3}, B4 {lsn 4}, B5{lsn 5}, B6{lsn 6}
```
will become
```
batch 1:
B1-B3
batch 2:
B4-B6
```

later on if delete comes in for B1-B3
it will be 
```
B1{TombStone 11}, B2{TombStone 12}, B3{{TombStone 13}, B4 {lsn 4}, B5{lsn 5}, B6{lsn 6}
```
so we only have  
```
batch 1:
B4-B6
```

src/lib/homestore_backend/hs_shard_manager.cpp
```diff
@@ -313,7 +316,8 @@ void HSHomeObject::on_shard_message_commit(int64_t lsn, sisl::blob const& h, hom
     switch (header->msg_type) {
     case ReplicationMessageType::CREATE_SHARD_MSG: {
         auto sb = r_cast< shard_info_superblk const* >(h.cbytes() + sizeof(ReplicationMessageHeader));
-        auto const shard_info = sb->info;
+        auto shard_info = sb->info;
+        shard_info.lsn = lsn;
```
JacksonYao287, line 320: should we also set lsn in line 354 for seal_shard?

sanebay: lsn is for create shard. To know when this shard was created so that we can ignore if it was created after snapshot taken.

src/lib/homestore_backend/pg_blob_iterator.cpp
```diff
@@ -0,0 +1,128 @@
...
+            num_blobs++;
+            total_bytes += blob.body.size() + blob.user_key.size();
+            if (num_blobs > max_num_blobs_in_batch || total_bytes > max_batch_size_bytes) { return 0; }
+
+            BlobInfoData blob_data{{info.shard_id, info.blob_id, std::move(info.pbas)}, std::move(blob)};
+            blob_data_vec.push_back(std::move(blob_data));
+            cur_blob_id_ = info.blob_id;
+        }
+
+        if (blob_info_vec.empty()) {
```
xiaoxichen, line 95: this is not right,  we may possibly get `max_num_blobs_in_batch` tombstones. We should based on `query_blobs_in_shard ` return us `r.value().size() == 0`

xiaoxichen (commented):
LGTM except a bug for reader size.  writer side code is ignored.

src/lib/homestore_backend/replication_state_machine.cpp
```diff
@@ -159,30 +164,148 @@ ReplicationStateMachine::create_snapshot(std::shared_ptr< homestore::snapshot_co
...
+    return true;
 }
 
-std::shared_ptr< homestore::snapshot_context > ReplicationStateMachine::last_snapshot() {
-    LOGE("last_snapshot not implemented");
-    return nullptr;
-}
+std::shared_ptr< homestore::snapshot_context > ReplicationStateMachine::last_snapshot() { return m_snapshot_context; }
 
 int ReplicationStateMachine::read_snapshot_data(std::shared_ptr< homestore::snapshot_context > context,
```
xiaoxichen, line 179: Not sure if it is too late/worth...

the naming `read_snapshot_data` is conflict with this 
 interface, which is misleading.

https://github.com/eBay/NuRaft/blob/99eeef34a2620686e0dd40ad7fbd5cab561140fc/include/libnuraft/state_machine.hxx#L235-L255

sanebay: Looks like its deprecated and nuraft_mesg is not using it too. And HO is not including it directly.

src/lib/homestore_backend/pg_blob_iterator.cpp
sanebay: We are skipping the tombstones at line 72.  Tombstone has tombstone pbas which has size but its values are 0

xiaoxichen: yes I see we skip all tombstone in L72, so tombstone will not be put into `blob_data_vec`.  Later on in this line we check `blob_info_vec.empty()`.

so in the case when all entries we got from `query_blobs_in_shard`( L67) are tombstone, the `blob_data_vec`  will be empty then we mark the shard as finished.  This is not right, there can still be live blobs after the array of tombstones

src/lib/homestore_backend/replication_state_machine.cpp
xiaoxichen: yes by conflict I mean, it is easy to be confused :).   I was wrongly assume we are using the deprecated endpoint then later I aware we translate that in homestore .

src/lib/homestore_backend/pg_blob_iterator.cpp
sanebay: We check the results of index. If its empty, there will be no more blobs in that shard. Changed the name to index_results_vec to avoid confusion.

src/lib/homestore_backend/replication_state_machine.cpp
sanebay: Yeah didnt know this name was used. We can rename it in future, as it needs changes in HS.

src/lib/homestore_backend/pg_blob_iterator.cpp
xiaoxichen: oh sorry it is my bad,  I mixed `blob_data_vec` vs `blob_info_vec`

xiaoxichen (dismissed):

src/lib/homestore_backend/pg_blob_iterator.cpp
```diff
@@ -0,0 +1,129 @@
...
+                                                     std::vector< BlobInfoData >& blob_data_vec, bool& end_of_shard) {
+    end_of_shard = false;
+    uint64_t total_bytes = 0, num_blobs = 0;
+    while (true) {
+        auto r = home_obj_.query_blobs_in_shard(pg_id_, cur_shard_seq_num_, cur_blob_id_ + 1, max_num_blobs_in_batch);
+        if (!r) { return -1; }
+        auto& index_results_vec = r.value();
+        for (auto& info : index_results_vec) {
+            if (info.pbas == HSHomeObject::tombstone_pbas) {
+                // Skip deleted blobs
```
JacksonYao287, line 72: not sure can we do it like this?

suppose leader has {1,1} to {5,5}, and {5,7} to {10,10}({shard_id, blob_id}) , last lsn is 120,  {5,6} is deleted at lsn 100. the log at leader has been compacted to 110.
follower has {1,1} to {6.0} and the last lsn is 80 , then if baseline resync occurs, the follower will never konw that {5,6} has been deleted since it is not aware fo lsn 100.

so I think here we should set some special data for the tombstone_pbas in the blob_info_vec which will be send to follower , so that the follower can identify that this blob is deleted.

another question, if GC happens and tombstone is also removed, so how can leader let the follower know this info when baseline resync happens.

pls correct me if i missunderstand anything

JacksonYao287 (commented):
sorry for the late review @sanebay , pls take a look at my comments and correct if i am wrong

src/lib/homestore_backend/pg_blob_iterator.cpp
xiaoxichen: We havent put LSN into each blob index so ATM it is a full resync --- i.e all existing data can be discard.  So there is no issue for deleted blob and no necessary for transferring tombstone.

Extending the discussion further ,  assuming we have LSN in blob index,  we can let follower to set its current LSN and leader will only send the [follower_lsn ,  snapshot_lsn] to follower.  In this case,  as you said , we care about blob deletion.   The trivial approach is leader send out active blob list in <shard_id =S , batch =0 > , follower mark all blobs not in active blob list as deleted.

I think we are not yet have solid thinking regarding the "incremental snapshot" especially with a good amount of reserved log entries.  Though personally I am loving it.

src/lib/homestore_backend/replication_state_machine.cpp
```diff
@@ -159,30 +164,148 @@ ReplicationStateMachine::create_snapshot(std::shared_ptr< homestore::snapshot_co
     auto ctx = dynamic_pointer_cast< homestore::nuraft_snapshot_context >(context);
     auto s = ctx->nuraft_snapshot();
     LOGI("create snapshot, last_log_idx_: {} , last_log_term_: {}", s->get_last_log_idx(), s->get_last_log_term());
+    m_snapshot_context = context;
```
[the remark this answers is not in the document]
xiaoxichen: I doubt it has this case as the API itself doesnt have replicaID for leader to distinguish between concurrent snapshot recovery.  But feel free to check nuraft code and let us know

src/lib/homestore_backend/pg_blob_iterator.cpp
```diff
@@ -0,0 +1,128 @@
...
+                                             uint64_t upto_lsn) :
+        home_obj_(home_obj), group_id_(group_id) {
+    auto pg = get_pg_metadata();
+    pg_id_ = pg->pg_info_.id;
+    repl_dev_ = static_cast< HS_PG* >(pg)->repl_dev_;
+    if (upto_lsn != 0) {
+        // Iterate all shards and its blob which have lsn <= upto_lsn
+        for (auto& shard : pg->shards_) {
+            auto sequence_num = home_obj_.get_sequence_num_from_shard_id(shard->info.id);
+            if (shard->info.lsn <= upto_lsn) { max_shard_seq_num_ = std::max(max_shard_seq_num_, sequence_num); }
```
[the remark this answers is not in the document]
sanebay: The reason was pg->shards_, was list and its not sorted and since its only one time. Else make shards_ a set, based on lsn.

src/lib/homestore_backend/replication_state_machine.cpp
```diff
@@ -159,30 +164,148 @@ ReplicationStateMachine::create_snapshot(std::shared_ptr< homestore::snapshot_co
...
+    // Once all the blobs are finished sending, end_of_scan will be true.
+    int64_t obj_id = snp_data->offset;
+    uint64_t shard_seq_num = obj_id >> 16;
+    uint64_t batch_number = obj_id & 0xFFFF;
+    auto log_str = fmt::format("group={} term={} lsn={} shard_seq={} batch_num={} size={}",
+                               boost::uuids::to_string(repl_dev()->group_id()), s->get_last_log_term(),
+                               s->get_last_log_idx(), shard_seq_num, batch_number, snp_data->blob.size());
+    if (obj_id == 0) {
+        // obj_id = 0 means its the first message and we send the pg and its shards metadata.
+        cur_snapshot_batch_num = 0;
```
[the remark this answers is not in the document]
sanebay: Yeah you are right. infact cur_snapshot_batch_num was part of iterator initially, then changed it. I will make it part of iterator.

src/lib/homestore_backend/replication_state_machine.cpp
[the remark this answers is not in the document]
sanebay: snapshot is single in leader. We need to protect it with lock.

src/lib/homestore_backend/replication_state_machine.cpp
```diff
@@ -159,30 +164,148 @@ ReplicationStateMachine::create_snapshot(std::shared_ptr< homestore::snapshot_co
...
+        // it crashed and want to continue from where it left.
+        LOGW("Shard or batch number not same as in iterator shard={}/{} batch_num={}/{}", shard_seq_num,
+             pg_iter->cur_shard_seq_num_, batch_number, cur_snapshot_batch_num);
+        if (shard_seq_num > pg_iter->cur_shard_seq_num_ || batch_number > cur_snapshot_batch_num) {
+            // If we retrieve some invalid values, return error.
+            return -1;
+        }
+
+        // Use the shard sequence number provided by the follower and we restart the batch.
+        pg_iter->cur_shard_seq_num_ = shard_seq_num;
```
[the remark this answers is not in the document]
sanebay: Yes. that is correct. Also we should periodically flush the last batch in write PR.

src/lib/homestore_backend/pg_blob_iterator.cpp
sanebay: Yes this is a valid scenario. We can do two ways.
1. Either follower see's a gap in blob-sequence and assume its deletion.
2. More safe approach is use scrubber. In scrubber leader sends the valid list, its crc across followers.  Follower can use this to delete.

src/lib/homestore_backend/replication_state_machine.cpp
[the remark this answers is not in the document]
xiaoxichen: ok, yes it is right to have a lock.  We should also check the <term, idx> of new snapshot is larger (newer) than m_snapshot_context?

src/lib/homestore_backend/pg_blob_iterator.cpp
xiaoxichen: I think before we have the `incremental snapshot` feature , i.e only transfer diff between 2 snapshots,    we would better erase everything on receiver side as anyway we start from scratch

JacksonYao287:
the code here LGTM.

there are two left questions:
1 if blob is deleted and the deleted blob log is compacted, when baseline resync occurs , how to make follower aware of the deletion of this blob?

2 if shard is sealed and the seal shard log is compacted, when baseline resync occurs , how to make follower aware of the seal of this shard?

they are essentially the same question. let`s think a bit more and discuss it in homestore meeting if necessary.

src/include/homeobject/shard_manager.hpp
sanebay: Discussed with Brian. ShardInfo used in public api also used everywhere in hs_shard_manager.cpp. May be cleanup in future.

src/lib/homestore_backend/replication_state_machine.cpp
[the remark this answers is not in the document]
sanebay: Ack

raakella1 (approved):
LG!

PR #205, opened 2024-08-30
Adaptive changes for HttpManager

yuwmao:
Relevant changes: https://github.com/eBay/IOManager/pull/100

### DIFF
--- a/conanfile.py
+++ b/conanfile.py
@@ -9,7 +9,7 @@ required_conan_version = ">=1.60.0"
 
 class HomeObjectConan(ConanFile):
     name = "homeobject"
-    version = "2.0.12"
+    version = "2.0.13"
 
     homepage = "https://github.com/eBay/HomeObject"
     description = "Blob Store built on HomeReplication"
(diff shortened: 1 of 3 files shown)


xiaoxichen:
seem like you forgot the change on conanfile.py

raakella1 (changes requested):
Can you fix this https://github.com/eBay/HomeObject/blob/4d999d4c1d8173a319f5b39946ccdf6af5bb5924/src/lib/homestore_backend/hs_homeobject.cpp#L117
This will restart http server and pass empty value for certs. We should only start http server once in SM. This restart should be removed

yuwmao:
> Can you fix this
> 
> https://github.com/eBay/HomeObject/blob/4d999d4c1d8173a319f5b39946ccdf6af5bb5924/src/lib/homestore_backend/hs_homeobject.cpp#L117
> 
> 
> This will restart http server and pass empty value for certs. We should only start http server once in SM. This restart should be removed

Sure.

yuwmao:
> seem like you forgot the change on conanfile.py

Yes, changed.

xiaoxichen (approved):

PR #216, opened 2024-10-01
Add replace member changes. Perist membership details.

sanebay:


src/lib/homestore_backend/hs_pg_manager.cpp
```diff
@@ -145,9 +145,55 @@ void HSHomeObject::on_create_pg_message_commit(int64_t lsn, sisl::blob const& he
...
+        group_id = repl_dev.group_id();
+    }
+
+    return hs_repl_service()
+        .replace_member(group_id, old_member, new_member.id, commit_quorum)
+        .via(executor_)
+        .thenValue([this, pg_id, old_member, new_member](auto&& v) mutable -> PGManager::NullAsyncResult {
+            if (v.hasError()) { return folly::makeUnexpected(toPgError(v.error())); }
+
+            on_pg_replace_member(pg_id, old_member, new_member);
```
xiaoxichen, line 171: The `on_pg_replace_member` here will only be execute on master which is wrong.

The member change is a nuraft config commit, we need to handle it in the `commit_config` callback(https://github.com/eBay/HomeStore/blob/0fa6a0c127680edb89291633ede8c937d10ba990/src/lib/replication/repl_dev/raft_state_machine.cpp#L204) and call up to HO, in HO we call the `on_pg_replace_member`.  That ensures the `on_pg_replace_member` be executed on every replica

xiaoxichen: but if we do that , we will get callback in both adding new member as well as remove old member

xiaoxichen (changes requested):

src/lib/homestore_backend/hs_pg_manager.cpp
```diff
@@ -145,9 +145,55 @@ void HSHomeObject::on_create_pg_message_commit(int64_t lsn, sisl::blob const& he
...
+    auto& pg = iter->second;
+    auto hs_pg = static_cast< HSHomeObject::HS_PG* >(pg.get());
+
+    // Update the pg info members.
+    pg->pg_info_.members.erase(PGMember(old_member));
+    pg->pg_info_.members.emplace(new_member);
+
+    uint32_t i{0};
+    for (auto const& m : pg->pg_info_.members) {
+        hs_pg->pg_sb_->members[i].id = m.id;
```
xiaoxichen, line 189: just a side note here,  if this function get called during adding new member (i.e 3-> 4). the s_pg->pg_sb_->members need to re-allocate to accommodate the new member. 

Or we can choose pre-allocate N+1 slot for the array.

src/lib/homestore_backend/hs_pg_manager.cpp
sanebay: Ack

src/lib/homestore_backend/hs_pg_manager.cpp
sanebay: This is for replace member. Expansion will be separate api and listener callback and different flow.

xiaoxichen (commented):
+1 from my side and waiting for the testing result

xiaoxichen:
As discussed in the meeting, we need to re-think the flow with the learner changes. My feeling is we need to wait till the learner flip to follower (voter) then do the step 3
```  
// Step 3. Append log entry to mark the old member is out and new member is added.
```
and setp 4
```
            // Step 4. Remove the old member. Even if the old member is temporarily
            // down and recovers, nuraft mesg see member remove from cluster log
            // entry and call exit_group() and leave().
```

Likely we need to hook up to the 
```
void RaftStateMachine::commit_config
```
to get notification  on cluster member change especially learner flipped to follower.

src/lib/homestore_backend/hs_pg_manager.cpp
```diff
@@ -145,9 +145,71 @@ void HSHomeObject::on_create_pg_message_commit(int64_t lsn, sisl::blob const& he
...
+    }
+
+    LOGI("PG replace member initated member_out={} member_in={}", boost::uuids::to_string(old_member_id),
+         boost::uuids::to_string(new_member.id));
+
+    homestore::replica_member_info in_replica, out_replica;
+    out_replica.id = old_member_id;
+    in_replica.id = new_member.id;
+    in_replica.priority = new_member.priority;
+    std::strncpy(in_replica.name, new_member.name.data(), in_replica.max_name_len);
```
xiaoxichen, line 172: Can we enforce the `len(name) <=31` to `PGMember` constructor?  and better add a constructor for `homestore::replica_member_info`.

`strncpy` is risky here as it can truncate make the `in_replca.name` not ended with `\0`.  Later in the 
`pg->pg_info_.members.emplace(PGMember(member_in.id, member_in.name, member_in.priority));` we construct a string by a char array doesnt ended with `\0`, the behavior is risky.

src/lib/homestore_backend/hs_pg_manager.cpp
```diff
@@ -145,9 +145,71 @@ void HSHomeObject::on_create_pg_message_commit(int64_t lsn, sisl::blob const& he
...
+        .thenValue([this](auto&& v) mutable -> PGManager::NullAsyncResult {
+            if (v.hasError()) { return folly::makeUnexpected(toPgError(v.error())); }
+            return folly::Unit();
+        });
+}
+
+void HSHomeObject::on_pg_replace_member(homestore::group_id_t group_id, const replica_member_info& member_out,
+                                        const replica_member_info& member_in) {
+    auto lg = std::shared_lock(_pg_lock);
+    for (const auto& iter : _pg_map) {
```
xiaoxichen, line 186: `_pg_map` is a map , should we use `find()` api?

src/lib/homestore_backend/hs_pg_manager.cpp
```diff
@@ -145,9 +145,71 @@ void HSHomeObject::on_create_pg_message_commit(int64_t lsn, sisl::blob const& he
...
+        auto& pg = iter.second;
+        if (pg_repl_dev(*pg).group_id() == group_id) {
+            // Remove the old member and add the new member
+            auto hs_pg = static_cast< HSHomeObject::HS_PG* >(pg.get());
+            pg->pg_info_.members.erase(PGMember(member_out.id));
+            pg->pg_info_.members.emplace(PGMember(member_in.id, member_in.name, member_in.priority));
+
+            uint32_t i{0};
+            for (auto const& m : pg->pg_info_.members) {
+                hs_pg->pg_sb_->members[i].id = m.id;
```
xiaoxichen, line 196: HomeObject `    struct pg_members ` is identical to `homestore::replica_member_info` as of now. 

As `homestore::replica_member_info` is the only information that follower can get regarding a PG,  I doubt there is a possibility `HomeObject::pg_members` can diverse from `homestore::replica_member_info`

Thinking about consolidate them together, not blocking,  more for discussion

xiaoxichen (commented):
Please if you can rebase.

src/lib/homestore_backend/hs_pg_manager.cpp
sanebay: pg_map is pgid to pg, here looking for group_id.

src/lib/homestore_backend/hs_pg_manager.cpp
sanebay: Made the max length to 127 and assert. Cant include homestore header files in pg_manager.hpp for maxlen.

sanebay:
> As discussed in the meeting, we need to re-think the flow with the learner changes. My feeling is we need to wait till the learner flip to follower (voter) then do the step 3
why should we wait ? why not treat as normal follower ?

xiaoxichen:
>why should we wait ? why not treat as normal follower ?

I think it is same when the out_member has down.  But if the out_member is healthy , i.e PG move for balancing/tech refresh etc,  keeping the redundancy has advantage. If we dont wait till new_member catch up before deleting out_member,  we made us degrade to 2 replica.

xiaoxichen (commented):
lgtm

sanebay:
> > why should we wait ? why not treat as normal follower ?
> 
> I think it is same when the out_member has down. But if the out_member is healthy , i.e PG move for balancing/tech refresh etc, keeping the redundancy has advantage. If we dont wait till new_member catch up before deleting out_member, we made us degrade to 2 replica.

Created a story for this. https://github.com/eBay/HomeStore/issues/573 This has to be optional depending on use case. Sometimes, its better to evict immediately to free up space.

xiaoxichen (approved):

xiaoxichen:
feel free to merge and as you said that can track by other ticket.

PR #217, opened 2024-10-16
Adaptive contract change for pg move API

yuwmao:
Based on the API proto change, we need to add some field in HomeObject.
https://github.corp.ebay.com/SDS/nuobject_proto/pull/14/files

### DIFF
--- a/src/include/homeobject/pg_manager.hpp
+++ b/src/include/homeobject/pg_manager.hpp
@@ -34,6 +34,7 @@ struct PGInfo {
     pg_id_t id;
     mutable MemberSet members;
     peer_id_t replica_set_uuid;
+    u_int64_t size;
 
     auto operator<=>(PGInfo const& rhs) const { return id <=> rhs.id; }
     auto operator==(PGInfo const& rhs) const { return id == rhs.id; }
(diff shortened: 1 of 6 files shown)


sanebay:
Some changes are already there in this PR https://github.com/eBay/HomeObject/pull/216/files because to compile , need to add those changes.

src/include/homeobject/pg_manager.hpp
```diff
@@ -76,7 +78,7 @@ struct PGStats {
 class PGManager : public Manager< PGError > {
 public:
     virtual NullAsyncResult create_pg(PGInfo&& pg_info) = 0;
-    virtual NullAsyncResult replace_member(pg_id_t id, peer_id_t const& old_member, PGMember const& new_member) = 0;
+    virtual NullAsyncResult replace_member(pg_id_t id, peer_id_t const& old_member, PGMember const& new_member, u_int32_t const& commit_quorum) = 0;
```
sanebay, line 81: why do we need const& for commit_qurorum. Simple uint32_t commit_quorum is enough.

yuwmao: Okay.

sanebay (approved):

PR #223, opened 2024-11-07
Update snapshot resync msg proto

yuwmao:
- Remove resync_payload since no need to use it
- Use blob state to replace is_deleted since the blob may be corrupted on the leader side. We need to highlight corrupted state because of two reasons:
  1. We decide to support in-place correcting data in the future, so at least we need to store the blob to occupy the space;
  2. Follower needs this state to differentiate it from the corrupted caused by network transmission.


xiaoxichen:
LGTM , this is an important edge case , better added into the doc (handling corrupted blob in leader)

xiaoxichen (dismissed):

koujl:
> LGTM , this is an important edge case , better added into the doc (handling corrupted blob in leader)

Sure. And we can add comments to the definition of the state enum constants as well (can be done in a future PR).

koujl (approved):

JacksonYao287 (dismissed):
looks good

JacksonYao287 (approved):

PR #226, opened 2024-11-11
support spare replicas in raft test framework

JacksonYao287:
this PR aims to support spare replicas in raft test framework , which is essential for testing pg move. what`s more, a basic replace_member UT is added

### DIFF
--- a/src/lib/homestore_backend/tests/test_homestore_backend_dynamic.cpp
+++ b/src/lib/homestore_backend/tests/test_homestore_backend_dynamic.cpp
@@ -1,0 +1,149 @@
+/*********************************************************************************
+ * Modifications Copyright 2017-2019 eBay Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *    https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed
+ * under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+ * CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ *
+ *********************************************************************************/
+/*
+ * Homeobject Replication testing binaries shared common definitions, apis and data structures
+ */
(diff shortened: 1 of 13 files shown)


src/lib/homestore_backend/tests/test_homestore_backend_dynamic.cpp
```diff
@@ -73,12 +73,15 @@ TEST_F(HomeObjectFixture, ReplaceMember) {
...
 
-    // the new member should wait until it joins the pg
+    // the new member should wait until it joins the pg and all the blobs are replicated to it
     if (in_member_id == g_helper->my_replica_id()) {
         while (!am_i_in_pg(pg_id)) {
             std::this_thread::sleep_for(std::chrono::milliseconds(1000));
             LOGINFO("new member is waiting to become a member of pg {}", pg_id);
         }
+
+        wait_for_all(pg_shard_id_vec[pg_id].back() /*the last shard id in this pg*/,
```
xiaoxichen, line 83: as discussed, this doesnt guarantee to be right once baseline resync is on , it doesnt sync follow the LSN order.

xiaoxichen (commented):
in general LGTm

src/lib/homestore_backend/tests/test_homestore_backend_dynamic.cpp
JacksonYao287: yes, but for now, we can not do this , since if a member is not leader , it can not get its own commit lsn
see 
https://github.com/eBay/HomeObject/blob/49b1b0cf8a3b9002a2a9eccc1bce878aa60ae680/src/lib/homestore_backend/hs_pg_manager.cpp#L396

and 
https://github.com/eBay/nuraft_mesg/blob/a277e8acae561a4b3873337bb179e32817f10b85/src/lib/repl_service_ctx.cpp#L92

will implement a new wait_for_all as soon as we can get commit_lsn at follower
cc @sanebay

xiaoxichen: OK,  it makes sense to add self lsn to `repl_service_ctx::get_raft_status()`.... When I implement this function I didnt realize we have this requirement./

sanebay (commented):
LGTM

sanebay (approved):

PR #227, opened 2024-11-15
Implement ReplicationStateMachine::write_snapshot_data

koujl:
* Implement `ReplicationStateMachine::write_snapshot_data()` and `SnapshotReceiveHandler` logic
* Extract `local_create_pg`, `local_create_shard` and `local_add_blob_info` for follower data creation

### DIFF
--- a/src/lib/homestore_backend/hs_homeobject.hpp
+++ b/src/lib/homestore_backend/hs_homeobject.hpp
@@ -337,7 +337,6 @@ public:
         shard_id_t get_next_shard() const;
 
         shard_id_t shard_cursor_{invalid_shard_id};
-        blob_id_t blob_cursor_{0};
         snp_batch_id_t cur_batch_num_{0};
         std::vector< shard_id_t > shard_list_;
 
--- a/src/lib/homestore_backend/hs_homeobject.hpp
+++ b/src/lib/homestore_backend/hs_homeobject.hpp
@@ -318,22 +318,34 @@ public:
 
     // SnapshotReceiverContext is the context used in follower side snapshot receiving. [drafting] The functions is not the final version.
     struct SnapshotReceiveHandler {
-        SnapshotReceiveHandler(HSHomeObject& home_obj, pg_id_t pg_id_, homestore::group_id_t group_id);
+        enum ErrorCode {
+            ALLOC_BLK_ERR = 1,
+            WRITE_DATA_ERR,
+            INVALID_BLOB_HEADER,
+            BLOB_DATA_CORRUPTED,
+            ADD_BLOB_INDEX_ERR,
+        };
+
+        constexpr static shard_id_t invalid_shard_id = 0;
+        constexpr static shard_id_t shard_list_end_marker = ULLONG_MAX;
+
+        SnapshotReceiveHandler(HSHomeObject& home_obj, pg_id_t pg_id_, homestore::group_id_t group_id, int64_t lsn,
+                               shared< homestore::ReplDev > repl_dev);
         void process_pg_snapshot_data(ResyncPGMetaData const& pg_meta);
-        void process_shard_snapshot_data(ResyncShardMetaData const& shard_meta, snp_obj_id_t& obj_id);
-        void process_blobs_snapshot_data(ResyncBlobDataBatch const& data_blob, snp_obj_id_t& obj_id);
+        int process_shard_snapshot_data(ResyncShardMetaData const& shard_meta);
+        int process_blobs_snapshot_data(ResyncBlobDataBatch const& data_blobs, snp_batch_id_t batch_num);
+        shard_id_t get_next_shard() const;
 
-        //snapshot start lsn
-        int64_t snp_lsn{0};
-        shard_id_t shard_cursor{0};
-        blob_id_t blob_cursor{0};
-        snp_batch_id_t cur_batch_num{0};
-        std::vector<shard_id_t> shard_list;
+        shard_id_t shard_cursor_{invalid_shard_id};
+        blob_id_t blob_cursor_{0};
+        snp_batch_id_t cur_batch_num_{0};
+        std::vector< shard_id_t > shard_list_;
 
+        const int64_t snp_lsn_{0};
         HSHomeObject& home_obj_;
-        homestore::group_id_t group_id_;
-        pg_id_t pg_id_;
-        shared< homestore::ReplDev > repl_dev_;
+        const homestore::group_id_t group_id_;
+        const pg_id_t pg_id_;
+        const shared< homestore::ReplDev > repl_dev_;
 
         //snapshot info, can be used as a checkpoint for recovery
         snapshot_info_superblk snp_info_;
@@ -357,17 +369,18 @@ private:
                                                     const homestore::MultiBlkId& blkid) const;
 
     // create pg related
-    PGManager::NullAsyncResult do_create_pg(cshared< homestore::ReplDev > repl_dev, PGInfo&& pg_info);
+    static PGManager::NullAsyncResult do_create_pg(cshared< homestore::ReplDev > repl_dev, PGInfo&& pg_info);
+    HS_PG* local_create_pg(shared< homestore::ReplDev > repl_dev, PGInfo pg_info);
     static std::string serialize_pg_info(const PGInfo& info);
     static PGInfo deserialize_pg_info(const unsigned char* pg_info_str, size_t size);
     void add_pg_to_map(unique< HS_PG > hs_pg);
 
     // create shard related
     shard_id_t generate_new_shard_id(pg_id_t pg);
-    uint64_t get_sequence_num_from_shard_id(uint64_t shard_id_t) const;
 
     static ShardInfo deserialize_shard_info(const char* shard_info_str, size_t size);
     static std::string serialize_shard_info(const ShardInfo& info);
+    void local_create_shard(ShardInfo shard_info, homestore::chunk_num_t chunk_num, homestore::blk_count_t blk_count);
     void add_new_shard_to_map(ShardPtr&& shard);
     void update_shard_in_map(const ShardInfo& shard_info);
 
@@ -453,6 +466,14 @@ public:
     std::optional< homestore::chunk_num_t > get_shard_chunk(shard_id_t id) const;
 
     /**
+     * @brief Get the sequence number of the shard from the shard id.
+     *
+     * @param shard_id The ID of the shard.
+     * @return The sequence number of the shard.
+     */
+    static uint64_t get_sequence_num_from_shard_id(uint64_t shard_id);
+
+    /**
      * @brief recover PG and shard from the superblock.
      *
      */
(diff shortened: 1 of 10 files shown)


src/lib/homestore_backend/snapshot_receive_handler.cpp
```diff
@@ -0,0 +1,213 @@
...
+        }
+        RELEASE_ASSERT(index_table != nullptr, "Index table instance null");
+        if (home_obj_.get_blob_from_index_table(index_table, shard_cursor_, blob->blob_id())) {
+            LOGD("Skip already persisted blob_id:{}", blob->blob_id());
+            continue;
+        }
+
+        auto blob_data = blob->data()->Data();
+        auto header = r_cast< BlobHeader const* >(blob_data);
+        if (header->valid()) {
```
yuwmao, line 124: nit: `if (!header->valid())`. 
And we consider the blob with invalid header as corrupted as well, so should not return error in this branch.

src/lib/homestore_backend/snapshot_receive_handler.cpp
```diff
@@ -0,0 +1,213 @@
...
+
+        // Add local blob info to index & PG
+        bool success = home_obj_.local_add_blob_info(pg_id_, BlobInfo{shard_cursor_, blob->blob_id(), {0, 0, 0}});
+        if (!success) {
+            LOGE("Failed to add blob info for blob_id:{}", blob->blob_id());
+            free_allocated_blks();
+            return ADD_BLOB_INDEX_ERR;
+        }
+
+        blob_cursor_ = blob->blob_id();
```
yuwmao, line 193: What is the purpose of this `blob_cursor_`?

src/lib/homestore_backend/snapshot_receive_handler.cpp
koujl: Nice catch, moved the header validation into the if statement body.

src/lib/homestore_backend/snapshot_receive_handler.cpp
koujl: Previously thought it might be useful for progress tracking or resuming... but doesn't really seem to be.
Let me remove it for now.

src/lib/homestore_backend/hs_pg_manager.cpp
```diff
@@ -93,6 +94,37 @@ PGManager::NullAsyncResult HSHomeObject::do_create_pg(cshared< homestore::ReplDe
...
+    {
+        auto lg = shared_lock(_pg_lock);
+        if (auto it = _pg_map.find(pg_id); it != _pg_map.end()) {
+            LOGW("PG already exists, pg_id {}", pg_id);
+            return dynamic_cast< HS_PG* >(it->second.get());
+        }
+    }
+
+    // create index table and pg
+    // TODO create index table during create shard.
```
yamingk, line 108: we may relax this TODO, is there a need to create index during shard create?

src/lib/homestore_backend/replication_state_machine.cpp
```diff
@@ -266,38 +266,101 @@ int ReplicationStateMachine::read_snapshot_data(std::shared_ptr< homestore::snap
...
+                                  obj_id.shard_seq_num, obj_id.batch_id, snp_data->blob.size());
+
+    if (snp_data->is_last_obj) {
+        LOGD("Write snapshot reached is_last_obj true {}", log_suffix);
+        return;
+    }
+
+    // Check if the snapshot context is same as the current snapshot context.
+    // If not, drop the previous context and re-init a new one
+    if (m_snp_rcv_handler && m_snp_rcv_handler->snp_lsn_ != context->get_lsn()) { m_snp_rcv_handler.reset(nullptr); }
```
yamingk, line 286: I don't think I follower in which case we should reuse this m_snp_rcv_handler. Every snapshot context will have distinguished lsn, no? Or there is a case two snapshots will have the same lsn?

src/lib/homestore_backend/replication_state_machine.cpp
```diff
@@ -266,38 +266,101 @@ int ReplicationStateMachine::read_snapshot_data(std::shared_ptr< homestore::snap
...
+
+    // Check message integrity
+    auto header = r_cast< const SyncMessageHeader* >(snp_data->blob.cbytes());
```
yamingk, line 289: We probably need to have error handling either to crash or raise some alert to SRE. Logging error will not stop leader to keep resyncing to this member, right? If the leader keeps resyncing to this member and this member finishes resyncing, is there a way for this PG to be alerted, or we just leave the unwritten data there in this PG and let next read to hit on this PG this blob to report BLOB not found. 

Just try to understand what should be our strategy as of this error handling. Botton line we should not return wrong blob to end-user, failing the baseline resync or to start over or raise alert are all possible solutions.

src/lib/homestore_backend/hs_homeobject.hpp
```diff
@@ -318,22 +318,33 @@ class HSHomeObject : public HomeObjectImpl {
...
-        int64_t snp_lsn{0};
-        shard_id_t shard_cursor{0};
-        blob_id_t blob_cursor{0};
-        snp_batch_id_t cur_batch_num{0};
-        std::vector<shard_id_t> shard_list;
+        shard_id_t shard_cursor_{invalid_shard_id};
+        snp_batch_id_t cur_batch_num_{0};
+        std::vector< shard_id_t > shard_list_;
 
+        const int64_t snp_lsn_{0};
```
yamingk, line 343: I didn't see this is being used expect for the check to reuse this handler instance. Can we remove this field and the check and always reuse this handler instance?

koujl: The follower may receive a new snapshot transmission request if the leader crashes or if the previous snapshot transmission times out. The current LSN is necessary for the follower to recognize such cases and discard the current handler instance.

src/lib/homestore_backend/hs_pg_manager.cpp
koujl: Good point, based on our current design it seems no longer necessary.

src/lib/homestore_backend/replication_state_machine.cpp
koujl: On errors we will retain the input `snp_data->offset` in order to trigger a resend of the current batch. Since message corruptions should be random, we expect that we'll eventually receive the correct data.

I'm not sure if there'll be any situations where the leader will keep generating & sending invalid data... but let's leave it for when they arise in the future.

src/lib/homestore_backend/snapshot_receive_handler.cpp
```diff
@@ -120,17 +118,18 @@ int HSHomeObject::SnapshotReceiveHandler::process_blobs_snapshot_data(ResyncBlob
...
-            return INVALID_BLOB_HEADER;
-        }
-        std::string user_key = header->user_key_size
-            ? std::string(r_cast< const char* >(blob_data + sizeof(BlobHeader)), header->user_key_size)
-            : std::string{};
 
         // Check integrity of normal blobs
         if (blob->state() != static_cast< uint8_t >(ResyncBlobState::CORRUPTED)) {
+            auto header = r_cast< BlobHeader const* >(blob_data);
+            if (!header->valid()) {
```
yuwmao, line 125: Please also check if `blob state != ResyncBlobState::CORRUPTED`. Because if the header is corrupted on the leader side, leader will treat it as corrupted and send it as it, for this case, just save it and correct later.

src/lib/homestore_backend/replication_state_machine.cpp
yamingk: It looks good to me if leader will just resend based on the `snp_data->offset`, can we set a flip point here simulating the header crc error and let leader to resend the same batch?

src/lib/homestore_backend/hs_shard_manager.cpp
```diff
@@ -58,8 +58,8 @@ shard_id_t HSHomeObject::generate_new_shard_id(pg_id_t pgid) {
     return make_new_shard_id(pgid, new_sequence_num);
 }
 
-uint64_t HSHomeObject::get_sequence_num_from_shard_id(uint64_t shard_id_t) const {
-    return shard_id_t & (max_shard_num_in_pg() - 1);
+uint64_t HSHomeObject::get_sequence_num_from_shard_id(uint64_t shard_id) {
```
JacksonYao287, line 61: why remove `const`?

src/lib/homestore_backend/replication_state_machine.cpp
```diff
@@ -266,38 +266,101 @@ int ReplicationStateMachine::read_snapshot_data(std::shared_ptr< homestore::snap
...
+    if (header->corrupted()) {
+        LOGE("corrupted message in write_snapshot_data, lsn:{}, obj_id {} shard {} batch {}", s->get_last_log_idx(),
+             obj_id.value, obj_id.shard_seq_num, obj_id.batch_id);
+        return;
+    }
+    if (auto payload_size = snp_data->blob.size() - sizeof(SyncMessageHeader); payload_size != header->payload_size) {
+        LOGE("payload size mismatch in write_snapshot_data {} != {}, lsn:{}, obj_id {} shard {} batch {}", payload_size,
+             header->payload_size, s->get_last_log_idx(), obj_id.value, obj_id.shard_seq_num, obj_id.batch_id);
+        return;
```
JacksonYao287, line 298: should we returen a false or a true in line 293 and 298 to indicate if the payload and header is good ? if not , maybe we should let leader resend it

src/lib/homestore_backend/replication_state_machine.cpp
```diff
@@ -266,38 +266,101 @@ int ReplicationStateMachine::read_snapshot_data(std::shared_ptr< homestore::snap
...
+        } else {
+            // Request for the next batch
+            snp_data->offset = objId(obj_id.shard_seq_num, 1).value;
+        }
+        LOGD("Write snapshot, processed shard data shard_seq_num:{} {}", obj_id.shard_seq_num, log_suffix);
+        return;
+    }
+
+    // Blob data message
+    RELEASE_ASSERT(obj_id.shard_seq_num ==
```
JacksonYao287, line 339: should we add some logic to handle this case. for example ,skip this and resend the same request to leader.

src/lib/homestore_backend/snapshot_receive_handler.cpp
```diff
@@ -0,0 +1,210 @@
...
+    shard_sb.info.created_time = shard_meta.created_time();
+    shard_sb.info.last_modified_time = shard_meta.last_modified_time();
+    shard_sb.info.available_capacity_bytes = shard_meta.total_capacity_bytes();
+    shard_sb.info.total_capacity_bytes = shard_meta.total_capacity_bytes();
+    shard_sb.info.deleted_capacity_bytes = 0;
+    shard_sb.chunk_id = chunk_id;
+
+    homestore::MultiBlkId blk_id;
+    const auto hints = home_obj_.chunk_selector()->chunk_to_hints(chunk_id);
+    auto status = homestore::data_service().alloc_blks(
```
JacksonYao287, line 63: here , we need get physical chunk of the vchunk, and put in in the hint

koujl:
Added several TODOs as reminders for suggestions mentioned in the comments.

Further optimization & refactor will be posted within next PR.

JacksonYao287 (dismissed):
LGTM

yuwmao (approved):

PR #229, opened 2024-11-20
Fix HeapChunkSelector::get_most_available_blk_chunk

Hooper9973:
Fix HeapChunkSelector::get_most_available_blk_chunk

1. Add check to ensure max_it points to an available chunk.
   If not, it indicates there are no available chunks left.
2. Add safeguards in the create and recover pg paths to ensure
   the pg size cannot be zero.

### DIFF
--- a/src/lib/homestore_backend/heap_chunk_selector.cpp
+++ b/src/lib/homestore_backend/heap_chunk_selector.cpp
@@ -129,6 +129,10 @@ std::optional< uint32_t > HeapChunkSelector::select_chunks_for_pg(pg_id_t pg_id,
         LOGWARNMOD(homeobject, "PG had already created, pg_id {}", pg_id);
         return std::nullopt;
     }
+    if (pg_size == 0) {
+        LOGWARNMOD(homeobject, "Not supported to create empty PG, pg_id {}, pg_size {}", pg_id, pg_size);
+        return std::nullopt;
+    }
 
     const auto chunk_size = get_chunk_size();
     const uint32_t num_chunk = sisl::round_down(pg_size, chunk_size) / chunk_size;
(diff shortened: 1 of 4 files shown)


JacksonYao287 (approved):
LG

src/lib/homestore_backend/heap_chunk_selector.cpp
```diff
@@ -293,7 +296,10 @@ std::optional< homestore::chunk_num_t > HeapChunkSelector::get_most_available_bl
                              if (a->available() && !b->available()) { return false; }
                              return false;
                          });
-
+    if (!(*max_it)->available()) {
```
Hooper9973, line 299: @xiaoxichen  check if have available chunk here.
 And this pr make sure creating an empty PG is not supported, so we neednt check if `m_pg_chunks` is empty.

src/lib/homestore_backend/heap_chunk_selector.cpp
```diff
@@ -293,7 +296,10 @@ std::optional< homestore::chunk_num_t > HeapChunkSelector::get_most_available_bl
                              if (a->available() && !b->available()) { return false; }
                              return false;
```
JacksonYao287, line 297: NIT: can we simplify this part as 
```
auto max_it =
        std::max_element(pg_chunks.begin(), pg_chunks.end(),
                         [](const std::shared_ptr< ExtendedVChunk >& a, const std::shared_ptr< ExtendedVChunk >& b) {
                             return !a->available() || b->available() &&  a->available_blks() < b->available_blks();

```

JacksonYao287 (approved):
LGTM， only a minor comment

src/lib/homestore_backend/heap_chunk_selector.cpp
Hooper9973: sure, refine it

xiaoxichen (approved):

PR #232, opened 2024-11-26
Refactor SnapshotReceiveHandler & Add UT

koujl:
* Extract `SnapshotContext` from `SnapshotReceiveHandler` to split constant and mutable members
* Add UT for `SnapshotReceiveHandler` and fix some bugs

### DIFF
--- a/src/lib/homestore_backend/hs_homeobject.hpp
+++ b/src/lib/homestore_backend/hs_homeobject.hpp
@@ -329,8 +329,7 @@ public:
         constexpr static shard_id_t invalid_shard_id = 0;
         constexpr static shard_id_t shard_list_end_marker = ULLONG_MAX;
 
-        SnapshotReceiveHandler(HSHomeObject& home_obj, homestore::group_id_t group_id,
-                               shared< homestore::ReplDev > repl_dev);
+        SnapshotReceiveHandler(HSHomeObject& home_obj, shared< homestore::ReplDev > repl_dev);
 
         void process_pg_snapshot_data(ResyncPGMetaData const& pg_meta);
         int process_shard_snapshot_data(ResyncShardMetaData const& shard_meta);
@@ -349,12 +348,12 @@ public:
             std::vector< shard_id_t > shard_list;
             const int64_t snp_lsn;
             const pg_id_t pg_id;
+            shared< BlobIndexTable > index_table;
 
             SnapshotContext(int64_t lsn, pg_id_t pg_id) : snp_lsn{lsn}, pg_id{pg_id} {}
         };
 
         HSHomeObject& home_obj_;
-        const homestore::group_id_t group_id_;
         const shared< homestore::ReplDev > repl_dev_;
 
         std::unique_ptr< SnapshotContext > ctx_;
--- a/src/lib/homestore_backend/hs_homeobject.hpp
+++ b/src/lib/homestore_backend/hs_homeobject.hpp
@@ -333,7 +333,8 @@ public:
 
         void process_pg_snapshot_data(ResyncPGMetaData const& pg_meta);
         int process_shard_snapshot_data(ResyncShardMetaData const& shard_meta);
-        int process_blobs_snapshot_data(ResyncBlobDataBatch const& data_blobs, snp_batch_id_t batch_num);
+        int process_blobs_snapshot_data(ResyncBlobDataBatch const& data_blobs, snp_batch_id_t batch_num,
+                                        bool is_last_batch);
 
         int64_t get_context_lsn() const;
         void reset_context(int64_t lsn, pg_id_t pg_id);
--- a/src/lib/homestore_backend/hs_homeobject.hpp
+++ b/src/lib/homestore_backend/hs_homeobject.hpp
@@ -316,8 +316,8 @@ public:
         uint64_t max_batch_size_;
     };
 
-    // SnapshotReceiverContext is the context used in follower side snapshot receiving. [drafting] The functions is not the final version.
-    struct SnapshotReceiveHandler {
+    class SnapshotReceiveHandler {
+    public:
         enum ErrorCode {
             ALLOC_BLK_ERR = 1,
             WRITE_DATA_ERR,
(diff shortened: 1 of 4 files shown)


src/lib/homestore_backend/tests/hs_blob_tests.cpp
```diff
@@ -285,4 +287,135 @@ TEST_F(HomeObjectFixture, PGBlobIterator) {
...
+    // Step 2: Test shard and blob batches
+    blob_id_t cur_blob_id{0};
+    for (uint64_t i = 1; i <= num_shards_per_pg; i++) {
+        LOGINFO("TESTING: applying meta for shard {}", i);
+
+        // Step 2-1: Test write shard meta
+        // Generate ResyncShardMetaData message
+        ShardInfo shard;
+        shard.id = i;
+        shard.state = ShardInfo::State::OPEN;
```
yuwmao, line 341: Suggest covering the sealed shard and delete shard process. In the current implementation, handler will call `home_obj_.local_create_shard` for all shards no matter their states, but `local_create_shard` will remove the selected chunk from the heap. So seems handler should call seal_shard for sealed shard. Please correct me if I missed some logic.

src/lib/homestore_backend/hs_homeobject.hpp
```diff
@@ -329,24 +329,37 @@ class HSHomeObject : public HomeObjectImpl {
         constexpr static shard_id_t invalid_shard_id = 0;
         constexpr static shard_id_t shard_list_end_marker = ULLONG_MAX;
 
-        SnapshotReceiveHandler(HSHomeObject& home_obj, pg_id_t pg_id_, homestore::group_id_t group_id, int64_t lsn,
+        SnapshotReceiveHandler(HSHomeObject& home_obj, homestore::group_id_t group_id,
```
xiaoxichen, line 332: nit: if group_id can get from repl_dev, it can be removed from parameters

src/lib/homestore_backend/snapshot_receive_handler.cpp
```diff
@@ -107,12 +107,12 @@ int HSHomeObject::SnapshotReceiveHandler::process_blobs_snapshot_data(ResyncBlob
         shared< BlobIndexTable > index_table;
         {
             std::shared_lock lock_guard(home_obj_._pg_lock);
-            auto iter = home_obj_._pg_map.find(pg_id_);
+            auto iter = home_obj_._pg_map.find(ctx_->pg_id);
```
xiaoxichen, line 110: can the index_table be put into ctx so you dont need to take it every time?

src/lib/homestore_backend/snapshot_receive_handler.cpp
```diff
@@ -107,12 +107,12 @@ int HSHomeObject::SnapshotReceiveHandler::process_blobs_snapshot_data(ResyncBlob
         shared< BlobIndexTable > index_table;
```
xiaoxichen, line 107: please add comments to explain where the duplication comes from,  they should from resent batch only if we always start from pristine

xiaoxichen (commented):
lgtm aside from a few nit

There is a comment from Yaming to add flip points into each step, please create ticket to track that if we want to finish the code first

src/lib/homestore_backend/snapshot_receive_handler.cpp
```diff
@@ -59,13 +58,14 @@ int HSHomeObject::SnapshotReceiveHandler::process_shard_snapshot_data(ResyncShar
     shard_sb.chunk_id = chunk_id;
 
     homestore::MultiBlkId blk_id;
-    const auto hints = home_obj_.chunk_selector()->chunk_to_hints(chunk_id);
+    const auto hints = home_obj_.chunk_selector()->chunk_to_hints(shard_sb.chunk_id);
     auto status = homestore::data_service().alloc_blks(
         sisl::round_up(sizeof(shard_sb), homestore::data_service().get_blk_size()), hints, blk_id);
     if (status != homestore::BlkAllocStatus::SUCCESS) {
         LOGE("Failed to allocate blocks for shard {}", shard_meta.shard_id());
```
JacksonYao287, line 65: not sure will we always baseline resync to a brand new member? if not, maybe we need to let emergency gc kick in here

src/lib/homestore_backend/snapshot_receive_handler.cpp
```diff
@@ -59,13 +58,14 @@ int HSHomeObject::SnapshotReceiveHandler::process_shard_snapshot_data(ResyncShar
...
         return ALLOC_BLK_ERR;
     }
+    shard_sb.chunk_id = blk_id.to_single_blkid().chunk_num(); // FIXME: remove this after intergating vchunk
 
     const auto ret = homestore::data_service()
```
JacksonYao287, line 70: NIT: maybe we can use `async_alloc_write` to combine line 62 and line 70

src/lib/homestore_backend/tests/hs_blob_tests.cpp
koujl: Thanks, nice catch. Fixed in latest commit.

Since currently it requires to change the method signature of `process_shard_snapshot_data` to know the allocated chunk ID, we'll finish the test after rebasing onto identical layout commits.

src/lib/homestore_backend/snapshot_receive_handler.cpp
koujl: Good suggestion, but seems it need several extra lines to build the input sg_list from buf. Will check later if there's any simplier way to do it.

src/lib/homestore_backend/replication_state_machine.cpp
```diff
@@ -268,10 +268,13 @@ void ReplicationStateMachine::write_snapshot_data(std::shared_ptr< homestore::sn
                                                   std::shared_ptr< homestore::snapshot_data > snp_data) {
     RELEASE_ASSERT(context != nullptr, "Context null");
     RELEASE_ASSERT(snp_data != nullptr, "Snapshot data null");
-    auto s = dynamic_pointer_cast< homestore::nuraft_snapshot_context >(context)->nuraft_snapshot();
-    auto obj_id = objId(static_cast< snp_obj_id_t >(snp_data->offset));
     auto r_dev = repl_dev();
+    if (!m_snp_rcv_handler) {
+        std::make_unique< HSHomeObject::SnapshotReceiveHandler >(*home_object_, r_dev->group_id(), r_dev);
```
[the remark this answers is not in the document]
koujl: Thanks for catching this! Fixed in latest push.

JacksonYao287 (approved):
LGTM

PR #233, opened 2024-11-28
fix create_snapshot

Besroy:
m_snapshot_context is nullptr in the first create_snapshot, add check to avoid crash.


xiaoxichen (approved):

JacksonYao287 (approved):
LGTM

PR #235, opened 2024-12-02
Fix and Improve `fix pg size & shard identical layout`

Hooper9973:
Bug Fixes:
- Resolved issues in create_shard and seal_shard functions to ensure correct shard management. Enhancements:
- Added validation to prevent creation of placement groups with sizes smaller than the chunk size. Test Improvements:
- pg test:
	- Added tests to verify that creating a placement group with a size smaller than the chunk size results in failure.
- shard test:
	- Created shard but shard size is larger than pg space left which should be failed.
	- Verified that successfully created shards have identical virtual chunk layouts.

### DIFF
--- a/src/lib/homestore_backend/tests/hs_repl_test_helper.hpp
+++ b/src/lib/homestore_backend/tests/hs_repl_test_helper.hpp
@@ -43,6 +43,7 @@ namespace bip = boost::interprocess;
 using namespace homeobject;
 
 #define INVALID_UINT64_ID UINT64_MAX
+#define INVALID_CHUNK_NUM UINT16_MAX
 
 namespace test_common {
 
@@ -56,6 +57,7 @@ protected:
                 sync_point_num_ = sync_point;
                 homeobject_replica_count_ = 0;
                 uint64_id_ = INVALID_UINT64_ID;
+                v_chunk_id_ = INVALID_CHUNK_NUM;
                 cv_.notify_all();
             } else {
                 cv_.wait(lg, [this, sync_point]() { return sync_point_num_ == sync_point; });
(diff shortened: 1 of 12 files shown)


src/lib/homestore_backend/tests/hs_repl_test_helper.hpp
```diff
@@ -56,6 +57,7 @@ class HSReplTestHelper {
                 sync_point_num_ = sync_point;
                 homeobject_replica_count_ = 0;
                 uint64_id_ = INVALID_UINT64_ID;
+                v_chunk_id_ = INVALID_CHUNK_NUM;
```
JacksonYao287, line 60: NIT: I prefer to  cast  homestore::chunk_num_t(uinit_16) to uint_64 here, and then cast it back.

JacksonYao287 (approved):
looks good, only a minor comment

src/lib/homestore_backend/tests/hs_repl_test_helper.hpp
Hooper9973: I'll merge first, and if necessary,  I will refine in next pr.
BTW, what are the pros in this way? 😃

JacksonYao287: hs_repl_test_helper.hpp  is the framework, it does not know what is a shard_id_t , pg_id_t , blob_id_t and chunk_id_t. these types are only know application layer.

if we add v_chunk_id here, should we also add  shard_id_t, pg_id_t and others? 

so I think we can use this 64bit as a common shared data across different processed, and application layer should translate this to what it wants

PR #244, opened 2024-12-17
Baseline Resync Enhancement

yuwmao:
This PR enhances baseline resync:
1. Address blob duplication issues caused by resync, it mainly happens when resend snapshot mesg during baseline resync and apply log after snapshot completion. This helps avoid unnecessary GC due to duplicated data. 
This mechanism is effective only for duplicated blobs. Since we do not record the blks of the `shardInfo` stored in the data service, we are unable to skip data writes for duplicated shards.
2. Reset PG and context at the beginning of the baseline resync if pg is already present.

### DIFF
--- a/src/lib/homestore_backend/hs_blob_manager.cpp
+++ b/src/lib/homestore_backend/hs_blob_manager.cpp
@@ -38,6 +38,8 @@ BlobError toBlobError(ReplServiceError const& e) {
     case ReplServiceError::RESULT_NOT_EXIST_YET:
         [[fallthrough]];
     case ReplServiceError::TERM_MISMATCH:
+        [[fallthrough]];
+    case ReplServiceError::DATA_DUPLICATED:
         return BlobError(BlobErrorCode::REPLICATION_ERROR);
     case ReplServiceError::NOT_LEADER:
         return BlobError(BlobErrorCode::NOT_LEADER);
@@ -111,6 +113,7 @@ BlobManager::AsyncResult< blob_id_t > HSHomeObject::_put_blob(ShardInfo const& s
     req->header()->payload_crc = 0;
     req->header()->shard_id = shard.id;
     req->header()->pg_id = pg_id;
+    req->header()->blob_id = new_blob_id;
     req->header()->seal();
 
     // Serialize blob_id as Replication key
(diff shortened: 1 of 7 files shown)


yuwmao:
This PR is in `draft` state, blocked by the [corresponding homestore change](https://github.com/eBay/HomeStore/pull/611), and tests should be added.

src/lib/homestore_backend/hs_shard_manager.cpp
```diff
@@ -379,17 +379,6 @@ void HSHomeObject::on_shard_message_commit(int64_t lsn, sisl::blob const& h, hom
         local_create_shard(shard_info, v_chunk_id, blkids.chunk_num(), blkids.blk_count());
         if (ctx) { ctx->promise_.setValue(ShardManager::Result< ShardInfo >(shard_info)); }
 
-        // update pg's total_occupied_blk_count
```
yuwmao, line 382: Redundant, these codes are already put into the local_create_shard.

src/lib/homestore_backend/hs_blob_manager.cpp
```diff
@@ -354,6 +360,13 @@ HSHomeObject::blob_put_get_blk_alloc_hints(sisl::blob const& header, cintrusive<
         return folly::makeUnexpected(homestore::ReplServiceError::FAILED);
     }
 
+    auto pg_iter = _pg_map.find(msg_header->pg_id);
```
JacksonYao287, line 363: NIT: can we change this part to `is_pg_present`

JacksonYao287 (commented):
lgtm

src/lib/homestore_backend/hs_blob_manager.cpp
yuwmao: Sure, thanks

yuwmao: Here we need to get the pg_iter as we are going to check index, so we can't use `is_pg_present` directly.

yuwmao: Found there is already has a `pg_exists` function, remove `is_pg_present`.

src/lib/homestore_backend/index_kv.cpp
```diff
@@ -48,8 +49,17 @@ std::pair< bool, homestore::btree_status_t > HSHomeObject::add_to_index_table(sh
...
     if (status != homestore::btree_status_t::success) {
+        if (existing_value.pbas().is_valid() && existing_value.pbas() == blob_info.pbas) {
+            LOGT(
+                "blob already exists, but existing pbas is the same as the new pbas, ignore it, blob_id={}, pbas={}, status={}",
+                blob_info.blob_id, blob_info.pbas.to_string(), status);
+            return {true, homestore::btree_status_t::success};
+        }
         if (existing_value.pbas().is_valid() || existing_value.pbas() == tombstone_pbas) {
             // Check if the blob id already exists in the index or its tombstone.
+            LOGW(
```
xiaoxichen, line 60: error even Panic if we dont know a valid case

xiaoxichen (approved):

koujl (approved):

PR #248, opened 2024-12-26
Implementation of Baseline Resync

koujl:
This PR contains all changes from the branch `baseline_resync`. 

The commit history has been reworked for simplicity and clarity.


yuwmao (dismissed):
LGTM

src/lib/homestore_backend/replication_message.hpp
```diff
@@ -11,43 +11,69 @@ namespace homeobject {
...
 static constexpr uint32_t HOMEOBJECT_REPLICATION_PROTOCOL_VERSION_V1 = 0x01;
+static constexpr uint32_t HOMEOBJECT_RESYNC_PROTOCOL_VERSION_V1 = 0x01;
 static constexpr uint32_t init_crc32 = 0;
+static constexpr uint64_t LAST_OBJ_ID =ULLONG_MAX;
+static constexpr uint64_t DEFAULT_MAX_BATCH_SIZE_MB =128;
 
 #pragma pack(1)
-struct ReplicationMessageHeader {
+template<typename Header>
+struct BaseMessageHeader {
```
Hooper9973, line 29: I was wondering if it might be worth considering an alternative approach here. Perhaps we could simplify things by declaring a base class, `BaseMessageHeader`, with the relevant interfaces as `virtual`. Then, the derived classes could `override` these implementations. I'm curious to hear your thoughts on this idea and whether it might offer any benefits in terms of clarity and maintainability.

src/lib/homestore_backend/hs_shard_manager.cpp
```diff
@@ -303,6 +303,36 @@ void HSHomeObject::on_shard_message_rollback(int64_t lsn, sisl::blob const& head
     }
 }
 
+// FIXME: Bugfix in progress from Hooper, will fix later.
```
Hooper9973, line 306: NIT: Already fixed, it can be removed now. 😄

src/lib/homestore_backend/replication_message.hpp
Hooper9973: If there are no scenarios where a BaseMessageHeader object is used directly, you might consider declaring getSize() as a pure virtual function to ensure that subclasses must implement it. Just a personal suggestion, feel free to proceed in the way you prefer.

```
class BaseMessageHeader {
    uint32_t calculate_crc() const {
        return crc32_ieee(init_crc32, reinterpret_cast<const unsigned char*>(this), getSize());
    }
    virtual size_t getSize() const = 0; 
};

class DerivedMessageHeader : public BaseMessageHeader {
size_t getSize() const override {
        return sizeof(DerivedMessageHeader);
    }
```

yuwmao: We tried that before, it will cause an error when we perform `reinterpret_cast` the header in the receiver side, because when using `sizeof` and filling the buffer, the virtual function table will taken into it as well.

src/lib/homestore_backend/hs_blob_manager.cpp
```diff
@@ -194,54 +242,19 @@ void HSHomeObject::on_blob_put_commit(int64_t lsn, sisl::blob const& header, sis
...
-                       // still get the up-to-date blob_sequence_num
-                       !de.blob_sequence_num.compare_exchange_weak(existing_blob_id, next_blob_id)) {}
-                de.active_blob_count.fetch_add(1, std::memory_order_relaxed);
-                de.total_occupied_blk_count.fetch_add(pbas.blk_count(), std::memory_order_relaxed);
-            });
-        }
+    bool success = local_add_blob_info(pg_id, blob_info);
+
+    if (ctx) {
+        ctx->promise_.setValue(success ? BlobManager::Result< BlobInfo >(blob_info)
```
Besroy, line 255: Could you pls help me understand why change the condition from 'status != success && !exist_alreay' to 'status!=success'? If add_to_index_table is failed because of already_exist, what is expect to happen? (or will it happend?)

yuwmao: Previously, if the blob already exists in index table, `add_to_index_table` fails but `on_blob_put_commit` returns success. I think it is intended to pass the replay logs scenario. But actually, there might be 2 cases: pbas is the same as the existing pbas and pbas is a different one(maybe by some mistakes). Currently, in `add_to_index_table` func, if the pbas is the same as the old one, it will return success, so we don't need `&& !exist_already` condition.

koujl: And actually it's not quite clear at this point in what cases we may encounter a conflict value in the index table. Let's wait for further tests to uncover more corner cases and optimize our handling accordingly.

src/lib/homestore_backend/pg_blob_iterator.cpp
```diff
@@ -2,26 +2,73 @@
...
-            auto sequence_num = home_obj_.get_sequence_num_from_shard_id(shard->info.id);
-            if (shard->info.lsn <= upto_lsn) { max_shard_seq_num_ = std::max(max_shard_seq_num_, sequence_num); }
+            if (shard->info.lsn <= upto_lsn) {
+                auto v_chunk_id = home_obj_.get_shard_v_chunk_id(shard->info.id);
+                shard_list_.emplace_back(shard->info, v_chunk_id.value());
+            }
         }
+        // sort shard list by <vchunkid, lsn> to ensure open shards positioned after sealed shards
+        std::ranges::sort(shard_list_, [](ShardEntry& a, ShardEntry& b) {
+            return a.v_chunk_num != b.v_chunk_num ? a.v_chunk_num < b.v_chunk_num : a.info.lsn < b.info.lsn;
```
Besroy, line 32: If I understood correctly, here assume shards with smaller v_chunk_num is more likely to be sealed. Could you pls explain the reason? (IIRC, shard is selected by available size)

src/lib/homestore_backend/hs_shard_manager.cpp
koujl: Thx, removed in latest push.

src/lib/homestore_backend/pg_blob_iterator.cpp
koujl: We don't care about the shard sequence across different chunks. This is just to ensure that **within each chunk** the opened shard is positioned last.

Let me edit the comment for clarity in the last push.

koujl:
To keep the commit log history clean, all additional changes in this PR will be placed in the latest commit `Preparation for merging into main`.

src/lib/homestore_backend/replication_state_machine.cpp
```diff
@@ -238,135 +238,219 @@ std::shared_ptr< homestore::snapshot_context > ReplicationStateMachine::last_sna
...
+    if (obj_id.shard_seq_num == 0) {
+        // PG metadata & shard list message
+        RELEASE_ASSERT(obj_id.batch_id == 0, "Invalid obj_id");
+
+        auto pg_data = GetSizePrefixedResyncPGMetaData(data_buf);
+
+        //Check if pg exists, if yes, clean the stale pg resources, may be due to previous snapshot failure. Let's resync on a pristine base
+        if (home_object_->pg_exists(pg_data->pg_id())) {
+            LOGI("pg already exists, clean pg resources before snapshot, pg_id:{} {}", pg_data->pg_id(), log_suffix);
+            home_object_->pg_destroy(pg_data->pg_id());
```
Besroy, line 369: Is this the reason for src/lib/homestore_backend/hs_blob_manager.cpp:L363-369?

Besroy:
I have an additional question regarding the baseline rsync: If the leader considers timeout during the baseline rsync, what will happen in leader and follower side? If leader resends snapshot, will the follower recreate a new snapshot? Or leader will not resend --- raft will try to re-read-save the snapshot obj until the obj is successfully written or fail?

src/lib/homestore_backend/replication_state_machine.cpp
yuwmao: `src/lib/homestore_backend/hs_blob_manager.cpp:L363-369` is more like a general defensive safety check, otherwise if pg is not exists, it can not proceed. Similar to the existing shard check.

koujl:
> I have an additional question regarding the baseline rsync: If the leader considers timeout during the baseline rsync, what will happen in leader and follower side? If leader resends snapshot, will the follower recreate a new snapshot? Or leader will not resend --- raft will try to re-read-save the snapshot obj until the obj is successfully written or fail?

On timeout, leader will drop the current `snapshot_sync_ctx` and initiate a new job from the beginning, using the latest snapshot at that point.

For the follower, upon receiving a new resync request with objId=0, currently it will reset all the data in PG and restart from the beginning as well - note that this behavior might be changed in the future with the implementation of progress resumption.

src/lib/homestore_backend/replication_state_machine.cpp
```diff
@@ -238,135 +238,219 @@ std::shared_ptr< homestore::snapshot_context > ReplicationStateMachine::last_sna
...
+    if (!pg_iter->create_blobs_snapshot_data(snp_obj->blob)) {
+        LOGE("Failed to create blob batch data for snapshot read, {}", log_str);
+        return -1;
     }
-
-    LOGT("Read snapshot num_blobs={} end_of_shard={} {}", blob_data_vec.size(), end_of_shard, log_str);
     return 0;
 }
 
 void ReplicationStateMachine::write_snapshot_obj(std::shared_ptr< homestore::snapshot_context > context,
```
Hooper9973, line 319: I have a question regarding the handling of a specific scenario during the baseline resync process. Could you please clarify how the system currently manages the situation where a follower crashes and the leader's `snapshot_sync_ctx` does not time out? Understanding this would be very helpful.

yuwmao: Follower should persist its transmission checkpoint during the baseline resync, so that it can resume from the point where it stops. This logic is not implemented yet.

Hooper9973:
LGTM

xiaoxichen (dismissed):

JacksonYao287 (approved):
LGTM

Hooper9973 (approved):

PR #249, opened 2024-12-30
add traffic check before write

Besroy:
add traffic check according to https://github.com/eBay/HomeStore/pull/610


src/lib/homestore_backend/hs_blob_manager.cpp
```diff
@@ -104,6 +104,11 @@ BlobManager::AsyncResult< blob_id_t > HSHomeObject::_put_blob(ShardInfo const& s
         return folly::makeUnexpected(BlobError(BlobErrorCode::NOT_LEADER, repl_dev->get_leader_id()));
     }
 
+    if (!repl_dev->is_ready_for_traffic()) {
+        LOGW("failed to put blob for pg [{}], shard [{}], not ready for traffic", pg_id, shard.id);
+        return folly::makeUnexpected(BlobError(BlobErrorCode::UNSUPPORTED_OP));
```
xiaoxichen, line 109: better if it can be `TRY_AGAIN`

Besroy: `TRY_AGAIN` seems to be used in somewhere (has compile confilct), so I update to `RETRY_REQUEST`

JacksonYao287 (dismissed):
LGTM

src/lib/homestore_backend/hs_blob_manager.cpp
```diff
@@ -381,6 +386,11 @@ BlobManager::NullAsyncResult HSHomeObject::_del_blob(ShardInfo const& shard, blo
         return folly::makeUnexpected(BlobError(BlobErrorCode::NOT_LEADER, repl_dev->get_leader_id()));
     }
 
+    if (!repl_dev->is_ready_for_traffic()) {
```
yuwmao, line 389: Do we need to add this check also for `_get_blob`? As `RETRY_REQUEST` can be a clear hint for client to retry.

xiaoxichen: I think we need this for get_blob as well

koujl (approved):
lgtm

PR #252, opened 2025-01-15
CP deadlock Bugfix and optimization of pg destory.

Hooper9973:
1. fix cp deadlock, trigger back-to-back cp instead of force cp.
2. optimize pg destroy, call cp before destroying pg superblk to avoid cp first then destroy pg index table.

### DIFF
--- a/src/lib/homestore_backend/hs_pg_manager.cpp
+++ b/src/lib/homestore_backend/hs_pg_manager.cpp
@@ -316,8 +316,15 @@ std::optional< pg_id_t > HSHomeObject::get_pg_id_with_group_id(homestore::group_
 void HSHomeObject::pg_destroy(pg_id_t pg_id) {
     mark_pg_destroyed(pg_id);
     destroy_shards(pg_id);
-    reset_pg_chunks(pg_id);
-    cleanup_pg_resources(pg_id);
+    chunk_selector_->reset_pg_chunks(pg_id);
+    destroy_pg_index_table(pg_id);
+    destroy_pg_superblk(pg_id);
+
+    // return pg chunks to dev heap
+    // which must be done after destroying pg super blk to avoid multiple pg use same chunks
+    bool res = chunk_selector_->return_pg_chunks_to_dev_heap(pg_id);
+    RELEASE_ASSERT(res, "Failed to return pg {} chunks to dev_heap", pg_id);
+
     LOGI("pg {} is destroyed", pg_id);
 }
 void HSHomeObject::mark_pg_destroyed(pg_id_t pg_id) {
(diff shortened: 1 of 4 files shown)


src/lib/homestore_backend/hs_pg_manager.cpp
```diff
@@ -333,25 +340,14 @@ void HSHomeObject::mark_pg_destroyed(pg_id_t pg_id) {
...
-    auto fut = homestore::hs()->cp_mgr().trigger_cp_flush(true /* force */);
-    auto on_complete = [&](auto success) {
-        RELEASE_ASSERT(success, "Failed to trigger CP flush");
-        LOGI("CP Flush completed");
-    };
-    on_complete(std::move(fut).get());
-}
-
-void HSHomeObject::cleanup_pg_resources(pg_id_t pg_id) {
+void HSHomeObject::destroy_pg_index_table(pg_id_t pg_id) {
```
xiaoxichen, line 343: @koujl if the index still has dirty buffer in wbcache, will the `destroy_pg_index_table` skip flushing those buffers in cp_flush?
It is fine if dirty buffer flush cannot be skipped (just the optimization is not working) but it should at least be released from wbcache

xiaoxichen (commented):
lgtm if Jilong approve from IndexService point of view.

src/lib/homestore_backend/hs_pg_manager.cpp
koujl: Yes, destroying an index btree will free all its nodes and bufs, and those bufs marked as freed won't be flushed - they may still be processed by `do_flush_one_buf` for dependency reasons but won't be actually written to the disk.

xiaoxichen: The  `destroy_pg_index_table` function is not calling `destroy_btree`  which as you mentioned mark all bufs freed.  But it just remove it from a map
```
void IndexService::remove_index_table(const std::shared_ptr< IndexTableBase >& tbl) {
    std::unique_lock lg(m_index_map_mtx);
    m_index_map.erase(tbl->uuid());
    m_ordinal_index_map.erase(tbl->ordinal());
}

```
which I dont quite get how the in-memory bufs/nodes will be release.

koujl: No it's the next line
```cpp
    if (nullptr != hs_pg->index_table_) {
        auto uuid_str = boost::uuids::to_string(hs_pg->index_table_->uuid());
        index_table_pg_map_.erase(uuid_str);
        hs()->index_service().remove_index_table(hs_pg->index_table_);
        hs_pg->index_table_->destroy();
    }
```
`IndexTable::destroy()` will call `Btree::destroy_btree`.

xiaoxichen (approved):

PR #254, opened 2025-01-15
Homeobject side graceful shutdown

JacksonYao287:
this PR add counters to record the pending request and start shutdown homestore when no pending request

### DIFF
--- a/src/lib/homestore_backend/hs_homeobject.cpp
+++ b/src/lib/homestore_backend/hs_homeobject.cpp
@@ -60,9 +60,7 @@ public:
         return it->second;
     }
 
-    void on_repl_devs_init_completed() override {
-        _home_object->on_replica_restart();
-    }
+    void on_repl_devs_init_completed() override { _home_object->on_replica_restart(); }
 
     std::pair< std::string, uint16_t > lookup_peer(homestore::replica_id_t uuid) const override {
         std::string endpoint;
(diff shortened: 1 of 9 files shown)


src/lib/homestore_backend/hs_homeobject.cpp
```diff
@@ -298,9 +296,22 @@ HSHomeObject::~HSHomeObject() {
     }
     trigger_timed_events();
 #endif
+
+    start_shutting_down();
+    // Wait for all pending requests to complete
+    while (true) {
+        auto pending_reqs = get_pending_request_num();
```
xiaoxichen, line 303: baseline resync be considered as HS or HO?

JacksonYao287: it will be considered as HS. 

Homeobject side graceful shutdown will only make sure that when shutting down homeobject ,  there is no pending external (or user api) calls from upper layer and reject later api calls .  

baseline resync is a mechanism of statemachin(raft), and all the baseline resync related code in homeobject is called by statemachine. homeobject can not do anything for this.  so , we need to handle this in raft level (homestore).

i will try to submit another pr for homestore graceful shutdown.

JacksonYao287:
https://docs.google.com/document/d/11SP84tI9AwB_FhRSFJoC8oZIHECOn4o2mbooZf4W0EY/edit?tab=t.0#heading=h.gitenqkosfz6

related doc

xiaoxichen (approved):

PR #255, opened 2025-01-24
Implementation of follower-side snapshot transmission resumption

koujl:
* Introduced a new `snp_rcvr_meta_blk` to record snapshot receiving progress on the follower. Since the `ReplicationStateMachine` instance cannot be directly referenced, the superblock is added to the `HS_PG` struct to act as a broker.
* Enabled the follower to request resumption from the beginning of a specific shard in `write_snapshot_obj`. Adjusted the cursor update handling logic to accommodate this scenario.
* Added a new test case for restarting the follower with a shorter delay to test resumption without causing a leader timeout.

### DIFF
--- a/src/lib/homestore_backend/pg_blob_iterator.cpp
+++ b/src/lib/homestore_backend/pg_blob_iterator.cpp
@@ -39,6 +39,23 @@ bool HSHomeObject::PGBlobIterator::update_cursor(objId id) {
     if (id.value == LAST_OBJ_ID) { return true; }
     //resend batch
     if (id.value == cur_obj_id_.value) { return true; }
+
+    // If cur_obj_id_ == 0|0 (PG meta), this may be a request for resuming from specific shard
+    if (cur_obj_id_.shard_seq_num == 0 && id.shard_seq_num != 0 && id.batch_id == 0) {
+        bool found = false;
+        for (size_t i = 0; i < shard_list_.size(); i++) {
+            if (get_sequence_num_from_shard_id(shard_list_[i].info.id) == id.shard_seq_num) {
+                found = true;
+                cur_shard_idx_ = i;
+                cur_start_blob_idx_ = 0;
+                cur_batch_blob_count_ = 0;
+                break;
+            }
+        }
+        if (found) { cur_obj_id_ = id; }
+        return found;
+    }
+
     auto next_obj_id = expected_next_obj_id();
     if (id.value != next_obj_id.value) {
         LOGE("invalid objId, expected={}, actual={}", next_obj_id.to_string(), id.to_string());
--- a/src/lib/homestore_backend/replication_state_machine.cpp
+++ b/src/lib/homestore_backend/replication_state_machine.cpp
@@ -226,6 +226,8 @@ ReplicationStateMachine::create_snapshot(std::shared_ptr< homestore::snapshot_co
 
 bool ReplicationStateMachine::apply_snapshot(std::shared_ptr< homestore::snapshot_context > context) {
     // TODO persist snapshot
+    m_snp_rcv_handler->destroy_context();
+
     std::lock_guard lk(m_snapshot_lock);
     m_snapshot_context = context;
     return true;
@@ -238,7 +240,7 @@ std::shared_ptr< homestore::snapshot_context > ReplicationStateMachine::last_sna
 }
 
 int ReplicationStateMachine::read_snapshot_obj(std::shared_ptr< homestore::snapshot_context > context,
-                                                std::shared_ptr< homestore::snapshot_obj > snp_obj) {
+                                               std::shared_ptr< homestore::snapshot_obj > snp_obj) {
     HSHomeObject::PGBlobIterator* pg_iter = nullptr;
     auto s = dynamic_pointer_cast< homestore::nuraft_snapshot_context >(context)->nuraft_snapshot();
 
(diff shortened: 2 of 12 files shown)


zhiteng:
The PR description should be in the commit message as well to reserve the context the change.

koujl:
Updated commit message.

The codecov report is incorrect due to lagged comparison base. Report from local build shows the overall line coverage of this PR is over 80%.

src/lib/homestore_backend/pg_blob_iterator.cpp
```diff
@@ -39,6 +39,23 @@ bool HSHomeObject::PGBlobIterator::update_cursor(objId id) {
     if (id.value == LAST_OBJ_ID) { return true; }
     //resend batch
     if (id.value == cur_obj_id_.value) { return true; }
+
+    // If cur_obj_id_ == 0|0 (PG meta), this may be a request for resuming from specific shard
+    if (cur_obj_id_.shard_seq_num == 0 && id.shard_seq_num != 0 && id.batch_id == 0) {
```
yuwmao, line 44: Consider the case that the required shard(`id.shard_seq_num`) might be the prior shard on the leader side.
t1: follower receives the last batch of the shard x, flushes the sb and required next shard;
t2: leader send the first batch of shard x+1 and update the cursor to x+1;
t3: follower crashes during handling the snapshot;
t4: follower restarts, and resumes from the beginning of shard x;
t5: leader rejects because `id.shard_seq_num = cur_obj_id_.shard_seq_num -1`.

src/lib/homestore_backend/replication_state_machine.cpp
```diff
@@ -323,16 +323,21 @@ void ReplicationStateMachine::write_snapshot_obj(std::shared_ptr< homestore::sna
...
     auto s = dynamic_pointer_cast< homestore::nuraft_snapshot_context >(context)->nuraft_snapshot();
-    auto obj_id = objId(static_cast< snp_obj_id_t >(snp_obj->offset));
+    auto obj_id = objId(snp_obj->offset);
     auto log_suffix = fmt::format("group={} term={} lsn={} shard={} batch_num={} size={}",
                                   uuids::to_string(r_dev->group_id()), s->get_last_log_term(), s->get_last_log_idx(),
                                   obj_id.shard_seq_num, obj_id.batch_id, snp_obj->blob.size());
 
     if (snp_obj->is_last_obj) {
         LOGD("Write snapshot reached is_last_obj true {}", log_suffix);
+        m_snp_rcv_handler->destroy_context();
```
yuwmao, line 340: Would it be better to move `destroy_context()` into the `apply_snapshot`? If follower crashes after last object yet before compact_log, the lsn will still be 0 after recovery then it has to do the snapshot again.

src/lib/homestore_backend/pg_blob_iterator.cpp
koujl: That's why the `update_snp_info_sb` is called only upon completion of each shard, persisting the cursor as the **next shard id**. In this case the follower will request for shard `x+1` batch 0 after a restart.

src/lib/homestore_backend/replication_state_machine.cpp
koujl: Good point, I put it here since `apply_snapshot` might not be called as we mentioned before, but leaking an invalid superblk seems better than a complete redo. I'll move it into `apply_snapshot`.

src/lib/homestore_backend/snapshot_receive_handler.cpp
```diff
@@ -256,17 +256,55 @@ int HSHomeObject::SnapshotReceiveHandler::process_blobs_snapshot_data(ResyncBlob
...
 int64_t HSHomeObject::SnapshotReceiveHandler::get_context_lsn() const { return ctx_ ? ctx_->snp_lsn : -1; }
+pg_id_t HSHomeObject::SnapshotReceiveHandler::get_context_pg_id() const { return ctx_ ? ctx_->pg_id : 0; }
+
+bool HSHomeObject::SnapshotReceiveHandler::load_prev_context() {
+    HS_PG* hs_pg = nullptr;
+    {
+        std::shared_lock lck(home_obj_._pg_lock);
+        auto iter = find_if(home_obj_._pg_map.begin(), home_obj_._pg_map.end(), [this](auto& pg) {
+            auto hs_pg = dynamic_cast< HS_PG* >(pg.second.get());
+            return hs_pg->repl_dev_ == repl_dev_;
```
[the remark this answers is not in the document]
koujl: I cannot come up with any case... since both HS_PG and ReplicationStateMachine are bound to a `homestore::ReplDev` and don't change once constructed.

src/lib/homestore_backend/replication_state_machine.cpp
yuwmao: Yes, is there any change that we can clear the leaked superblk?

yuwmao (commented):
LGTM

src/lib/homestore_backend/replication_state_machine.cpp
koujl: Currently no - but even if there're any outdated superblocks they'll be destroyed by reset_context() at the beginning of the next baseline resync due to LSN mismatch. So I don't think the leak will affect our resync process.

xiaoxichen (dismissed):

koujl:
Found that unit test may crash at [here](https://github.com/eBay/HomeObject/blob/ae448558f8a4c3276d877e7e569e19232f195a9a/src/lib/homestore_backend/snapshot_receive_handler.cpp#L178-L180) due to corrupted header bytes.

It seems we lack protection for the integrity of BlobHeader itself?

yuwmao (approved):
LGTM

PR #256, opened 2025-02-12
purge repl dev when destroy pg

Besroy:
corresponding to https://github.com/eBay/HomeStore/pull/643


xiaoxichen (dismissed):

xiaoxichen (commented):
nit: if it require HS >= certain version , better update 
```
self.requires("homestore/[^6.6]@oss/master")
```

PR #258, opened 2025-02-18
Commit Blocks in Snapshot and Enhance UT

yuwmao:
Commit Blocks in snapshot, or the blocks will be overwritten after restart.
Add more UTs to test logic when follower/leader restart.

### DIFF
--- a/conanfile.py
+++ b/conanfile.py
@@ -9,7 +9,7 @@ required_conan_version = ">=1.60.0"
 
 class HomeObjectConan(ConanFile):
     name = "homeobject"
-    version = "2.2.5"
+    version = "2.2.6"
 
     homepage = "https://github.com/eBay/HomeObject"
     description = "Blob Store built on HomeReplication"
--- a/src/lib/homestore_backend/CMakeLists.txt
+++ b/src/lib/homestore_backend/CMakeLists.txt
@@ -69,13 +69,29 @@ target_link_libraries(homestore_test_dynamic PUBLIC
     ${COMMON_TEST_DEPS}
 )
 
-add_test(NAME HomestoreTestDynamic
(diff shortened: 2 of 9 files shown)


xiaoxichen (approved):
lgtm, a bit concern on the leader we assigned during creation might not be the leader when we doing the test (if leader switch happened in between).  But as far as this is testing code,  it 's LGTM

yuwmao:
> lgtm, a bit concern on the leader we assigned during creation might not be the leader when we doing the test (if leader switch happened in between). But as far as this is testing code, it 's LGTM

Yes, it could occur, will improve it by using get_leader_id instead of constant initial_leader_replica_id as far as possible to reduce the chances of this case occurring.

xiaoxichen:
> > lgtm, a bit concern on the leader we assigned during creation might not be the leader when we doing the test (if leader switch happened in between). But as far as this is testing code, it 's LGTM
> 
> Yes, it could occur, will improve it by using get_leader_id instead of constant initial_leader_replica_id as far as possible to reduce the chances of this case occurring.

But if the leader is replica-0 it is hard to handle :).  Maybe explicity request leadership is better . Anyway lets improve later, feel free to merge as is

yuwmao:
> > > lgtm, a bit concern on the leader we assigned during creation might not be the leader when we doing the test (if leader switch happened in between). But as far as this is testing code, it 's LGTM
> > 
> > 
> > Yes, it could occur, will improve it by using get_leader_id instead of constant initial_leader_replica_id as far as possible to reduce the chances of this case occurring.
> 
> But if the leader is replica-0 it is hard to handle :). Maybe explicity request leadership is better . Anyway lets improve later, feel free to merge as is

Ok, if it supports to force change leadership to the specified member, that will be easy, we can request and wait for leader.

PR #259, opened 2025-02-21
Implement local snapshot persistency

koujl:
Persist last local snapshot and reload it on restart to avoid getting stuck at last_snapshot() in some corner cases where NuRaft tries to append logs without any existing log.


yuwmao (commented):
LGTM

xiaoxichen (approved):

PR #261, opened 2025-02-25
fix pg members name copy logic

Besroy:


src/lib/homestore_backend/hs_pg_manager.cpp
```diff
@@ -517,7 +519,9 @@ HSHomeObject::HS_PG::HS_PG(PGInfo info, shared< homestore::ReplDev > rdev, share
     pg_members* pg_sb_members = pg_sb_->get_pg_members_mutable();
     for (auto const& m : pg_info_.members) {
         pg_sb_members[i].id = m.id;
```
xiaoxichen, line 521: better we can have debug_assert if overflow.

xiaoxichen (approved):
lgtm except a nit

src/lib/homestore_backend/hs_pg_manager.cpp
Besroy: Do we have a restriction like sm_name_len < 32 when creating SM? I am considering whether the assert might cause unexpected failures. If there is no such restriction, how about changing it to LOGW?

xiaoxichen (approved):

PR #264, opened 2025-02-27
Fix bugs in HS dynamic UT framework

koujl:
1. Introduce intervals between spawning processes to prevent conflicts when accessing config files.
This may cause the following failure
```
unknown file: Failure
C++ exception with description "[json.exception.parse_error.101] parse error at line 1, column 1: syntax error while parsing value - unexpected end of input; expected '[', '{', or a literal" thrown in SetUp().
```
2. Bump sync point num by 1000 at the start of each test case to prevent interference.
3. Remove existing `gtest_filter` argument when spawning restarted processes to ensure only the remaining test cases are executed.

------

There are still some random failures witness during local runs which seem to be caused by replica 2 getting stuck at `pg_destroy()` - so I also added some additional changes & logs to make it easier to identify.

### DIFF
--- a/conanfile.py
+++ b/conanfile.py
@@ -9,7 +9,7 @@ required_conan_version = ">=1.60.0"
 
 class HomeObjectConan(ConanFile):
     name = "homeobject"
-    version = "2.2.10"
+    version = "2.2.11"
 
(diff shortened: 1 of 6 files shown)


JacksonYao287 (approved):
LGTM

PR #265, opened 2025-03-04
Add BlobHeader integrity validation

koujl:
Implement CRC32 hash integrity check for BlobHeader to prevent crashes caused by faults in critical fields (size/algorithm...).

### DIFF
--- a/src/lib/homestore_backend/hs_homeobject.hpp
+++ b/src/lib/homestore_backend/hs_homeobject.hpp
@@ -157,7 +157,7 @@ public:
         data_type_t type{data_type_t::BLOB_INFO};
     };
 
-    struct shard_info_superblk : public DataHeader {
+    struct shard_info_superblk : DataHeader {
         ShardInfo info;
         homestore::chunk_num_t p_chunk_id;
         homestore::chunk_num_t v_chunk_id;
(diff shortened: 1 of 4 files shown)


src/lib/homestore_backend/hs_homeobject.hpp
```diff
@@ -325,6 +326,37 @@ class HSHomeObject : public HomeObjectImpl {
...
                                spdlog::to_hex(hash, hash + blob_max_hash_len));
         }
+
+        bool valid() const {
+            if (!DataHeader::valid()) { return false; }
+
+            switch (hash_algorithm) {
+            case HashAlgorithm::NONE:
+                return true;
+            case HashAlgorithm::CRC32: {
```
JacksonYao287, line 336: we should use the same hash_algorithm between `valid` and `seal` , right?
`seal` is now always using crc32,  so other hash_algorithm does not make sense to `valid`.

I mean can we add a member to indicate the hash_algorithm, and `valid` and `seal` will share it

yuwmao (dismissed):
LGTM

src/lib/homestore_backend/hs_homeobject.hpp
koujl: Good catch, will fix in the next push.

koujl: https://github.com/eBay/HomeObject/compare/43b9bdeec20fb53b9e8ba3fe5833e3dd23cc1c1c..942c9f40d43c9824192564c47db9108cf06f77cf

JacksonYao287 (approved):

PR #266, opened 2025-03-05
Add metrics for baseline resync

yuwmao:
Add statistic information to the pg meta/shard meta message.
Add metrics to track the progress.
- DonerSnapshotMetrics is in memory;
- ReceiverSnapshotMetrics persists the progress. 

Currently, use `DefaultBucket` for histogram metrics. We can refine it after the BR performance test.

### DIFF
--- a/src/lib/homestore_backend/hs_homeobject.hpp
+++ b/src/lib/homestore_backend/hs_homeobject.hpp
@@ -170,11 +170,54 @@ public:
         char data[1];
     };
 
+    struct durable_snapshot_progress {
+        uint64_t start_time{0};
+        uint64_t total_blobs{0};
+        uint64_t total_bytes{0};
+        uint64_t total_shards{0};
+        uint64_t complete_blobs{0};
+        uint64_t complete_bytes{0};
+        uint64_t complete_shards{0};
+        uint64_t corrupted_blobs{0};
+    };
+
+    struct snapshot_progress {
+        uint64_t start_time{0};
+        uint64_t total_blobs{0};
+        uint64_t total_bytes{0};
+        uint64_t total_shards{0};
+        uint64_t complete_blobs{0};
+        uint64_t complete_bytes{0};
+        uint64_t complete_shards{0};
+        // The count of the blobs which have been corrupted on the leader side.
+        uint64_t corrupted_blobs{0};
+        // Record the stats of the current batch to avoid double counting.
+        uint64_t cur_shard_total_blobs{0};
+        uint64_t cur_shard_complete_blobs{0};
+        // Used to handle the retried batch message.
+        uint64_t cur_batch_blobs{0};
+        uint64_t cur_batch_bytes{0};
+        uint64_t error_count{0};
+
+        snapshot_progress() = default;
+        explicit snapshot_progress(durable_snapshot_progress p) {
+            start_time = p.start_time;
+            total_blobs = p.total_blobs;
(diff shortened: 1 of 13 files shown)


src/lib/homestore_backend/hs_homeobject.hpp
```diff
@@ -400,6 +448,25 @@ class HSHomeObject : public HomeObjectImpl {
...
         bool end_of_scan() const;
 
+        // All of the leader's metrics are in-memory
+        struct DonerSnapshotMetrics : sisl::MetricsGroup {
+            explicit DonerSnapshotMetrics(pg_id_t pg_id) : sisl::MetricsGroup("snapshot_doner", std::to_string(pg_id)) {
+                REGISTER_COUNTER(snp_dnr_load_blob, "Loaded blobs in baseline resync");
+                REGISTER_COUNTER(snp_dnr_load_bytes, "Loaded bytes in baseline resync");
+                REGISTER_COUNTER(snp_dnr_resend_count, "Mesg resend times in baseline resync");
+                REGISTER_COUNTER(snp_dnr_error_count, "Error times when reading blobs in baseline resync");
+                REGISTER_HISTOGRAM(snp_dnr_blob_process_time, "Time cost of successfully process a blob in baseline resync",
```
xiaoxichen, line 458: Suggesting use batch_process_time as batch size is less variance compared to blob size.

xiaoxichen: Also can we have a latnecy between two `update_cursor` call? It is more of a end-to-end latency of a batch which can include encoding/decoding and network.

src/lib/homestore_backend/hs_homeobject.hpp
```diff
@@ -466,14 +534,66 @@ class HSHomeObject : public HomeObjectImpl {
...
+                REGISTER_GAUGE(snp_rcvr_total_blob, "Total blobs in baseline resync");
+                REGISTER_GAUGE(snp_rcvr_total_bytes, "Total bytes in baseline resync")
+                REGISTER_GAUGE(snp_rcvr_total_shards, "Total shards in baseline resync")
+                REGISTER_GAUGE(snp_rcvr_complete_blob, "Complete blob in baseline resync")
+                REGISTER_GAUGE(snp_rcvr_complete_bytes, "Complete bytes in baseline resync")
+                REGISTER_GAUGE(snp_rcvr_complete_shards, "Complete shards in baseline resync")
+                REGISTER_GAUGE(snp_rcvr_current_shard_total_blobs,
+                               "Total blob of the current shard in baseline resync")
+                REGISTER_GAUGE(snp_rcvr_current_shard_complete_blobs,
+                               "Compelete blob of the current blob in baseline resync");
```
xiaoxichen, line 554: how these two metrics helps?

xiaoxichen (commented):
a few suggestion for metrics.


Q: What is the reason to introduce metrics in testing code ?

src/lib/homestore_backend/hs_homeobject.hpp
yuwmao: Removed these two metrics.

src/lib/homestore_backend/hs_homeobject.hpp
yuwmao: Add batch_e2e_time and batch_process_time. 
Also keep blob_process_time in the performance test. We can remove it later if not needed together with time bucket improvement.

PR #267, opened 2025-03-10
Ensure CP flush is triggered during `pg_destroy`

Besroy:
HS uses the `force` parameter to determine the action when a previous checkpoint is being flushed:
- If `force` is set to `false`, HS will directly return `false`.
- If `force` is set to `true`, HS will add it to the back-to-back queue and return a future. This is required for `pg_destroy`.

### DIFF
--- a/conanfile.py
+++ b/conanfile.py
@@ -9,7 +9,7 @@ required_conan_version = ">=1.60.0"
 
 class HomeObjectConan(ConanFile):
     name = "homeobject"
-    version = "2.2.13"
+    version = "2.2.14"
 
     homepage = "https://github.com/eBay/HomeObject"
     description = "Blob Store built on HomeReplication"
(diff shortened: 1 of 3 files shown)


xiaoxichen:
How we solve the deadlock issue previously causing by fiber thread??

JacksonYao287 (dismissed):
LG

Besroy:
> How we solve the deadlock issue previously causing by fiber thread??

That is solved by  https://github.com/eBay/HomeStore/pull/630. Setting force to false did not actually solve the deadlock issue, as it was based on a misunderstanding of HS back-to-back operations.

xiaoxichen:
lgtm 
cc @yamingk

xiaoxichen (dismissed):

JacksonYao287 (dismissed):
LGTM

yuwmao (approved):

yamingk:
LGTM

PR #270, opened 2025-03-13
Fix UT

yuwmao:
- Add ut for graceful shutdown.
- In RestartLeader UT, the actual leader may not be the assigned leader(with highest priority), thus both replica0 and replica1 could be the leader. Let out member process takes the responsibility to spawn process to simulate restart.


xiaoxichen (approved):

JacksonYao287 (approved):
LGTM

PR #271, opened 2025-03-18
Implement fetch_data callback

JacksonYao287:
in homeobject, we have three kinds of requests that will write data(thus fetch_data might happen) to a chunk:
1 create_shard : will write a shard header to a chunk
2 seal_shard : will write a shard footer to a chunk
3 put_blob: will write user data to a chunk

 for any type that writes data to a chunk, we need to handle the fetch_data request for it.

### DIFF
--- a/src/lib/homestore_backend/hs_homeobject.hpp
+++ b/src/lib/homestore_backend/hs_homeobject.hpp
@@ -453,12 +453,15 @@ public:
                 REGISTER_COUNTER(snp_dnr_load_bytes, "Loaded bytes in baseline resync");
                 REGISTER_COUNTER(snp_dnr_resend_count, "Mesg resend times in baseline resync");
                 REGISTER_COUNTER(snp_dnr_error_count, "Error times when reading blobs in baseline resync");
-                REGISTER_HISTOGRAM(snp_dnr_blob_process_time, "Time cost(us) of successfully process a blob in baseline resync",
-                   HistogramBucketsType(DefaultBuckets));
+                REGISTER_HISTOGRAM(snp_dnr_blob_process_time,
+                                   "Time cost(us) of successfully process a blob in baseline resync",
+                                   HistogramBucketsType(DefaultBuckets));
                 REGISTER_HISTOGRAM(snp_dnr_batch_process_time,
-                                   "Time cost(ms) of successfully process a batch in baseline resync", HistogramBucketsType(DefaultBuckets));
+                                   "Time cost(ms) of successfully process a batch in baseline resync",
+                                   HistogramBucketsType(DefaultBuckets));
                 REGISTER_HISTOGRAM(snp_dnr_batch_e2e_time,
-                                   "Time cost(ms) of a batch end-to-end round trip in baseline resync", HistogramBucketsType(DefaultBuckets));
+                                   "Time cost(ms) of a batch end-to-end round trip in baseline resync",
+                                   HistogramBucketsType(DefaultBuckets));
                 register_me_to_farm();
             }
 
@@ -490,7 +493,7 @@ public:
         pg_id_t pg_id_;
         shared< homestore::ReplDev > repl_dev_;
         uint64_t max_batch_size_;
-        std::unique_ptr<DonerSnapshotMetrics> metrics_;
+        std::unique_ptr< DonerSnapshotMetrics > metrics_;
     };
 
     class SnapshotReceiveHandler {
@@ -542,9 +545,9 @@ public:
             SnapshotContext(int64_t lsn, pg_id_t pg_id) : snp_lsn{lsn}, pg_id{pg_id} {}
         };
 
-        struct ReceiverSnapshotMetrics: sisl::MetricsGroup {
-            ReceiverSnapshotMetrics(std::shared_ptr<SnapshotContext> ctx) : sisl::MetricsGroup("snapshot_receiver", std::to_string(ctx->pg_id)),
-                    ctx_{ctx} {
+        struct ReceiverSnapshotMetrics : sisl::MetricsGroup {
+            ReceiverSnapshotMetrics(std::shared_ptr< SnapshotContext > ctx) :
+                    sisl::MetricsGroup("snapshot_receiver", std::to_string(ctx->pg_id)), ctx_{ctx} {
                 REGISTER_GAUGE(snp_rcvr_total_blob, "Total blobs in baseline resync");
                 REGISTER_GAUGE(snp_rcvr_total_bytes, "Total bytes in baseline resync")
                 REGISTER_GAUGE(snp_rcvr_total_shards, "Total shards in baseline resync")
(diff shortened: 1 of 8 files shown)


src/lib/homestore_backend/hs_homeobject.hpp
```diff
@@ -453,12 +453,15 @@ class HSHomeObject : public HomeObjectImpl {
                 REGISTER_COUNTER(snp_dnr_load_bytes, "Loaded bytes in baseline resync");
                 REGISTER_COUNTER(snp_dnr_resend_count, "Mesg resend times in baseline resync");
                 REGISTER_COUNTER(snp_dnr_error_count, "Error times when reading blobs in baseline resync");
-                REGISTER_HISTOGRAM(snp_dnr_blob_process_time, "Time cost(us) of successfully process a blob in baseline resync",
-                   HistogramBucketsType(DefaultBuckets));
+                REGISTER_HISTOGRAM(snp_dnr_blob_process_time,
```
yamingk, line 456: I know it is show here for formatting change, can we rename it to _latency, instead of _time?

src/lib/homestore_backend/hs_homeobject.hpp
```diff
@@ -490,7 +493,7 @@ class HSHomeObject : public HomeObjectImpl {
         pg_id_t pg_id_;
         shared< homestore::ReplDev > repl_dev_;
         uint64_t max_batch_size_;
-        std::unique_ptr<DonerSnapshotMetrics> metrics_;
+        std::unique_ptr< DonerSnapshotMetrics > metrics_;
```
yamingk, line 496: nit: Looks like the previous originator of these code doesn't have clang-format installed? Is it possible to kindly inform that we all have clang-format applied and make sure they are formatted on save?

src/lib/homestore_backend/replication_state_machine.cpp
```diff
@@ -478,6 +480,163 @@ void ReplicationStateMachine::set_snapshot_context(std::shared_ptr< homestore::s
...
+
+        std::memcpy(given_buffer, sb, total_size);
+        return folly::makeFuture< std::error_code >(std::error_code{});
+    }
+
+        // TODO: for shard header and footer, follower can generate it itself according to header, no need to fetch it
+        // from leader. this can been done by adding another callback, which will be called before follower tries to
+        // fetch data.
+
+    case ReplicationMessageType::PUT_BLOB_MSG: {
```
yamingk, line 534: I probably didn't quite get fetch blob would have message type put_blob_msg? Is there a reason behind it?

src/lib/homestore_backend/hs_homeobject.hpp
JacksonYao287: ack

src/lib/homestore_backend/hs_homeobject.hpp
JacksonYao287: will do

src/lib/homestore_backend/replication_state_machine.cpp
```diff
@@ -478,6 +480,163 @@ void ReplicationStateMachine::set_snapshot_context(std::shared_ptr< homestore::s
     m_snapshot_context = context;
 }
 
+folly::Future< std::error_code > ReplicationStateMachine::on_fetch_data(const int64_t lsn, const sisl::blob& header,
```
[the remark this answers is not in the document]
JacksonYao287: in homeobject, we have three kind of request that will write user data:
1 create shard : will write a shard header to a chunk
2 seal shard : will write a shard footer to a chunk
3 put_blob: will write user data to a chunk

so fetch_data might happens to these three kinds of message.

will try to add some detailed description

src/lib/homestore_backend/replication_state_machine.cpp
JacksonYao287: user header is included in the fetch_data request https://github.com/eBay/HomeStore/blob/36600b1ae949347f0ce652ee7f552c07e298bedd/src/lib/replication/fetch_data_rpc.fbs#L7
so , at homeobject level, we can identify the message type from the user header.
actually , the message type here means which kind of  raft op we are trying to fetch data for.  PUT_BLOB_MSG means we received a put_blob in leader, and the corresponding log has been sent to and received by follower, but follower does not receive the data, so this follower wants to get the data of this PUT_BLOB_MSG message( which is also the type of the corresponding log ).

src/lib/homestore_backend/pg_blob_iterator.cpp
JacksonYao287, line 1: only format change, no log change

xiaoxichen (commented):
I would suggest to wrap the `Read from Index` into a future.

Then the `fetch_data` could be cleaner like

```
#ifdef _PRERELEASE
 if (iomgr_flip::instance()->test_flip("force_to_read_by_index_table")) {
    return read_from_index ();
}
#endif
datasvc.async_read().thenValue( []() { 
  if (!validate_blob()) { return read_from_index());

}

It is OK in the `thenValue` that return another future.

**Comment by xiaoxichen** [comment]
sample code
```
class Test {
public:
  folly::Promise<int> promise_;
  void fulfill_with_delay() {
    sleep(3);
    setPromise(3);
  }
  void setPromise(int value) {
    std::cout << "Promise fulfilled at " << time(NULL) << " value: " << value << std::endl;
    promise_.setValue(value);
  }
  folly::Future<int> put_delay() {
    auto initialFuture = folly::makeFuture(10).via(folly::getGlobalCPUExecutor());

    // Add a continuation that returns another future
    auto chainedFuture =
        std::move(initialFuture)
            .thenValue([this](int value) {
              std::cout << "First continuation, time: " << time(NULL) << " value: " << value << std::endl;
              std::cout << "returning another future" << std::endl;
              return this->promise_.getFuture();
            })
            .thenValue([](int newValue) {
              std::cout << "Second continuation, time: " << time(NULL) << " new value: " << newValue
                        << std::endl;
              // Perform another operation or return another future if needed
              return newValue + 5;
            });
    return chainedFuture;
  }
};

int main() {

  Test foo;
  auto start_time = time(NULL);
  auto fut = foo.put_delay();
  foo.fulfill_with_delay();
  auto res = std::move(fut).get();
  std::cout << "res: " << res << std::endl;
  return 0;
}
```

Output
```
First continuation, time: 1742874121 value: 10
returning another future
Promise fulfilled at 1742874124 value: 3
Second continuation, time: 1742874124 new value: 3
res: 8
```

**Comment by JacksonYao287** [comment]
>I would suggest to wrap the Read from Index into a future.

I got your point here. well , IMO, test_flip("force_to_read_by_index_table") is used to Simulate the blob invalidation of the first async_read,  not bypassing async_read.

**Comment by xiaoxichen** [comment]
> > I would suggest to wrap the Read from Index into a future.
> 
> I got your point here. well , IMO, test_flip("force_to_read_by_index_table") is used to Simulate the blob invalidation of the first async_read, not bypassing async_read.

then change it to `simulate_data_moved` or `local_blk_data_invalid` etc.

I think it is also regarding the mixture of sync and async behavior.   If read_from_index itself is a future then we can chain it gracefully without `.get()`.  I think you aware this and move the thenValue to GlobalIOThreadPool.  I cannot say it is wrong , just providing a better implementation idea.
```
/// When practical, prefer SemiFuture/Future's nonblocking style/pattern:
///
/// - the nonblocking style uses continuations, e.g., `.thenValue`, etc.; the
///   continuations are deferred until the result is available.
/// - the blocking style blocks until complete, e.g., `.wait()`, `.get()`, etc.
/// - the two styles cannot be mixed within the same future; use one or the
///   other.

```

JacksonYao287:
@xiaoxichen thanks for the reminder , that makes sense. I have update the PR to avoid a blocking call inside an unblocking call. PTAL

xiaoxichen (approved):
lgtm

yamingk:
LGTM

PR #272, opened 2025-03-25
Handle empty shard in baseline resync

yuwmao:
Previously, when a shard is empty, follower will ask for data batch 1 after handling the metadata, however the leader expect_next_obj_id will return objId for next shard, which leads to inconsistency objId.


koujl (approved):
lgtm

xiaoxichen (approved):

PR #279, opened 2025-04-03
Reserve space in a chunk

yuwmao:
There is a side effect introduced by identical layout: After several restarts or leader switch, some garbage would cause NO_SPACE_ERROR on the follower side which leads to PG NotClean(lag and degraded) ultimately. So reserve some space in a chunk to mitigate/avoid this issue.

### DIFF
--- a/src/lib/homestore_backend/hs_blob_manager.cpp
+++ b/src/lib/homestore_backend/hs_blob_manager.cpp
@@ -1,3 +1,4 @@
+#include "hs_backend_config.hpp"
 #include "hs_homeobject.hpp"
 #include "replication_message.hpp"
 #include "replication_state_machine.hpp"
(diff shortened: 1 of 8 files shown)


src/lib/homestore_backend/hs_blob_manager.cpp
```diff
@@ -410,7 +411,7 @@ HSHomeObject::blob_put_get_blk_alloc_hints(sisl::blob const& header, cintrusive<
     auto hs_shard = d_cast< HS_Shard* >((*shard_iter->second).get());
     BLOGD(tid, msg_header->shard_id, msg_header->blob_id, "Picked p_chunk_id=[{}]", hs_shard->sb_->p_chunk_id);
```
xiaoxichen, line 412: move this down to include the reserved_blks
also L419 if we can refine the logging .

src/lib/homestore_backend/replication_state_machine.cpp
```diff
@@ -172,6 +173,7 @@ ReplicationStateMachine::get_blk_alloc_hints(sisl::blob const& header, uint32_t
         static_assert(std::is_same< homestore::chunk_num_t, uint16_t >::value, "chunk_num_t is not uint16_t");
         homestore::chunk_num_t v_chunk_id = v_chunkID.value();
         hints.application_hint = ((uint64_t)pg_id << 16) | v_chunk_id;
+        if (hs_ctx->is_proposer()) { hints.reserved_blks = hs_pg->get_reserved_blks(); }
```
xiaoxichen, line 176: can we use this feature to have minimal  shard size?

src/lib/homestore_backend/hs_pg_manager.cpp
```diff
@@ -538,6 +541,11 @@ HSHomeObject::HS_PG::HS_PG(PGInfo info, shared< homestore::ReplDev > rdev, share
     pg_sb_->active_blob_count = 0;
     pg_sb_->tombstone_blob_count = 0;
     pg_sb_->total_occupied_blk_count = 0;
+    if (HS_BACKEND_DYNAMIC_CONFIG(reserved_bytes_in_chunk) > 0) {
+        reserved_blks_ =
```
xiaoxichen, line 545: nit:dup code

src/lib/homestore_backend/hs_blob_manager.cpp
yuwmao: Sure.

src/lib/homestore_backend/hs_pg_manager.cpp
yuwmao: Will change it a global variable.

src/lib/homestore_backend/replication_state_machine.cpp
yuwmao: Let's postpone this discussion when we have shard size management.

xiaoxichen (dismissed):

xiaoxichen:
lgtm

xiaoxichen (approved):

PR #280, opened 2025-04-07
fix destroy pg deadlock

JacksonYao287:
for now, we first acquire the pg_lock , and then do index_table_->destroy() , which will trigger a cp flush
https://github.com/eBay/HomeStore/blob/3c10d48f562483a600bad9e95fa02d87fe7dc072/src/include/homestore/index/index_table.hpp#L104

in cp_flush , homeobject#cp_flush will be called , and will try to acquire pg_lock again 
https://github.com/eBay/HomeObject/blob/0c16dcc73b1b1e7fe287430cba8b5518309b4dd2/src/lib/homestore_backend/hs_cp_callbacks.cpp#L36-L37

then a dead_lock will happen and cause a crash.

this PR try to fix this issue


JacksonYao287:
this is one crash case of issue 36

Besroy (approved):
Thanks for identifying and resolving one of the crash issue!

xiaoxichen (approved):

PR #284, opened 2025-04-10
fix fetch data bug

JacksonYao287:
fix a memcopy issuse, we need copy the exact size of shard_super_blk，not the whole blk size

### DIFF
--- a/src/lib/homestore_backend/replication_state_machine.cpp
+++ b/src/lib/homestore_backend/replication_state_machine.cpp
@@ -149,7 +149,8 @@ void ReplicationStateMachine::on_error(ReplServiceError error, const sisl::blob&
 }
 
 homestore::ReplResult< homestore::blk_alloc_hints >
-ReplicationStateMachine::get_blk_alloc_hints(sisl::blob const& header, uint32_t data_size, cintrusive< homestore::repl_req_ctx >& hs_ctx) {
+ReplicationStateMachine::get_blk_alloc_hints(sisl::blob const& header, uint32_t data_size,
+                                             cintrusive< homestore::repl_req_ctx >& hs_ctx) {
     const ReplicationMessageHeader* msg_header = r_cast< const ReplicationMessageHeader* >(header.cbytes());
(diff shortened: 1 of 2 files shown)


src/lib/homestore_backend/heap_chunk_selector.cpp
```diff
@@ -421,4 +420,35 @@ uint64_t HeapChunkSelector::total_blks(uint32_t dev_id) const {
...
+        LOGWARNMOD(homeobject, "No pdev heap found for pdev {}", pdev_id);
+        return nullptr;
+    }
+    auto pdev_heap = it->second;
+    std::unique_lock lock(pdev_heap->mtx);
+    if (pdev_heap->m_heap.empty()) {
+        LOGWARNMOD(homeobject, "No available chunk found for pdev {}", pdev_id);
+        return nullptr;
+    }
+    auto chunk = pdev_heap->m_heap.top();
```
Hooper9973, line 444: we can check chunk state
```
// sanity check
        RELEASE_ASSERT(chunk->get_total_blks() == chunk->available_blks(), "chunk should be empty");
        RELEASE_ASSERT(chunk->available(), "chunk state should be available");
```

JacksonYao287: ack

src/lib/homestore_backend/replication_state_machine.cpp
```diff
@@ -557,7 +557,7 @@ folly::Future< std::error_code > ReplicationStateMachine::on_fetch_data(const in
 
         // TODO：：return error_code if assert fails, so it will not crash here because of the assert failure.
 
-        std::memcpy(given_buffer, sb, total_size);
+        std::memcpy(given_buffer, sb, raw_size);
```
Besroy, line 560: I have 2 small questions:
1. Do we need to initialize buffer to prevent interference from existing data.
2. Do we need to add handle for del_blob_msg

xiaoxichen (approved):

PR #285, opened 2025-04-17
implement gc thread model

JacksonYao287:


src/lib/homestore_backend/gc_manager.cpp
```diff
@@ -0,0 +1,389 @@
...
+                // TODO:: use the future to check if the normal task is completed when necessary
+                actor->add_gc_task(
+                    TASK_PRIORITY::NORMAL,
+                    {chunk->get_chunk_id(), static_cast< uint8_t >(TASK_PRIORITY::NORMAL), std ::move(promise)});
+            }
+        }
+    }
+}
+
+/* pdev_gc_actor */
```
xiaoxichen, line 103: s/pdev_gc_reactor/GCReactor?

src/lib/homestore_backend/gc_manager.cpp
```diff
@@ -0,0 +1,389 @@
...
+
+void GCManager::pdev_gc_actor::start() {
+    bool stopped = true;
+    if (!m_is_stopped.compare_exchange_strong(stopped, false, std::memory_order_release, std::memory_order_relaxed)) {
+        LOGERROR("pdev gc actor for pdev_id={} is already started, no need to start again!", m_pdev_id);
+        return;
+    }
+    // thread number is the same as reserved chunk, which can make sure every gc thread can take a reserved chunk
+    // for gc
+    m_gc_executor = std::make_shared< folly::IOThreadPoolExecutor >(RESERVED_CHUNK_NUM_PER_PDEV);
```
xiaoxichen, line 126: this implementation is wired,  as it is using `folly::IOThreadPoolExecutor` there is no point to keep a busy loop.
We can simply active an executor in both
1.  `scan_chunks_for_gc` where the topK  normal GC candidates are picked.
2.   in the emergent GC handler also add a task into the executor.
by doing 
```
  m_gc_executor->add([this] {
                GCManager::gc_task task;
                if (m_gc_task_queue.try_dequeue(task)) {
                    process_gc_task(task);
                } 
}
```

xiaoxichen: Even more, the task queue can be removed as its only functionality is multiplexing emergent_GC(only one) and normal_gc(K). 

Assuming RESERVED_CHUNK_NUM_PER_PDEV = K,  we can get K-1 chunks for normal GC and always reserve one chunk for emergent GC.

xiaoxichen (commented):
The data copy logic is not fully done so just comment a bit based on what we have.
Data copy logic provides good details which is helpful for understanding the thread model however when merging lets try to separate them

src/lib/homestore_backend/gc_manager.cpp
```diff
@@ -0,0 +1,453 @@
...
+        LOGINFO("gc schduler timer is not running, no need to stop it");
+        return;
+    }
+
+    LOGINFO("stop gc schduler timer");
+    iomanager.cancel_timer(m_gc_timer_hdl, true);
+    m_gc_timer_hdl = iomgr::null_timer_handle;
+
+    for (const auto& [pdev_id, gc_atcor] : m_pdev_gc_actors) {
+        gc_atcor->stop();
```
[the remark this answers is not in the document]
JacksonYao287: ack

src/lib/homestore_backend/gc_manager.cpp
```diff
@@ -0,0 +1,453 @@
...
+    auto defrag_blk_num = chunk->get_defrag_nblks();
+    auto total_blk_num = chunk->get_total_blks();
+
+    // defrag_blk_num > (GC_THRESHOLD_PERCENT/100) * total_blk_num, to avoid floating point number calculation
+    // TODO: avoid overflow here.
+    return 100 * defrag_blk_num > total_blk_num * GC_GARBAGE_RATE_THRESHOLD;
+}
+
+void GCManager::scan_chunks_for_gc() {
+    for (const auto& [_, chunk] : m_chunk_selector->get_all_chunks()) {
```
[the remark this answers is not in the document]
JacksonYao287: if this, we need to maintain a heap to identify which chunk has more available space, which will involve more concurrent logic to manage this heap.

actually, this topic was discussed earlier.  if we set the GC_GARBAGE_RATE_THRESHOLD to a high number , like 60% or above, it is not that important to always gc the chunk with the highest garbage rate. 
pls refer to the last paragraph of [this chapter](https://docs.google.com/document/d/1zW82aMCf3obxS4bccnyG2YD6enzt-R5C10-rDwx6gP0/edit?tab=t.0#heading=h.86qhu2g9zqjm) 

for example , we have two chunks with the garbage rate of 75% and 80%,  respectively. it is not that important we make sure the chunk with higher garbage rate will be gc before the one with a lower garbage rate.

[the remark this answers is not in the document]
xiaoxichen: we dont need a heap but just iterate all chunks and get top K , which is a 
O(N)Log(K) (scan) + O(NlogN) (pop).

Comparing to the implementation of now it is a O(N) (scan) + K*O(NlgN) (pop)

src/lib/homestore_backend/heap_chunk_selector.cpp
```diff
@@ -421,4 +420,87 @@ uint64_t HeapChunkSelector::total_blks(uint32_t dev_id) const {
...
 
+std::list< uint32_t > HeapChunkSelector::get_pdev_ids() const {
+    std::list< uint32_t > pdev_ids;
+    for (const auto& [pdev_id, _] : m_per_dev_heap) {
+        pdev_ids.emplace_back(pdev_id);
+    }
+    return pdev_ids;
+}
+
+std::shared_ptr< HeapChunkSelector::ExtendedVChunk > HeapChunkSelector::select_chunk_from_pdev(uint32_t pdev_id) {
```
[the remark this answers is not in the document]
JacksonYao287: will take a look

src/lib/homestore_backend/heap_chunk_selector.cpp
```diff
@@ -421,4 +420,87 @@ uint64_t HeapChunkSelector::total_blks(uint32_t dev_id) const {
...
+    pdev_heap->available_blk_count -= chunk->available_blks();
+
+    RELEASE_ASSERT(!chunk->m_pg_id.has_value(), "chunk {} is selected from pdev heap, but it has a pg_id {}!",
+                   chunk->get_chunk_id(), chunk->m_pg_id.value());
+
+    return chunk;
+}
+
+std::shared_ptr< HeapChunkSelector::ExtendedVChunk >
+HeapChunkSelector::select_specific_chunk_from_pdev(uint32_t pdev_id, const chunk_num_t chunk_id) {
```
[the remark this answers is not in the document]
JacksonYao287: ok, will take a look

JacksonYao287 (commented):
thanks @xiaoxichen  for the comments, I will rebase and update this PR

src/lib/homestore_backend/gc_manager.cpp
xiaoxichen: Good point on multiple emergent GC.

This is a chicken-egg problem :  If we choose to put all GC candidate into a Q, then we need a priority queue to prioritize Emergent GC, then the executor(threads) need to pulling on the Q,  it is a small optimization to use CV to avoid busy loop.

However,  if the design changed from very beginning, that do not put all GC candidate into Q,  just enough tasks to keep threads busy( meaning keeping only K submitted task  where K <= thread-num) .  Then we dont need to maintain priority as there will be no pending normal GC in the Q.  In this case, the task-Q built into folly thread pool is enough.

src/lib/homestore_backend/gc_manager.cpp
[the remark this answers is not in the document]
JacksonYao287: oh, I misunderstand your comment. now, I thinks the main difference between us is only how many chunks we we select for gc in a period.

mine: all the chunks that eligible for gc 
yours:  only topK chunks

the multi-priority task queue will not involve pop basically.  each priority has a dedicate normal queue. push and pop to the queue is basically o(1), which will not involve heap sort

so, the current implementation  O(N) (scan) +O(N) (or (N*O(1)) (insert)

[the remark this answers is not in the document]
xiaoxichen: the pop is involved in `reserve_specific_chunk`.

>generally speaking, it does not matter too much that which gc-candidate chunk will be gc first , since the garbage rate of them are close to each other

I can agree on this based on the assumption "we will have GC_GARBAGE_RATE_THRESHOLD to high number".

It is more about the thread-model and Q, if only get K chunks each time, will it simplify the implementation ?

the best candidate get GC first,  as a bonus, can be achieve in the mean time. 


Also,  I would suggest separate the concurrency (number of on-going GC) away from reserve_chunk_num,  the concurrency should be <= reserve_chunk is fine. The main reason is #reserve_chunk is defined when bootstrap an SM,  however concurrent should be a runtime tuning.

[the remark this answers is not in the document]
JacksonYao287: >however concurrent should be a runtime tuning.

we can add a semaphore to control if we really need,  it is easy to achieve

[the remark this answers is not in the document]
JacksonYao287: or we can postpone this to the time we really need it

src/lib/homestore_backend/gc_manager.cpp
```diff
@@ -0,0 +1,360 @@
...
+    if (!GCManager::is_eligible_for_gc(move_from_vchunk)) {
+        LOGWARN("move_from_chunk={} is not eligible for gc, so we can not process the gc task", move_from_chunk);
+        task.complete(false);
+        return;
+    }
+
+    LOGINFO("move_from_chunk={} belongs to pg {} ", move_from_chunk, move_from_vchunk->m_pg_id.value());
+
+    // make chunk to gc state, so that it can be select for creating shard
+    auto succeed = m_chunk_selector->try_mark_chunk_to_gc_state(
```
JacksonYao287, line 277: @xiaoxichen I now use a new state `gc` for vchunk, so that it will not be selected for creating shard

src/lib/homestore_backend/gc_manager.cpp
```diff
@@ -0,0 +1,360 @@
...
+    for (uint32_t i = 0; i < RESERVED_CHUNK_NUM_PER_PDEV; i++) {
+        m_gc_executor->add([this] {
+            while (!m_is_stopped.load(std::memory_order_relaxed)) {
+                GCManager::gc_task task;
+                if (m_gc_task_queue.try_dequeue(task)) {
+                    process_gc_task(task);
+                } else {
+                    // if we use a condition variable(cv), when stop, the cv might lost the notification. so, it`s
+                    // better to do sleep and check
+                    LOGINFO("no gc task found, sleep 1s!");
```
JacksonYao287, line 143: @xiaoxichen I recall why I use this busy loop instead of a condition variable notification.

when we stop, we set the stop flag and notify the gc thread. but if the gc thread is working on gc and not cv.wait(), it will lose this notification and never be notified again. we need a more complex mechanism to achieve. I can later use a seperate PR to do this .

src/lib/homestore_backend/heap_chunk_selector.cpp
```diff
@@ -421,4 +438,88 @@ uint64_t HeapChunkSelector::total_blks(uint32_t dev_id) const {
...
+    RELEASE_ASSERT(chunk->available_blks() == chunk->get_total_blks(),
+                   "the selected chunk should be empty, chunk_id={}, pdev_id={}", chunk->get_chunk_id(), pdev_id);
+    RELEASE_ASSERT(!chunk->m_pg_id.has_value(), "chunk {} is selected from pdev heap, but it has a pg_id {}!",
+                   chunk->get_chunk_id(), chunk->m_pg_id.value());
+
+    return chunk;
+}
+
+std::shared_ptr< HeapChunkSelector::ExtendedVChunk >
+HeapChunkSelector::select_specific_chunk_from_pdev(uint32_t pdev_id, const chunk_num_t chunk_id) {
```
xiaoxichen, line 474: The only user of this function is the recovery of reserve chunk.

We have similar requirement that exclude chunks belongs to PG from  adding into the per_dev_heap,  can we use similar approach , i.e changing `recover_per_dev_chunk_heap` to exclude reserve chunk as well?

Then we can kill this function

src/lib/homestore_backend/gc_manager.cpp
```diff
@@ -0,0 +1,384 @@
...
+
+    for (const auto& [_, chunk] : m_chunk_selector->get_all_chunks()) {
+        if (is_eligible_for_gc(chunk)) {
+            auto pdev_id = chunk->get_pdev_id();
+            auto it = m_pdev_gc_actors.find(pdev_id);
+            if (it != m_pdev_gc_actors.end()) {
+                auto& actor = it->second;
+                auto chunk_id = chunk->get_chunk_id();
+                auto future = actor->add_gc_task(static_cast< uint8_t >(task_priority::normal), chunk_id);
+                if (future.isReady()) {
```
xiaoxichen, line 102: the future has return value , can we check return value bool?

xiaoxichen (commented):
lgtm in general

src/lib/homestore_backend/heap_chunk_selector.cpp
JacksonYao287: good, point, will take a look

JacksonYao287: I have remove those two function, ptal

JacksonYao287:
@xiaoxichen I have remove the fun you mentions , ptal

xiaoxichen (approved):

PR #287, opened 2025-04-21
Add on_commit for list of blkid's.

sanebay:


conanfile.py
```diff
@@ -9,7 +9,7 @@
 
 class HomeObjectConan(ConanFile):
     name = "homeobject"
-    version = "2.3.11"
+    version = "2.3.12"
 
     homepage = "https://github.com/eBay/HomeObject"
```
yamingk, line 14: Can we change the dep to HomeStore from
```self.requires("homestore/[>=6.9.0]@oss/master")``` 
To:
```self.requires("homestore/[~6.10.0]@oss/master")```

so that any HomeStore minor version update won't break HomeObject build?

sanebay: Done

xiaoxichen:
We need to rework this change after HS pr merged
https://github.com/eBay/HomeStore/pull/700/files

xiaoxichen (approved):

PR #293, opened 2025-04-28
handling config change where we also need to check no_space_left error info

JacksonYao287:
now that we have continuous notifications of the last committed lsn, we can always set error info and wait until got the notification of this lsn is committed.

this way, we no longer need to pause statemachine and check whether the last append lsn is committed


xiaoxichen (approved):
lgtm
@yuwmao @Besroy pls take a look as well.

PR #294, opened 2025-05-08
Enhance BR read_snapshot_obj

yuwmao:
This pr is used to fix [SH issue48](https://docs.google.com/document/d/16DMqv9J-JuNs5c25IBuX01QXe5WbxPm4B4hAfz4JNZM/edit?tab=t.0#heading=h.8j5ij57iwi9) 

- Use a shared_ptr< PGBlobIterator >* instead of PGBlobIterator* as user_ctx to avoid freeing ctx which is being used;
- Add a lock to avoid race condition of accessing user_ctx. In `read_snapshot_obj`, only lock the part getting shared_ptr< PGBlobIterator > from the user_ctx rather than the whole read process to avoid blocking free_user_snp_ctx, because read could spend a long time while free_user_snp_ctx could be called by become_leader which should not wait for it.
- Reset cursor to the beginning when next obj id is inconsistent with the cursor. In this case, follower will validate the context lsn and reset the cursor to resume from its checkpoint.

### DIFF
--- a/conanfile.py
+++ b/conanfile.py
@@ -9,7 +9,7 @@ required_conan_version = ">=1.60.0"
 
 class HomeObjectConan(ConanFile):
     name = "homeobject"
-    version = "2.3.16"
+    version = "2.3.17"
 
     homepage = "https://github.com/eBay/HomeObject"
     description = "Blob Store built on HomeReplication"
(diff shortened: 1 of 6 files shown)


xiaoxichen (commented):
in current implementation， the `snp_obj->user_ctx` seems redundant? 

Any possibility that there are two concurrent read_snapshot? for example two followers are all experiencing BR?

yuwmao:
> in current implementation， the `snp_obj->user_ctx` seems redundant?
> 
> Any possibility that there are two concurrent read_snapshot? for example two followers are all experiencing BR?

Yes, you are right, concurrent BR is possible and it's hard for us to differentiate followers in HO.

xiaoxichen (approved):

Besroy (approved):
LGTM

PR #295, opened 2025-05-09
ignore no_space_left on a higher lsn than the one we are waiting for

JacksonYao287:
if lsn 11 triggers no_space_left on chunk 0 , then we will set and error info [lsn:10, chunk_id:0]. 

now, lsn 13 might comes later than lsn 11, and trigger another no_space_left , we need ignore this case, but not assert. 

this is the log example,

1 we received lsn 13- 38 and  lsn 16 triggers no_space_left, so we set error info [lsn:15, chunk_id:0]， existing_target_lsn=15 in item 3

```
/25 21:26:31.315] [storage_mgr] [trace] [68] [raft_repl_dev.cpp:1402:raft_event] [traceID=n/a] [rdev0:fceb0ab7-ebd5-466b-9028-e66712667a38] Raft channel: Received 26 append entries on follower from leader, term 1, lsn 13 ~ 38 , my committed lsn 10 , leader committed lsn 33
```

2 we received lsn 16- 21
```
 [05/08/25 21:26:34.654] [storage_mgr] [trace] [68] [raft_repl_dev.cpp:1402:raft_event] [traceID=n/a] [rdev0:fceb0ab7-ebd5-466b-9028-e66712667a38] Raft channel: Received 6 append entries on follower from leader, term 1, lsn 16 ~ 21 , my committed lsn 15 , leader committed lsn 39
```

3 existing lsn 15 ,  and lsn 18 (17+1) triggers another no_space_left, which will lead to the assert failure
```
[critical] [68] new target lsn should be less than or equal to the existing target lsn, new_target_lsn=17, existing_target_lsn=15
```

### DIFF
--- a/conanfile.py
+++ b/conanfile.py
@@ -9,7 +9,7 @@ required_conan_version = ">=1.60.0"
 
 class HomeObjectConan(ConanFile):
     name = "homeobject"
-    version = "2.3.16"
+    version = "2.3.17"
 
     homepage = "https://github.com/eBay/HomeObject"
     description = "Blob Store built on HomeReplication"
--- a/src/lib/homestore_backend/replication_state_machine.cpp
+++ b/src/lib/homestore_backend/replication_state_machine.cpp
@@ -830,10 +830,13 @@ void ReplicationStateMachine::on_no_space_left(homestore::repl_lsn_t lsn, homest
 
     const auto [target_lsn, error_chunk_id] = get_no_space_left_error_info();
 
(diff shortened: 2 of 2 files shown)


JacksonYao287:
this issue is caused by filp , which can not always guarantee on the same lsn

xiaoxichen:
can I comfirm this is NOT AN ISSUE ?

JacksonYao287:
> can I comfirm this is NOT AN ISSUE ?

yes, this not an issue, it caused by flip

PR #297, opened 2025-05-14
Update blk_size from 1K to 4K to align with the physical sector size of HDDs

Besroy:
This change also supports larger blob sizes due to the limitation of max_blks_per_blkid


src/lib/homestore_backend/hs_homeobject.hpp
```diff
@@ -45,7 +45,7 @@ class HSHomeObject : public HomeObjectImpl {
     inline static auto const _snp_rcvr_meta_name = std::string("SnapshotReceiver");
     inline static auto const _snp_rcvr_shard_list_meta_name = std::string("SnapshotReceiverShardList");
     static constexpr uint64_t HS_CHUNK_SIZE = 2 * Gi;
-    static constexpr uint32_t _data_block_size = 1024;
+    static constexpr uint32_t _data_block_size = 4096;
```
JacksonYao287, line 48: 4096 -> 4 * Ki

xiaoxichen (approved):

PR #299, opened 2025-05-14
Move snapshot data write to async mode.

xiaoxichen:
In order to provider higher queue depth for disk, which is optimal for scheduler to merge requests.

Also IO and compute(checksum) can parallel.

### DIFF
--- a/src/lib/homestore_backend/pg_blob_iterator.cpp
+++ b/src/lib/homestore_backend/pg_blob_iterator.cpp
@@ -90,8 +90,8 @@ void HSHomeObject::PGBlobIterator::reset_cursor() {
     cur_obj_id_ = {0, 0};
     cur_shard_idx_ = -1;
     std::vector< BlobInfo > cur_blob_list_{0};
-    cur_start_blob_idx_=0;
-    cur_batch_blob_count_=0;
+    cur_start_blob_idx_ = 0;
+    cur_batch_blob_count_ = 0;
     cur_batch_start_time_ = Clock::time_point{};
 }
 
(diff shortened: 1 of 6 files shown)


src/lib/homestore_backend/snapshot_receive_handler.cpp
```diff
@@ -241,65 +260,68 @@ int HSHomeObject::SnapshotReceiveHandler::process_blobs_snapshot_data(ResyncBlob
...
+    if (ec != std::error_code{}) {
+        LOGE("Errors in writing this batch, code={}, message={}", ec.value(), ec.message());
+        for (auto rit = allocated_blk_ids.rbegin(); rit != allocated_blk_ids.rend(); ++rit) {
+            auto err = homestore::data_service().async_free_blk(*rit).get();
+            if (err) {
+                LOGD("Freed blk_id={} due to failure in adding blob info, err {}", rit->to_string(),
+                     err ? err.message() : "nil");
+            }
+        }
+        std::unique_lock< std::shared_mutex > lock(mutex);
```
koujl, line 318: Seems should be
```
std::unique_lock< std::shared_mutex > lock(ctx_->progress_lock);
```

yuwmao:
If a blob batch returns an error due to alloc/write failure, the async_write might still be in progress. 
Then when the retry batch with the same blobs comes, is there any problem? Seems no correctness issue due to the duplication handling but could have some space waste(the previous write might not be committed in indexservice when second write happens)?

src/lib/homestore_backend/pg_blob_iterator.cpp
```diff
@@ -288,32 +285,44 @@ bool HSHomeObject::PGBlobIterator::create_blobs_snapshot_data(sisl::io_blob_safe
...
+            blob_entries.push_back(CreateResyncBlobDataDirect(builder_, res->blob_id_, (uint8_t)res->state_, &data));
         }
-
-        HISTOGRAM_OBSERVE(*metrics_, snp_dnr_blob_process_latency, get_elapsed_time_us(blob_start));
-        std::vector< uint8_t > data(blob.cbytes(), blob.cbytes() + blob.size());
-        blob_entries.push_back(CreateResyncBlobDataDirect(builder_, info.blob_id, (uint8_t)state, &data));
-        total_bytes += blob.size();
+    }
+    if (hit_error) {
+        return false;
```
yuwmao, line 325: Needs to call builder_.Clear() before return.

PR #300, opened 2025-05-20
add default constructor for PGStats with initialization list

Hooper9973:
add default constructor for PGStats with initialization list


yuwmao (approved):
LGTM

PR #305, opened 2025-06-03
Prefetching batch data.

xiaoxichen:


src/lib/homestore_backend/pg_blob_iterator.cpp
```diff
@@ -251,18 +252,17 @@ BlobManager::AsyncResult< blob_read_result > HSHomeObject::PGBlobIterator::load_
         });
 }
 
-bool HSHomeObject::PGBlobIterator::create_blobs_snapshot_data(sisl::io_blob_safe& data_blob) {
-    auto batch_start = Clock::now();
+bool HSHomeObject::PGBlobIterator::prefetch_blobs_snapshot_data() {
     std::vector< ::flatbuffers::Offset< ResyncBlobData > > blob_entries;
```
Besroy, line 256: LGTM. Just a small reminder: blob_entries is not being used in this function, so it can be removed.

Besroy (dismissed):

Besroy (dismissed):

src/lib/homestore_backend/snapshot_receive_handler.cpp
```diff
@@ -442,45 +442,48 @@ shard_id_t HSHomeObject::SnapshotReceiveHandler::get_next_shard() const {
...
-        sb->progress.complete_bytes = ctx_->progress.complete_bytes;
-        sb->progress.complete_blobs = ctx_->progress.complete_blobs;
-        sb->progress.corrupted_blobs = ctx_->progress.corrupted_blobs;
-    }
-
-    hs_pg->snp_rcvr_info_sb_.write();
-
+    RELEASE_ASSERT(home_obj_.get_hs_pg(ctx_->pg_id) != nullptr, "PG not found, pg={}", ctx_->pg_id);
+    // ensure previous cp finished.
+    std::move(cp_fut).get();
```
Besroy, line 447: Not sure about the BR representation, but I have a small concern: If the follower crashes after sending LAST_OBJ_ID but before the final shard flush (cp+sb) is complete, will the last shard be retransmitted?

xiaoxichen: this is a nice catch

src/lib/homestore_backend/pg_blob_iterator.cpp
```diff
@@ -309,30 +322,76 @@ bool HSHomeObject::PGBlobIterator::create_blobs_snapshot_data(sisl::io_blob_safe
...
+
+        LOGT("Getting prefetched data for blob {}", info.blob_id);
+        auto it = prefetched_blobs.find(info.blob_id);
+        if (it == prefetched_blobs.end()) {
+            hit_error = true;
+            LOGE("blob {} not found in prefetched blob map", info.blob_id);
+            break;
+        }
+        auto res = std::move(it->second).get();
+        prefetched_blobs.erase(it);
```
yuwmao, line 365: If it is removed here, then the retry batch will fail in L359.

xiaoxichen: the retry batch will re-fetch， isnt it？

xiaoxichen: no we cannot , the future can only be consumed once (i.e  you cannot .get() a future twice).

If we want to do optimization to save the extra read,   we need to create another future,  fulfilled it immediately with the data we just got, and put the new future back to the map.

Besroy (dismissed):
LGTM

src/lib/homestore_backend/snapshot_receive_handler.cpp
```diff
@@ -459,45 +459,52 @@ shard_id_t HSHomeObject::SnapshotReceiveHandler::get_next_shard() const {
...
+            sb->progress.complete_bytes = ctx_->progress.complete_bytes;
+            sb->progress.complete_blobs = ctx_->progress.complete_blobs;
+            sb->progress.corrupted_blobs = ctx_->progress.corrupted_blobs;
+        }
+        hs_pg->snp_rcvr_info_sb_.write();
+        return success;
+    });
+    // sync wait for last shard before returning LAST_OBJ_ID.
+    if (get_next_shard() == shard_list_end_marker) {
+	    std::move(cp_fut).get();
```
yuwmao, line 506: I have a question about the last batch(not only limited in this pr). 
If the follower crashes after sb cp flush or just leader doesn't receive the response, what would happen if leader retries this last batch? Because the shard_cursor is already set to shard_list_end_marker, then the offset validation will fail, reset offset to a weird `shard_list_end_marker|0`, and leader's update_cursor fails, reset its cursor and send empty data, finally follower gets error due to invalid data.

PR #307, opened 2025-06-06
BaselineResync: Enhance objID validation on follower

koujl:
With `snapshot_ctx_timeout` set to a much larger value, the likelihood increases that the follower may restart within the timeout and receives the previous message which should be sent before the restart.

An additional validation is added to prevent unexpected behavior and reset the cursor to the beginning of the current shard on invalid objID.


xiaoxichen (approved):
lgtm.
Great to see we are fixing bugs in this path

PR #308, opened 2025-06-10
optimize blob deletion

JacksonYao287:
1 do not schedule a separate get before update , using existing_value to get the existing blob pba in a single update.
2 move the durable entity change into thenValue of async_free_blks.

### DIFF
--- a/src/lib/homestore_backend/hs_blob_manager.cpp
+++ b/src/lib/homestore_backend/hs_blob_manager.cpp
@@ -532,8 +532,8 @@ void HSHomeObject::on_blob_del_commit(int64_t lsn, sisl::blob const& header, sis
     auto status = index_table->put(update_req);
 
     if (sisl_unlikely(status == homestore::btree_status_t::not_found)) {
-        // in baseline resync case, the blob might has already been deleted at leader and follower deos not receives
-        // this blob, so when committing this log,  not_found will happpen.
+        // in baseline resync case, the blob might have already been deleted at leader and follower does not receive
+        // this blob, so when committing this log, not_found will happen.
         LOGW("shard_id={}, blob_id={} not found, probably already deleted, lsn={}", shard_id, blob_id, lsn);
     } else if (sisl_unlikely(status != homestore::btree_status_t::success)) {
         // if we return directly, for the perspective of statemachine, this lsn has been committed successfully. so


src/lib/homestore_backend/hs_blob_manager.cpp
```diff
@@ -518,37 +518,43 @@ void HSHomeObject::on_blob_del_commit(int64_t lsn, sisl::blob const& header, sis
...
-    LOGD("shard_id={}, blob_id={} has been moved to tombstone, lsn={}", blob_info.shard_id, blob_info.blob_id, lsn);
+    LOGD("shard_id={}, blob_id={} has been moved to tombstone, lsn={}", shard_id, blob_id, lsn);
 
-    auto& multiBlks = r.value();
-    if (multiBlks != tombstone_pbas) {
-        repl_dev->async_free_blks(lsn, multiBlks);
-        const_cast< HS_PG* >(hs_pg)->durable_entities_update([](auto& de) {
-            de.active_blob_count.fetch_sub(1, std::memory_order_relaxed);
-            de.tombstone_blob_count.fetch_add(1, std::memory_order_relaxed);
+    auto existing_pbas = existing_value.pbas();
```
xiaoxichen, line 541: This is an behavior change that the response of deleting a non-exists blob changed from UNKNOWN_BLOB to SUCCESS.

Also , if the blob doesnt exists , the existing_pba will be default value of all zero.  is that right to free in this case?

I was surprise there is no UT covered this but lets add one.

JacksonYao287: I have updated the code to cover the case that blob does not exist, will try to add a UT later

Besroy (dismissed):
Thanks for the fix for deleting blobs not_found

JacksonYao287:
A UT is added to exercise the case of deleting a non-exist blob

Besroy (approved):
LGTM. Just curious, what does `recovery_done_` do before this? During `HO start`, could there be errors other than 'delete blob not found'? If not, it looks good to remove `recover_done`.

JacksonYao287:
>  what does `recovery_done_` do before this? 

I believe this is a legacy useless variable.  no matter recovery_done_ is true or false, we  can not let it pass if fail to update index and the error code is not `not_found`.

PR #310, opened 2025-06-16
Change BlobHeader to include user key and block_hashes.

xiaoxichen:
As a preparation for partial read support, we need finer granuarity for checksum rather than blob level, so that partial read do not need amplificating to full blob for checksum verification.

support up to 64 blocks which translate to 256k block size when max_blob_size is 16MB.

Also changing user_key inline to blob_header.

Header should be exactly a physical block size (4KB).


yuwmao (approved):
LGTM

PR #315, opened 2025-06-24
return no_space_left error for put_blob

JacksonYao287:
if no_space_left happens when leader call put_blob, we nee return to error to the caller


Besroy (approved):
LGTM, thanks for the update

PR #320, opened 2025-06-29
enhance gc

JacksonYao287:
1 add task_id for each gc task
2 fix a bug: when fails to create a shard, we need to release the chunk, so that it can be gc

### DIFF
--- a/src/lib/homestore_backend/gc_manager.cpp
+++ b/src/lib/homestore_backend/gc_manager.cpp
@@ -169,7 +169,7 @@ bool GCManager::is_eligible_for_gc(chunk_id_t chunk_id) {
 
     const auto defrag_blk_num = chunk->get_defrag_nblks();
 
-    if (!defrag_blk_num) return false;
+    if (!defrag_blk_num) { return false; }
 
     // 1 if the chunk state is inuse, it is occupied by a open shard, so it can not be selected and we don't need gc it.
     // 2 if the chunk state is gc, it means this chunk is being gc, or this is a reserved chunk, so we don't need gc it.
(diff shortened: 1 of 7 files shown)


src/lib/homestore_backend/hs_shard_manager.cpp
```diff
@@ -192,20 +195,33 @@ ShardManager::AsyncResult< ShardInfo > HSHomeObject::_create_shard(pg_id_t pg_ow
...
+            // if (err == ShardError::NOT_LEADER) { err.current_leader = repl_dev->get_leader_id(); }
+
+            bool res = chunk_selector()->release_chunk(pg_owner, v_chunk_id);
+            RELEASE_ASSERT(res, "Failed to release v_chunk_id={}, pg={}", v_chunk_id, pg_owner);
+
+            SLOGE(tid, new_shard_id, "got {} when creating shard at leader, failed to create shard {}!", err,
+                  new_shard_id);
+
+            if (err == ShardError::NO_SPACE_LEFT) {
+                gc_manager()->submit_gc_task(task_priority::normal,
```
xiaoxichen, line 212: If this shard doesnt meet gc standard, why we want to GC it?

src/lib/homestore_backend/hs_blob_manager.cpp
```diff
@@ -423,9 +423,16 @@ HSHomeObject::blob_put_get_blk_alloc_hints(sisl::blob const& header, cintrusive<
         return folly::makeUnexpected(homestore::ReplServiceError::RESULT_NOT_EXIST_YET);
     }
 
-    homestore::blk_alloc_hints hints;
+    if ((*shard_iter->second)->info.state != ShardInfo::State::OPEN) {
```
xiaoxichen, line 426: I think this check can only be done on leader, if needed... rejecting log on follower doesnt make any sense.

JacksonYao287: it will be helpful both on leader and follower, the aim here is not only to reject the put_blob on an unopen shard,  but also  to avoid garbage blob in follower .

for leader, if the shard is unopen, we reject it to fail asysc_alloc_write.
for follower, if the shard is unopen, we reject it to avoid a garbage , since the new blob is trying to be placed into a unopen chunk

JacksonYao287: sorry, `reject` here is not that accurate  and a little misleading for follower. I think it`s better to use `drop` for follower, which means follower will ignore this request.
 
blk allocation will happen not only when received a raft log, but also when received a push_data request.

no matter where this blk allocation happens(raft channel or data channel), if the shard , which the blk is allocated for, is not open, this means this request(log or push_data) is a stale request, which probably wandered in the network for a long time, and now it comes. it we do not drop this reqeust, it will allocate a blk(it will become garbage) and not  on_commit will be called for this blob again. 

on the opposite side, what we should do if we receive a request to try to put_blob on an not open shard. I think we should not make this happen.

JacksonYao287: basically, for log, it`s a stale log, it will not reach here, since it will be filtered out by nuraft before `raft_repl_dev#raft_event` is called.

so , this case probably happen when we receive a stale push data

JacksonYao287: ok, I think you are correct. if we have a concurrent put_blob and seal_shard, for the same shard, we will have a put blob committed for a already sealed shard. thanks @xiaoxichen  for this good catch, I will remove this part.

we can later add a optimization for rejeceting the request if it is a push_data request

src/lib/homestore_backend/gc_manager.cpp
```diff
@@ -363,7 +370,7 @@ void GCManager::pdev_gc_actor::handle_recovered_gc_task(
                        move_to_chunk);
     }
 
-    if (!process_after_gc_metablk_persisted(gc_task_sb, valid_blob_indexes)) {
+    if (!process_after_gc_metablk_persisted(gc_task_sb, valid_blob_indexes, 0)) {
```
xiaoxichen, line 373: nit： do not use magic number 0

xiaoxichen (commented):
Please add summary to commit message.

Also,  

- Suggest moving GC to a dedicate mod (logdebugmod) 
- Suggest similar pattern like RD_LOG so that task_id can be print more consistent across the code (avoid extra space/capital by mistake).
```
define GC_LOG(level, task_id, msg, ...)                                                                               \
    LOG##level##MOD(GC, "[task_id={}] " msg, traceID, identify_str(), ##__VA_ARGS__)
```

src/lib/homestore_backend/gc_manager.cpp
JacksonYao287:
```
inline static atomic_uint64_t _gc_task_id{1}; // 0 is used for crash recoveryf
```

for the simplicity , 0 is dedicated for crash recovery

PR #321, opened 2025-07-03
Wait for prefetch complete before freeing pg_iter

koujl:


src/lib/homestore_backend/pg_blob_iterator.cpp
```diff
@@ -386,23 +397,17 @@ bool HSHomeObject::PGBlobIterator::create_blobs_snapshot_data(sisl::io_blob_safe
 
     if (hit_error) {
         builder_.Clear();
-        // drain the prefetched data before return which destroy the iterator.
-        for (auto& blob : prefetched_blobs) {
```
Besroy, line 390: Did I understand correctly: if hit error here, let nuraft call clear_snapshot_sync_ctx to trigger 'stop'?

koujl: Correct. The cleanup part will be done by `stop()`.

PR #323, opened 2025-07-04
Fix create shard

JacksonYao287:
1 when the follower receives creating shard but the pchunk of this vchunk is being gc, we need to wait until gc is completed

2 for the sake of safety, move the code , which change the vchunk state , into chunk selector so that they will be under the protection of the lock of chunkselector

3 add more detailed log , which will be helpful for debugging

### DIFF
--- a/src/lib/homestore_backend/gc_manager.cpp
+++ b/src/lib/homestore_backend/gc_manager.cpp
@@ -437,8 +437,9 @@ bool GCManager::pdev_gc_actor::replace_blob_index(
                        homestore::BtreeValue const& new_value) -> homestore::put_filter_decision {
                 BlobRouteValue existing_value{value_in_btree};
                 BlobRouteValue new_pba_value{new_value};
-
-                if (existing_value.pbas() == HSHomeObject::tombstone_pbas) {
+                const auto& existing_pbas = existing_value.pbas();
+                const auto& new_pbas = new_pba_value.pbas();
+                if (existing_pbas == HSHomeObject::tombstone_pbas) {
                     LOGDEBUG("gc task_id={}, remove tombstone when updating pg index after data copy , pg_id={}, "
                              "shard_id={}, blob_id={}, move_from_chunk={}, move_to_chunk={}",
                              task_id, pg_id, shard, blob, move_from_chunk, move_to_chunk);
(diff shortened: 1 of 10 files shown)


xiaoxichen (commented):
Can we just failed the append_entry  and let raft retry ?

JacksonYao287:
> Can we just failed the append_entry and let raft retry ?

if this happens in raft channel and  we return nullptr for `select_specific_chunk`, then no_space_left will be triggered, and trigger anther gc about this chunk.

src/lib/homestore_backend/gc_manager.cpp
```diff
@@ -915,9 +923,45 @@ bool GCManager::pdev_gc_actor::purge_reserved_chunk(chunk_id_t chunk) {
...
+        LOGERROR("Failed to query blobs after purging reserved chunk={} in gc index table, index ret={}", chunk,
+                 status);
+        return false;
+    }
+
+    if (!valid_blob_indexes.empty()) {
+        LOGERROR("gc index table is not empty for chunk={} after purging, valid_blob_indexes.size={}", chunk,
+                 valid_blob_indexes.size());
+        return false;
+    }
```
JacksonYao287, line 944: @shosseinimotlagh  sometimes I find that after I remove all the entries in a range , I can still get some entries in this range, which means it does not clear all the entries in the rage.
https://console.sherlock.io/explore?schemaVersion=1&panes=%7B%22us6%22:%7B%22datasource%22:%22ctGRFcsGz%22,%22queries%22:%5B%7B%22refId%22:%22A%22,%22expr%22:%22%7B_namespace_%3D%5C%22sdsapp%5C%22,%20applicationinstance%3D%5C%22nuobject2sm%5C%22,%20pod%3D%5C%22sm-long-running2-15%5C%22%7D%20%7C%3D%20%60gc%20task_id%3D155,%20can%20not%20purge%20move_to_chunk%3D22317%60%22,%22queryType%22:%22range%22,%22datasource%22:%7B%22type%22:%22loki%22,%22uid%22:%22ctGRFcsGz%22%7D,%22editorMode%22:%22builder%22,%22direction%22:%22backward%22%7D%5D,%22range%22:%7B%22from%22:%221751645136000%22,%22to%22:%22now%22%7D,%22panelsState%22:%7B%22logs%22:%7B%22visualisationType%22:%22logs%22%7D%7D%7D%7D&orgId=1

since range remove if very crucial for gc, and in the worst case data loss will happen.  so pls help check this.

JacksonYao287: also, the index crash(which I sent to homestore channel) also happened here when range_remove is called

JacksonYao287:
@xiaoxichen  this need some homestore side changes(to handle the the case of nullptr for the returned chunk) if we do not want to block here.

can we revisit here later?

xiaoxichen:
sure so already approved

PR #331, opened 2025-07-17
Fix: Adjust capacity calculation in `HSHomeObject::_get_stats`

Hooper9973:
- Updated the total capacity calculation to subtract reserved GC bytes based on the number of physical devices and reserved chunks per device.
- Added logic to compute used capacity using the number of pg used blocks and data block size.

### DIFF
--- a/src/lib/homestore_backend/heap_chunk_selector.cpp
+++ b/src/lib/homestore_backend/heap_chunk_selector.cpp
@@ -280,7 +280,8 @@ std::optional< uint32_t > HeapChunkSelector::select_chunks_for_pg(pg_id_t pg_id,
                                               });
     auto& pdev_heap = most_avail_dev_it->second;
     if (num_chunk > pdev_heap->size()) {
-        LOGWARNMOD(homeobject, "Pdev has no enough space to create pg={} with num_chunk={}", pg_id, num_chunk);
+        LOGWARNMOD(homeobject, "Pdev has no enough space to create pg={} with num_chunk={}, available_num_chunk={}",
+                   pg_id, num_chunk, pdev_heap->size());
         return std::nullopt;
     }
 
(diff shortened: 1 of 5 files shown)


xiaoxichen (dismissed):

src/lib/homestore_backend/tests/test_heap_chunk_selector.cpp
```diff
@@ -39,6 +39,9 @@ class Chunk : public std::enable_shared_from_this< Chunk > {
     uint16_t get_chunk_id() const { return m_chunk_id; }
 
     blk_num_t get_total_blks() const { return m_available_blks; }
+
+    blk_num_t get_used_blks() const { return m_available_blks; }
```
yuwmao, line 43: typo, `return m_used_blks`

src/lib/homestore_backend/heap_chunk_selector.cpp
```diff
@@ -581,6 +582,17 @@ uint64_t HeapChunkSelector::total_blks(uint32_t dev_id) const {
     return it->second->m_total_blks;
 }
 
+uint64_t HeapChunkSelector::get_used_blks() const {
+    std::shared_lock lock_guard(m_chunk_selector_mtx);
+    uint64_t used_blks = 0;
+    for (const auto& [_, pg_chunk_collection] : m_per_pg_chunks) {
+        for (const auto& chunk : pg_chunk_collection->m_pg_chunks) {
+            if (chunk->m_state == ChunkState::INUSE) { used_blks += chunk->get_used_blks(); }
```
yuwmao, line 590: There is a behavior change:
Previously, we didn't take INUSE into account, so the used_blks were actual physically used blocks.

src/lib/homestore_backend/tests/test_heap_chunk_selector.cpp
Hooper9973: fixed

yuwmao (approved):
LGTM

PR #334, opened 2025-08-11
Add rollback http API for replace_member

yuwmao:


src/lib/homestore_backend/hs_pg_manager.cpp
```diff
@@ -410,6 +398,26 @@ void HSHomeObject::on_pg_complete_replace_member(group_id_t group_id, const std:
...
+        if (repl_dev.group_id() == group_id) {
+            auto hs_pg = static_cast< HSHomeObject::HS_PG* >(pg.get());
+            if (pg->pg_info_.members.erase(PGMember(member)) > 0) {
+                hs_pg->update_membership(pg->pg_info_.members);
+                LOGI("PG remove member done member={} member_nums={}, trace_id={}", boost::uuids::to_string(member),
+                     pg->pg_info_.members.size(), tid);
+            } else {
+                LOGI("PG remove member done, member doesn't exist, member={} member_nums={}, trace_id={}",
+                     boost::uuids::to_string(member), pg->pg_info_.members.size(), tid);
+            }
```
JacksonYao287: should we break after this line? IIRC, we only update the superblk of the given PG, no?

src/lib/homestore_backend/hs_pg_manager.cpp
```diff
@@ -472,6 +611,51 @@ bool HSHomeObject::pg_destroy(pg_id_t pg_id, bool need_to_pause_pg_state_machine
...
+        auto lg = std::shared_lock(_pg_lock);
+        auto iter = std::find_if(_pg_map.begin(), _pg_map.end(), [group_id](const auto& entry) {
+            return pg_repl_dev(*entry.second).group_id() == group_id;
+        });
+        if (iter != _pg_map.end()) {
+            pg_id = iter->first;
+        } else {
+            // There is a known case during adding member: the new member may think itself already in group but actually
+            // not, so the pg is not created yet.
+            LOGI("no pg found, group_id={}, trace_id={}", group_id, tid);
```
JacksonYao287: should we return here directly?

PR #337, opened 2025-08-20
fix emergent gc in creating shard

JacksonYao287:
in most cases(put_blob and seal_shard), the last shard in the chunk, which triggers emergent gc, should be in open state. but if the emergent gc is triggered by a creat_shard request, then the last shard is not in open state, it is in sealed state.


yuwmao (dismissed):
LGTM

Besroy (dismissed):

src/lib/homestore_backend/gc_manager.cpp
```diff
@@ -856,7 +859,7 @@ bool GCManager::pdev_gc_actor::copy_valid_data(chunk_id_t move_from_chunk, chunk
 
                             // the shard that triggers emergent gc should be the last shard in the chunk, so it should
                             // be open and we skip writing the footer for this case.
```
Besroy, line 861: could you update the comment as well

PR #338, opened 2025-08-28
Add generic handle logic for non-data-op

Besroy:
This pr avoids the false alarm of corrupted message when HS adds ctrl type log but HO is not updated

### DIFF
--- a/src/lib/homestore_backend/replication_state_machine.cpp
+++ b/src/lib/homestore_backend/replication_state_machine.cpp
@@ -77,20 +77,16 @@ bool ReplicationStateMachine::on_pre_commit(int64_t lsn, sisl::blob const& heade
     // For shard creation, since homestore repldev inside will write shard header to data service first before this
     // function is called. So there is nothing is needed to do and we can get the binding chunk_id with the newly shard
     // from the blkid in on_commit()
-    if (ctx->op_code() == homestore::journal_type_t::HS_CTRL_START_REPLACE) {
-        LOGI("pre_commit HS_CTRL_START_REPLACE log entry, lsn={}", lsn);
-        return true;
-    }
-    if (ctx->op_code() == homestore::journal_type_t::HS_CTRL_COMPLETE_REPLACE) {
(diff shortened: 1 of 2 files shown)


src/lib/homestore_backend/replication_state_machine.cpp
```diff
@@ -91,6 +91,11 @@ bool ReplicationStateMachine::on_pre_commit(int64_t lsn, sisl::blob const& heade
         return true;
     }
 
+    if (ctx->op_code() == homestore::journal_type_t::HS_CTRL_UPDATE_TRUNCATION_BOUNDARY) {
+        LOGI("pre_commit HS_CTRL_UPDATE_TRUNCATION_BOUNDARY log entry, lsn={}", lsn);
+        return true;
+    }
+
```
xiaoxichen, line 98: IIUC the real issue before this commit was `homestore::journal_type_t::HS_CTRL_UPDATE_TRUNCATION_BOUNDARY` reaches L99 and got NPE somewhere?

Can we cleanup the L80 - L98 like
```
if (ctx->op_code() != homestore::journal_type_t::HS_DATA_LINKED) && (ctx->op_code() != homestore::journal_type_t::HS_DATA_INLINED)  {
        LOGI("pre_commit TO_STRING(ctx->op_code()) log entry, lsn={}", lsn);
        return true;
    }
else {
   assert(ctx->op_code() == journal_type_t::HS_DATA_LINKED || ctx->op_code() == journal_type_t::HS_DATA_INLINED);
       .....
}
```

JacksonYao287: i am curious why we hit this issue until now. Theoretically, any HS_CTRL_UPDATE_TRUNCATION_BOUNDARY will hit this issue.  Don`t we enable UPDATE_TRUNCATION_BOUNDARY in previous tests?

Besroy: This PR doesn't fix the crash issue; it simply reduces the log noise from corrupted messages due to unhandled control messages. 
Thank you for your comment ! It's a great idea, and I've updated the code

xiaoxichen (approved):

PR #349, opened 2025-10-14
improve gc log

JacksonYao287:
add gcmgr log mod, which will give us the capability to enable or disable gc related log through one single setting， like blobmgr and shardmgr

### DIFF
--- a/src/lib/homestore_backend/gc_manager.cpp
+++ b/src/lib/homestore_backend/gc_manager.cpp
@@ -44,6 +44,13 @@ GCManager::GCManager(HSHomeObject* homeobject) :
             on_gc_task_meta_blk_found(std::move(buf), voidptr_cast(mblk));
         },
         nullptr, true);
+
+    auto gcmgr_log_level = sisl::logging::GetModuleLogLevel("gcmgr");
+    if (gcmgr_log_level == spdlog::level::level_enum::err) {
+        // if we do not set gcmgr log level by start command line(for exampel, --log_mods gcmgr:info), then set it to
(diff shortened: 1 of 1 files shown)


xiaoxichen:
> LGTM. But I’m curious about the decision to extract a separate log mod. This change requires updating the log mods in the sm configmap used to start the SM, which would also necessitate a rolling update. Could you please add the reasoning to the description?

Default level should be properly set via 
https://github.com/eBay/sisl/blob/master/src/logging/test/example.cpp#L68
then the changes to sm configmap should be minimal.

src/lib/homestore_backend/gc_manager.cpp
```diff
@@ -33,6 +44,13 @@ GCManager::GCManager(HSHomeObject* homeobject) :
             on_gc_task_meta_blk_found(std::move(buf), voidptr_cast(mblk));
         },
         nullptr, true);
+
+    auto gcmgr_log_level = sisl::logging::GetModuleLogLevel("gcmgr");
+    if (gcmgr_log_level == spdlog::level::level_enum::err) {
+        // if we do not set gcmgr log level throught start command line(for exampel, --log_mods gcmgr:info), set it to
+        // debug level by default
+        sisl::logging::SetModuleLogLevel("gcmgr", spdlog::level::debug);
+    }
```
JacksonYao287: if we do not set the log level of gcmgr, we set it to debug by default.
otherwise, we do nothing and use the log level set by command line.
so, we do not make any change to config map
@xiaoxichen @Besroy

PR #355, opened 2025-11-03
Enhance UT

yuwmao:
Reproduce a stuck case in ReplaceMember UT:
```
[11/03/25 06:01:59+00:00] [I] [test_homeobject_dynamic_replica_0] [10722] [pg_manager.cpp:26:replace_member] [pg=1] replace member [00020123-4567-89ab-cdef-0123456789ab] with [00030123-4567-89ab-cdef-0123456789ab]  task_id [task_id] quorum [0] trace_id [0]
[11/03/25 06:01:59+00:00] [I] [test_homeobject_dynamic_replica_0] [10722] [hs_pg_manager.cpp:307:_replace_member] PG replace member initated member_out=00020123-4567-89ab-cdef-0123456789ab member_in=00030123-4567-89ab-cdef-0123456789ab trace_id=0
[11/03/25 06:01:59+00:00] [E] [test_homeobject_dynamic_replica_0] [10722] [raft_repl_dev.cpp:209:start_replace_member] [traceID=0] [rdev1:6d676b70-7f35-49c4-9d86-a029a2472593] Step1. Replace member, quorum safety check failed, active_peers=1, active_peers_exclude_out/in_member=1, required_quorum=2, commit_quorum=0, task_id=task_id
/home/runner/.conan2/p/b/homeo845fd3d185586/b/src/lib/homestore_backend/tests/test_homestore_backend_dynamic.cpp:301: Failure
Value of: r
  Actual: false
Expected: true
```

Previously, laggy_threshold=0 which means we don't allow the peers have any lag logs([get_active_peers](https://github.com/eBay/HomeStore/blob/7f0616dcd0b117a21637bfbdfb205756557dfccd/src/lib/replication/repl_dev/raft_repl_dev.cpp#L1909-L1911)). 
Since we already verify the blob from index before proceed to replace member, so it's ok to loosen the active_peer check.

### DIFF
--- a/src/lib/homestore_backend/tests/homeobj_fixture.hpp
+++ b/src/lib/homestore_backend/tests/homeobj_fixture.hpp
@@ -242,7 +242,7 @@ public:
     std::map< shard_id_t, std::map< blob_id_t, uint64_t > >
     put_blobs(std::map< pg_id_t, std::vector< shard_id_t > > const& pg_shard_id_vec, uint64_t const num_blobs_per_shard,
               std::map< pg_id_t, blob_id_t >& pg_blob_id, bool need_sync_before_start = true) {
-        g_helper->sync();
+        if (need_sync_before_start) { g_helper->sync(); }
         std::map< shard_id_t, std::map< blob_id_t, uint64_t > > shard_blob_ids_map;
         for (const auto& [pg_id, shard_vec] : pg_shard_id_vec) {
             if (!am_i_in_pg(pg_id)) continue;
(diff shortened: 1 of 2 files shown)


JacksonYao287 (dismissed):
LGTM

xiaoxichen (approved):
LGTM

Context HBI = Heart Beat Interval.

Sequence during node removal:

1. The leader sends leave_cluster_request to the node being removed (out_member).
2. The out_member sets steps_to_down = 2 and replies to the leader.
This means the out_member will stop initiating leader elections after two HBI.
This state is persisted to disk; after restart, its election timer will not be started.
3. The leader receives leave_cluster_response and proposes a configuration change to remove the out_member.
4. After the leader commits the config change, if the out_member does not respond within 5 × HBI, the leader force-removes the out_member.


Known issue 

If the out_member’s state machine is stuck (e.g., due to no space left, missing shard, or similar errors), it completes step 2 but fails to apply/commit the leader’s config change from step 3. The leader proceeds to step 4/5 and removes the out_member. Because the out_member’s “steps_to_down” state is persisted, the out_member remains in a semi-disabled state and cannot recover or clean itself up even after restart, leaving it unable to rejoin or clear its down state.

PR #361, opened 2025-11-07
Enhance PGBlobIterator stop() protection with operation-level mutex

koujl:
Use an operation-level mutex to serialize all public operations and 
shutdown. The previous prefetch_lock_ was insufficient to prevent access
after stop() while stop() could be invoked while the instance was still
running other operations, causing it to be deleted mid-operation.


xiaoxichen (commented):
please add more background in the commit message

yuwmao (dismissed):
LGTM

xiaoxichen (approved):

PR #366, opened 2025-11-28
build chunk heap before joining raft group

JacksonYao287:
we need to build chunk heap before joining raft group, otherwise , if repl_dev receives a snapshot and try to destroy pg before chunk heap is built and after repl_dev has joined raft group, a crash will happen


src/lib/homestore_backend/hs_homeobject.cpp
```diff
@@ -237,41 +237,6 @@ void HSHomeObject::init_homestore() {
             });
         }
 
-        // We dont have any repl dev now, explicitly call init_completed_cb() where we register PG/Shard meta types.
-        repl_app->on_repl_devs_init_completed();
```
JacksonYao287: this will be called when initializing raft_repl_service, no need to call it here explicitly

xiaoxichen (approved):
lgtm.

PR #367, opened 2025-12-01
add aggressive check for copied blobs

JacksonYao287:
we find a potential bug of index table that will cause some blobs missing in gc index table after put, so add this aggressive check to make sure the copied blobs are the same as that in gc index table

moreover, this PR also fix some typos.


JacksonYao287:
I prefer to merge it since this can strongly protect our production data correctness. also , as you said, we will keep trying to reproduce this issue and see if we can reproduce it successfully or get any new finding.

PR #370, opened 2025-12-03
Enable partial blob data read

koujl:
This PR contains 2 commits:

* Refactor verify_blob to eliminate duplicate codes
* Enable partial data reading for HDD performance

Suggest reviewing by commits for better clarity.


xiaoxichen (approved):
lgtm

PR #373, opened 2025-12-11
Fix resync tests by enabling log truncation

koujl:
Log truncation in resync tests has not worked since [HomeStore#763](https://github.com/eBay/HomeStore/pull/763). 

Add missing raft_logstore_reserve_threshold config and trigger checkpoints on blob writes to enable proper log truncation.


Besroy (approved):
LGTM. Does HS ut require the same change?

PR #377, opened 2025-12-24
return delete marker for tombstone blob in on_fetch_data

JacksonYao287:
this PR also:
1 add more detailed info in gc log
2 make gc data verification configurable.


xiaoxichen:
please rebase and fix ut

xiaoxichen (approved):

PR #378, opened 2025-12-25
add a config to allow get_blob without checking traffic ready or not

Besroy:
This is a workaround for 1-replica issue

### DIFF
--- a/src/lib/homestore_backend/hs_backend_config.fbs
+++ b/src/lib/homestore_backend/hs_backend_config.fbs
@@ -47,6 +47,9 @@ table HSBackendSettings {
     // Number of times to pause the state machine
     state_machine_pause_retry_count: uint8 = 3;
 
+    // Check traffic ready before get, default true, set false for 1-replica workaround mode to allow read
+    check_traffic_ready_before_get: bool = true;
+
 }
 
 root_type HSBackendSettings;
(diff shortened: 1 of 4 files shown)


src/lib/homestore_backend/hs_blob_manager.cpp
```diff
@@ -294,7 +294,10 @@ BlobManager::AsyncResult< Blob > HSHomeObject::_get_blob(ShardInfo const& shard,
     RELEASE_ASSERT(repl_dev != nullptr, "Repl dev instance null");
     RELEASE_ASSERT(index_table != nullptr, "Index table instance null");
 
-    if (!repl_dev->is_ready_for_traffic()) {
+    // if all followers are down, leader will trigger re-election, this way will increase the log idx but new logs
+    // don't have chance to be committed, which let leader not ready for traffic
+    // so here we add a config to define whether check ready for traffic before get blob as a workaround for this issue
```
xiaoxichen: could you explain more on this? 
when leader trigger the re-election and no follower respond, will it go through the become_leader_cb one more time?

Besroy: I can't clearly recall what happened in cluster `isofruit-apple `at that time. It’s possible that the leader was re-elected multiple times with `become_leader`, as the error log at the time was `Not yet ready for traffic, committed to 82223 but gate is 82226`. However, by reviewing the NuRaft code, I found that if a follower goes down during the `become-leader` process, a similar issue could also occur. So I added this scenario to the comment.

src/lib/homestore_backend/hs_backend_config.fbs
```diff
@@ -44,6 +44,9 @@ table HSBackendSettings {
     // Number of times to pause the state machine
     state_machine_pause_retry_count: uint8 = 3;
 
+    // Check traffic ready before get, default true, set false for 1-replica workaround mode to allow read
+    check_traffic_ready_before_get: bool = true;
```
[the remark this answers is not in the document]
yuwmao: Okay.

JacksonYao287 (approved):

PR #379, opened 2025-12-26
add trigger gc api

JacksonYao287:
1 trigger gc for all chunks
curl -X POST "http://127.0.0.1:5000/api/v1/trigger_gc"

2 trigger gc for a specific chunk
curl -X POST "http://127.0.0.1:5000/api/v1/trigger_gc?chunk_id=87"

3 trigger gc for a specific pg
curl -X POST "http://127.0.0.1:5000/api/v1/trigger_gc?pg_id=87"

4 query the status of a gc job. ATM, we only support retain the status of the latest 100 gc jobs
curl -k "https://localhost:10443/api/v1/gc_job_status?job_id=trigger-gc-task-0"

5 query the status of all gc jobs. 
curl -k "https://localhost:10443/api/v1/gc_job_status"

all the above apis have been tested manually,  UT will be added later in a separate PR

### DIFF
--- a/src/lib/homestore_backend/hs_http_manager.cpp
+++ b/src/lib/homestore_backend/hs_http_manager.cpp
@@ -48,7 +48,11 @@ HttpManager::HttpManager(HSHomeObject& ho) : ho_(ho) {
         {Pistache::Http::Method::Get, "/api/v1/chunk/dump",
          Pistache::Rest::Routes::bind(&HttpManager::dump_chunk, this)},
         {Pistache::Http::Method::Get, "/api/v1/shard/dump",
-         Pistache::Rest::Routes::bind(&HttpManager::dump_shard, this)}};
+         Pistache::Rest::Routes::bind(&HttpManager::dump_shard, this)},
+        {Pistache::Http::Method::Get, "/api/v1/trigger_gc",
+         Pistache::Rest::Routes::bind(&HttpManager::trigger_gc, this)},
+        {Pistache::Http::Method::Get, "/api/v1/gc_job_status",
+         Pistache::Rest::Routes::bind(&HttpManager::get_gc_job_status, this)}};
 
     auto http_server = ioenvironment.get_http_server();
     if (!http_server) {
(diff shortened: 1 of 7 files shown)


src/lib/homestore_backend/hs_http_manager.cpp
```diff
@@ -239,6 +243,301 @@ void HttpManager::dump_shard(const Pistache::Rest::Request& request, Pistache::H
...
+                        if (priority == task_priority::emergent) { hs_pg->repl_dev_->clear_chunk_req(chunk_id); }
+
+                        // Submit GC task for this chunk
+                        auto future = gc_mgr->submit_gc_task(priority, chunk_id);
+                        pg_futures.push_back(std::move(future));
+                        LOGDEBUG("GC job {} submitted task for chunk_id={} in PG {} with priority={}", job_id, chunk_id,
+                                 pg_id, (priority == task_priority::emergent) ? "emergent" : "normal");
+                    }
+
+                    // 3.5 Wait for all GC tasks in this PG to complete
```
yuwmao: It may take a long time, right? Just return the job id and let user query the job status?

src/lib/homestore_backend/hs_http_manager.cpp
```diff
@@ -239,6 +243,301 @@ void HttpManager::dump_shard(const Pistache::Rest::Request& request, Pistache::H
...
+                (current_gc_job_->status == GCJobStatus::PENDING || current_gc_job_->status == GCJobStatus::RUNNING)) {
+                nlohmann::json error;
+                error["error"] = "A GC job is already in progress";
+                error["active_job_id"] = current_gc_job_->job_id;
+                error["active_job_status"] = (current_gc_job_->status == GCJobStatus::PENDING) ? "pending" : "running";
+                response.send(Pistache::Http::Code::Conflict, error.dump());
+                return;
+            }
+            // Create new job and set as current (replacing any completed job)
+            current_gc_job_ = job_info;
```
yuwmao: If we want to specify the chunk_id, then we can only clean chunks one by one, can we add some concurrency?

src/lib/homestore_backend/hs_http_manager.hpp
```diff
@@ -33,12 +38,47 @@ class HttpManager {
...
+    void get_gc_job_status(const Pistache::Rest::Request& request, Pistache::Http::ResponseWriter response);
 
 #ifdef _PRERELEASE
     void crash_system(const Pistache::Rest::Request& request, Pistache::Http::ResponseWriter response);
 #endif
 
+private:
+    enum class GCJobStatus { PENDING, RUNNING, COMPLETED, FAILED };
+
+    struct GCJobInfo {
```
yuwmao: From user's perspective, suggest adding `pg_id` as well.

src/lib/homestore_backend/hs_http_manager.cpp
JacksonYao287: yes, if too many chunk need to be gc, it probably take a while, so just return a job id, with which the user can query the job status.

src/lib/homestore_backend/hs_http_manager.cpp
```diff
@@ -239,6 +243,235 @@ void HttpManager::dump_shard(const Pistache::Rest::Request& request, Pistache::H
     response.send(Pistache::Http::Code::Ok, j.dump());
 }
 
+void HttpManager::trigger_gc(const Pistache::Rest::Request& request, Pistache::Http::ResponseWriter response) {
+    const auto chunk_id_param = request.query().get("chunk_id");
```
yuwmao: This is `p_chunk_id` right? Please add a comment or make it clear to avoid misuse.

JacksonYao287: yes, this is pchunk_id, will add the comments

src/lib/homestore_backend/hs_http_manager.cpp
```diff
@@ -239,6 +243,235 @@ void HttpManager::dump_shard(const Pistache::Rest::Request& request, Pistache::H
...
+}
+
+std::string HttpManager::generate_job_id() {
+    auto counter = job_counter_.fetch_add(1, std::memory_order_relaxed);
+    return fmt::format("trigger-gc-task-{}", counter);
+}
+
+void HttpManager::get_gc_job_status(const Pistache::Rest::Request& request, Pistache::Http::ResponseWriter response) {
+    auto job_id_param = request.query().get("job_id");
+    if (!job_id_param) {
```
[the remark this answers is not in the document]
JacksonYao287: is it reasonable that one client can see the jobs results that submitted by others?
I suggest to postpone this until we really need the function.

src/lib/homestore_backend/hs_http_manager.cpp
```diff
@@ -239,6 +243,235 @@ void HttpManager::dump_shard(const Pistache::Rest::Request& request, Pistache::H
...
+        LOGINFO("GC job {} stopping GC scan timer", job_id);
+        gc_mgr->stop_gc_scan_timer();
+
+        std::vector< pg_id_t > pg_ids;
+        ho_.get_pg_ids(pg_ids);
+        LOGINFO("GC job {} will process {} PGs", job_id, pg_ids.size());
+
+        std::vector< folly::SemiFuture< bool > > gc_task_futures;
+
+        for (const auto& pg_id : pg_ids) {
```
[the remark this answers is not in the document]
JacksonYao287: I have change the api to supporting triggering gc for a specific chunk or gc. 
if we trigger all the chunks, then it will be scheduled pg by pg

src/lib/homestore_backend/hs_http_manager.cpp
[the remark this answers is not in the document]
JacksonYao287: done ,  use the following command to query the status of all gc jobs.
curl -k "https://localhost:10443/api/v1/gc_job_status"

yuwmao (approved):

PR #385, opened 2026-01-20
add blk metrics for gc

JacksonYao287:
add two metrics for gc:
gc_read_blk_count : total read blk count in this pdev issued by gc
gc_write_blk_count : total write blk count in this pdev issued by gc

these two metrics are not persisted, which means they will start from 0 if sm restarts


src/lib/homestore_backend/gc_manager.hpp
```diff
@@ -132,6 +132,8 @@ class GCManager {
                 REGISTER_GAUGE(failed_egc_task_count, "Number of failed emergent gc tasks");
                 REGISTER_GAUGE(total_reclaimed_space_by_gc, "Total reclaimed space by gc task");
                 REGISTER_GAUGE(total_reclaimed_space_by_egc, "Total reclaimed space by emergent gc task");
+                REGISTER_GAUGE(gc_read_blk_count, "Total read blk count by gc in this pdev");
```
xiaoxichen: use REGISTER_COUNTER, then you can use `COUNTER_INC(gc_read_blk_count,  new_pba.blk_count()`

xiaoxichen (approved):

PR #386, opened 2026-01-28
Fix pg membership inconsistency

yuwmao:
- Implement on_pg_clean_replace_member_task callback which actually rollback the membership change
- Add a reconcile_membership API which is used to correct the existing inconsistency problem.

Rely on https://github.com/eBay/HomeStore/pull/856

### DIFF
--- a/src/lib/homestore_backend/hs_pg_manager.cpp
+++ b/src/lib/homestore_backend/hs_pg_manager.cpp
@@ -398,6 +398,108 @@ void HSHomeObject::on_pg_complete_replace_member(group_id_t group_id, const std:
          boost::uuids::to_string(member_out.id), boost::uuids::to_string(member_in.id), tid);
 }
 
+//This function actually perform rollback for replace member task:  Remove in_member, and ensure out_member exists
+void HSHomeObject::on_pg_clean_replace_member_task(group_id_t group_id, const std::string& task_id,
+                                                    const replica_member_info& member_out,
+                                                    const replica_member_info& member_in, trace_id_t tid) {
+    std::unique_lock lck(_pg_lock);
+    for (const auto& iter : _pg_map) {
+        auto& pg = iter.second;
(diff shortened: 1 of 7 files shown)


JacksonYao287 (approved):
LGTM

src/lib/homestore_backend/hs_pg_manager.cpp
```diff
@@ -398,6 +398,108 @@ void HSHomeObject::on_pg_complete_replace_member(group_id_t group_id, const std:
          boost::uuids::to_string(member_out.id), boost::uuids::to_string(member_in.id), tid);
 }
 
+//This function actually perform rollback for replace member task:  Remove in_member, and ensure out_member exists
+void HSHomeObject::on_pg_clean_replace_member_task(group_id_t group_id, const std::string& task_id,
+                                                    const replica_member_info& member_out,
+                                                    const replica_member_info& member_in, trace_id_t tid) {
+    std::unique_lock lck(_pg_lock);
+    for (const auto& iter : _pg_map) {
```
JacksonYao287: nit, suggest using `get_pg_id_with_group_id` instead of a loop here

yuwmao: Thanks, will update it.

src/lib/homestore_backend/hs_pg_manager.cpp
```diff
@@ -398,6 +398,108 @@ void HSHomeObject::on_pg_complete_replace_member(group_id_t group_id, const std:
...
+
+    for (auto& old_member : hs_pg->pg_info_.members) {
+        if (new_members.find(old_member) == new_members.end()) {
+            removed_members.push_back(old_member.id);
+        }
+    }
+
+    for (auto& new_member : new_members) {
+        if (hs_pg->pg_info_.members.find(new_member) == hs_pg->pg_info_.members.end()) {
+            added_members.push_back(new_member.id);
```
Besroy: same as above

PR #395, opened 2026-03-08
[scrubber] phase1: add scrub manager

JacksonYao287:
this pr implements the framwork and basic logic of scrubber, including: 
1 thread model
2 scrubber rpc
3 local scrub: deep and shallow scrub for pg, shard and blob

later PR will :
1 add http interface to trigger pg scrub
2 add more UT to cover more case 
3 do optimization


src/lib/homestore_backend/hs_homeobject_fbs/deep_blob_scrub_map.fbs
```diff
@@ -0,0 +1,31 @@
...
+
+table DeepBlobScrubMap {
+  scrub_info: ScrubInfo;
+  start: uint64;
+  end: uint64;
+  deep_blob_scrub_results: [DeepBlobScrubResultEntry];
+}
+
+// this is used for sending deep scrub result
+root_type DeepBlobScrubMap;
```
xiaoxichen: DeepBlobScrubResp?

xiaoxichen (changes requested):
The FBS part

Flatbuffer handles and skipped non-present fields , avoiding those fields be transmit in the wire,  either "default" or "null" will be returned.   With this, it is more nature to use a flat structure.

For example

```
table BlobScrubResult {
   key: BlobKey;
   code: ResultCode = 0;
    hash:[ubyte] = null;
}
```
It can be used for both shallow and deep, with negligible overhead.


With this pattern, the FBS can be greatly simplificed.

we should only have one request type across board
```
ShardScrubReq {
  scrub_info: ScrubInfo;
  start: uint64;
  end: uint64;
  isDeep: bool;
}
```
also probably one response type ( if not, at least one response type per class(blob/shard/pg))

src/lib/homestore_backend/hs_homeobject_fbs/blob_scrub_req.fbs
```diff
@@ -0,0 +1,13 @@
+include "scrub_common.fbs";
+
+namespace homeobject;
+
+table BlobScrubReq {
+  scrub_info: ScrubInfo;
+  start: uint64;
```
xiaoxichen: uint64 is used for Req however {ShardID, BlobID} is returned  in Response, can you share more idea how this works?   It seems like they need to match each other.

src/lib/homestore_backend/hs_homeobject_fbs/deep_blob_scrub_map.fbs
xiaoxichen: OK, feel free to pick a name but "Map" is a wired name for it.

src/lib/homestore_backend/hs_pg_manager.cpp
```diff
@@ -1115,6 +1125,212 @@ void HSHomeObject::HS_PG::update_membership(const MemberSet& members) {
...
+    }
+
+    auto scrub_mgr = home_obj_.scrub_manager();
+    if (!scrub_mgr) {
+        LOGW("ScrubManager is not initialized in HS_PG::on_scrub_map_received for pg={}", pg_id);
+        return;
+    }
+    scrub_mgr->add_scrub_map(pg_id, scrub_map);
+}
+
```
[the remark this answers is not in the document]
JacksonYao287: the change of FBS part will bring a lot of code change, so let`s first finalize this part.

I agree to your proposal. we can have only one req definition.
```
//new enum to specify scrubtype
enum ScrubType : uint8 {
  PG_META = 0,
  SHALLOW_SHARD = 1,
  DEEP_SHARD = 2,
  SHALLOW_BLOB = 3,
  DEEP_BLOB =4
}

// add scrub_type into ScrubInfo
table ScrubInfo {
  pg_id: uint16;
  task_id: uint64;
  req_id: uint64;
  scrub_lsn: int64;
  scrub_type: ScrubType;
  issuer_uuid : [ubyte];
}

// after change , we have only one scrub req
// 1 start and end can be omitted of pg_meta scrub.
// 2 start and end will be the shard_id for shard_scrub and blob_id for blob scrub.
table ScrubReq {
  scrub_info: ScrubInfo;
  start: uint64;
  end: uint64;
}
```

for scrub map
```
//for shallow blob scrub, ResultCode and hash will be omitted.
table BlobScrubResult {
   key: BlobKey;
   code: ResultCode = 0;
    hash:[ubyte] = null;
}

//only one blob scrub map definition for both deep and shallow scrub.
table blobScrubMap {
  scrub_info: ScrubInfo;
  start: uint64;
  end: uint64;
  blob_scrub_results: [BlobScrubResult];
}

// shard, for shllow shard scrub, ResultCode will be omitted.
table ShardScrubResult {
   shard_id: uint64;
   code: ResultCode = 0;
}

table ShardScrubMap {
  scrub_info: ScrubInfo;
  start: uint64;
  end: uint64;
  shard_scrub_results: [ShardScrubResult];
}

table PGMetaScrubResult { 
scrub_info: ScrubInfo;
pg_meta_scrub_result: ResultCode
}
```

@xiaoxichen  pls first check this proposal, and left your comments

src/lib/homestore_backend/scrub_manager.hpp
```diff
@@ -0,0 +1,480 @@
...
+    inline static auto const pg_scrub_meta_name = std::string("PG_SCRUB");
+    static constexpr uint64_t blob_max_hash_len = 32;
+    using BlobHashArray = std::array< uint8_t, blob_max_hash_len >;
+    using chunk_id_t = homestore::chunk_num_t;
+    // TODO: persist this into metablk.
+    inline static atomic_uint64_t scrub_task_id{1};
+
+    // pg scrub superblk
+#pragma pack(1)
+    struct pg_scrub_superblk {
```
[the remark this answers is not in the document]
JacksonYao287: I suggest not to record the progress of a scrub task just like ceph,  a new scrub stask will be scheduled. 
in homeobject, a task will be cancelled if any error happens when scrubbing. for example , leader change.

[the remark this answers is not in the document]
xiaoxichen: I am not sure if ceph is a good example as we dont have a successful story  to do deep-scrub once a month, on 4TB drives, now we are running at 20TB drives...

Anyway, fine to postpone it till necessary.

src/lib/homestore_backend/scrub_manager.cpp
```diff
@@ -0,0 +1,2021 @@
...
+    return shard_srcub_map;
+}
+
+folly::SemiFuture< std::shared_ptr< ScrubManager::ShallowScrubReport > >
+ScrubManager::submit_scrub_task(const pg_id_t& pg_id, const bool is_deep, const bool force,
+                                SCRUB_TRIGGER_TYPE trigger_type) {
+    LOGINFOMOD(scrubmgr, "submit a scrub task for pg={}, deep_scrub:{}", pg_id, is_deep);
+
+    // Check if a scrub task is already running for this PG
+    // Note: There's still a small race window between this check and task execution in handle_pg_scrub_task,
```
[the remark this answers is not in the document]
xiaoxichen: Feel free to READ any PG state,  but , need strong reason to WRITE an PG state as it is external visible.  

I would prefer we change it now by moving the setting of SCRUBBING to the start of scrubbing task. This ensures external correctness.  Then,  with the assumption that we dont have automatically pg scrub scheduling,  the racing window is small enough due to 
1.  the Q is expected to be short.
2.  only human can trigger GC via api.

With the reasoning,  we can add a fixme,  and address the de-dedup question in next PR.

src/lib/homestore_backend/hs_homeobject_fbs/blob_scrub_req.fbs
JacksonYao287: > typical CPU core can do 1M *O(1) per second

I doubt this, typical cpu core should be much faster than this in my mind. following is the answer from gpt
```
For a “typical” modern CPU core, a plain O(1) uint64 comparison (a < b, a == b) can be done on the order of hundreds of millions to a few billions per second—not just 1 million

A practical rule of thumb (single core, ~3 GHz):

Best case (register-to-register compare, tight loop, branchless):
~1–3 × 10⁹ comparisons/sec/core (billions/sec)

Common case (streaming through arrays, mostly L1/L2 hits):
~3 × 10⁸ to 1 × 10⁹ comparisons/sec/core

Worst case (random memory access, frequent cache misses):
can drop to ~1 × 10⁷ to 1 × 10⁸ comparisons/sec/core (tens of millions/sec), because you’re waiting on memory, not the compare instruction.

So the compare itself is extremely cheap; performance is usually dominated by memory access, branching/predictability, and loop overhead.
```
>we limit each round trip within one shard

1 in the worst case , if there is only one blob in each shard, we will have a huge number of round-trips for 128Million blob. so if there are too many small shard, we will have to schedule too many round trip.

2 if one shard is lost(shard meta blk is lost) on leader , all the blobs in this shard will not be scrubbed.


actually , this is why  I proposed the idea of scrubbing the entire pg  vchunk by vchunk in my earlier design.  this will solve all the problems we met here. a vchunk can limit the amount of shard and blobs to a acceptable batch size in a single scrub req. and the number of round trip for a pg scrubbing is fixed since the number of vchunks in a pg is fixed. 

the only concern of vchunk scrubbing is ramdom io on hdd for each single blob. 
```
2GB data in a chunk, 4K (min blob size) random read, 7200 RPM HDD ~200 IOPS
    // Total read operations: 2GB / 4KB = 524,288 reads
    // Estimated time: 524,288 / 200 = 5,243 seconds ≈ 44minutes
```
if we sort all blobs in the vchunk  by pba to mimic a sequentially read, seems this will significantly reduce the time consuming?  do we need to change back to vchunk based scrubbing?

xiaoxichen: I dont want to argue with GPT as the btree access pattern is far different from pure memory access as well as there are locking invloved .   You can have a POC to prove for a 128Million record Btree how long you can traverse it. We can based on the number to discuss.

1.  Yes it is super slow but no CPU/MEM/IO pressure.
2. we check shard consistency before going into blob consistency check --- if there is a shard missing we knows there.
3.  I am fine to use vchunk based scrubbing however it is fully depends on how we break one task (full vchunk) into smaller batches on follower side .  We cannot submit 256K IOs,  even they are sequential , to the disk at once,  letting the disk read at 200MB/s for 10s that will add up latency for other request , especially write,  by 10s.   A reasonable batch might be 4MB/8MB(20-40ms added) with proper sleep in between. 

So the bound we are solving are:

-  Maximum batch size (regardless it is shard/vchunk) 256K blobs (8K per blob).
-  Maximum shard count , literary infinity after we remove shard_header from disk, reasonable upper limit: 256K.

Combined both , we need   1) split huge task(shard/vchunk) into small unit to avoid impacting foreground traffic.  2) similar to 1), but reasonable execute time of a task to allow timeout-based retry. 3) effective consolidate K shards in single RT, for less message.  

Tools we have:
1.  the flatbuffer is efficient for default value --- we can have {shardid,  blobid_low, blobiid_high} as a task unit.
2. we can carry a list of task units in single request 
3. Potentially if we let leader scrub first, it knows the size of each blob so it can controls the size per Req, to the best extend.

JacksonYao287:
1 according to the adr, a lot of code changes. 
2 the target branch has been moved to stable/v4.x

so we close this PR , and use #413 instead

PR #405, opened 2026-04-08
SDSTOR-21763 Fix GC job status update for multi-PG scenarios

JacksonYao287:
Refactor trigger_gc_for_pg to return folly::Future and defer job status updates to callers. Previously, when triggering GC for all PGs, the job status was incorrectly set to COMPLETED after the first PG finished. Now the job status is only set to COMPLETED after all PGs have completed their GC operations.

Changes:
 - trigger_gc_for_pg now returns folly::Future<folly::Unit>
 - Removed set_job_status parameter and internal status updates
 - Callers use thenValue to set job status after completion
 - Single PG GC: sets status after that PG completes
 - Multi PG GC: collects all futures and sets status after all complete

### DIFF
--- a/src/lib/homestore_backend/hs_http_manager.cpp
+++ b/src/lib/homestore_backend/hs_http_manager.cpp
@@ -608,23 +608,26 @@ void HttpManager::trigger_gc(const Pistache::Rest::Request& request, Pistache::H
         std::vector< pg_id_t > pg_ids;
         ho_.get_pg_ids(pg_ids);
 
-        if (pg_ids.empty()) {
-            result["error"] = "no PG found";
-            response.send(Pistache::Http::Code::Not_Found, result.dump());
-            return;
-        }
-
         const auto job_id = generate_job_id();
         result["job_id"] = job_id;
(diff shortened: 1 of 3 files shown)


src/lib/homestore_backend/hs_http_manager.cpp
```diff
@@ -595,14 +595,25 @@ void HttpManager::trigger_gc(const Pistache::Rest::Request& request, Pistache::H
...
+        });
     } else {
         LOGINFO("Received trigger_gc request for all chunks");
         nlohmann::json result;
+        std::vector< pg_id_t > pg_ids;
+        ho_.get_pg_ids(pg_ids);
+
+        if (pg_ids.empty()) {
+            result["error"] = "no PG found";
+            response.send(Pistache::Http::Code::Not_Found, result.dump());
```
Besroy: Can we set completed (no pgs need to be gc = all needed pgs gced)

JacksonYao287: done

Besroy (approved):
LGTM

PR #407, opened 2026-04-10
SDSTOR-21797 Propose scrub flatbuffers schema and blob scrub design adr

JacksonYao287:


docs/adr/scrub-blob-range-coverage.md
```diff
@@ -0,0 +1,197 @@
...
+
+## Alternatives Considered
+
+### Option A: Static blob_id Equal-Width Slicing (original PR approach)
+
+Slice `[0, last_blob_id)` into fixed-width intervals of size N. The leader issues them in order and advances to the next interval once all followers respond.
+
+**Problems:**
+
+1. **Sparsity causes empty RTTs.** In low-write-frequency workloads, consecutive blob IDs may differ by millions. The number of RTTs is `max_blob_id / N`, not `blob_count / N` — the gap can be 10x to 100x.
```
xiaoxichen: it is not low-write-requency workload, it is delete intensive workload.

docs/adr/scrub-blob-range-coverage.md
```diff
@@ -0,0 +1,197 @@
...
+---
+
+### Option B: Follower-Driven Scanning
+
+The leader sends `last_shard_id` and `last_blob_id` to all followers once. Each follower independently scans from `{0, 0}` to `{last_shard_id, last_blob_id}`, reporting every N blobs. The leader passively receives results.
+
+**Rejected for the following reasons:**
+
+1. **No progress alignment.** Two followers are in different blob ranges at the same time. The leader cannot release memory for any range until it has confirmed all replicas have finished that range — effectively degenerating into full buffering.
+2. **Message loss is ambiguous.** If a follower's report for a segment is dropped, the leader cannot distinguish "no blobs in that range" from "the result message was lost." This fault-tolerance edge case is easy to implement incorrectly.
```
xiaoxichen: is hard to be fully covered.

docs/adr/scrub-blob-range-coverage.md
```diff
@@ -0,0 +1,197 @@
...
+
+After shard scrub completes, the leader takes the **maximum blob_count across all replicas** for each shard and builds:
+
+```cpp
+std::map<uint64_t /*shard_id*/, uint32_t /*max_blob_count*/> shard_blob_cnt;
+```
+
+**Key property**: This count comes from the most populated replica, not the leader itself. Even if the leader has lost blobs in a shard, it uses the follower's larger count to estimate that shard's size.
+
+Additionally, blob scrub only begins if shard scrub completes with **no missing shards**. If any missing shard is detected during shard scrub, the task terminates immediately and returns the shard-level findings — blob scrub is not attempted. By the time blob scrub begins, `shard_blob_cnt` therefore contains only shards confirmed to exist on all replicas.
```
xiaoxichen: >blob scrub only begins if shard scrub completes with **no missing shards**

Clarification is needed,  if a member F, lost a shard metadata, but the data blobs exists , why dont we scrub it?  The Blob read path doesnt need involving shard metadata.

docs/adr/scrub-blob-range-coverage.md
```diff
@@ -0,0 +1,197 @@
...
+
+if current_batch is not empty:
+    batch_list.append(current_batch)
+```
+
+This provides two important guarantees:
+
+- **Small-shard batching**: Many small shards (e.g., each holding 1–2 blobs) are packed into a single batch and covered in one RPC. RTT count = `Σ cnt / BLOB_BATCH_SIZE`, proportional to the actual blob count, not the blob_id range.
+
+- **Large-shard isolation**: When a single shard's blob count exceeds `BLOB_BATCH_SIZE`, it is placed alone in its own batch. A single shard holds at most 256K blobs; at ~20 bytes per entry, three replicas require ~15 MB — small enough for the leader to hold all results in memory and compare them directly, with no follower paging needed.
```
xiaoxichen: max_chunk_size = 4GB(due to chunksize is 32bit)? so technically it can bd 512K blobs... please include the calculation.

docs/adr/scrub-blob-range-coverage.md
```diff
@@ -0,0 +1,197 @@
...
+---
+
+## Correctness Analysis
+
+### Full Coverage
+
+Blob scrub only runs when shard scrub has completed with no missing shards (see Step 1). Given that precondition, for any committed blob `<shard_id, blob_id>`:
+
+- Its shard is present on all replicas — guaranteed by the shard scrub gate.
+- `shard_blob_cnt[shard_id]` holds the blob count from the most populated replica, which is ≥ 1, ensuring the shard is never skipped. we will have delete shard, so there will not be too much empty shards in a single batch.
```
xiaoxichen: We dont have it now,  lets add `max(shard_blob_cnt[shard_id], 1)` when calculating the scrub batch to release the concern.

docs/adr/scrub-blob-range-coverage.md
```diff
@@ -0,0 +1,197 @@
...
+- The batch range for that shard is `[UINT64_MIN, last_blob_id]`; the blob_id falls within this range by definition.
```
xiaoxichen: lets use 0 instead of UINT64_MIN....

docs/adr/scrub-blob-range-coverage.md
```diff
@@ -0,0 +1,197 @@
...
+- For large shards, the full range `[UINT64_MIN, UINT64_MAX]` is issued in one batch; no early truncation is possible.
```
xiaoxichen: ??? WHAT IS THE UPPER BOUND??

docs/adr/scrub-blob-range-coverage.md
xiaoxichen: >with no follower paging needed.

How do we reach this conclusion?  A shard can be as big as a chunk, and if there is no much errors the shard is expected to be chunk-size.  The 2GB/4GB amount of READ IO is concerning.

docs/adr/scrub-flatbuffers-schema-design.md
```diff
@@ -0,0 +1,219 @@
...
+**Date**: 2026-04-10  
+**Deciders**: Jie  
+
+---
+
+## Context
+
+The scrub manager needs to serialize two categories of messages over the wire between the leader and replicas:
+
+1. **ScrubReq** — leader broadcasts a scrub request to all replicas, telling them what range to check and at what LSN.
```
xiaoxichen: Not sure if it is `broadcasts`, based on the review of previous doc.  I believe it is gossip to particular follower

docs/adr/scrub-flatbuffers-schema-design.md
```diff
@@ -0,0 +1,219 @@
...
+  shard_id:     uint64;
+  blob_id:      uint64;
+  scrub_result: ScrubResult;  // default = NONE → absent for healthy blobs (saves space)
+  hash:         [ubyte];      // absent for shallow scrub or corrupted blobs
+}
+table BlobScrubMap {
+  scrub_info:         ScrubInfo;
+  blob_scrub_results: [BlobScrubResultEntry];
+}
+
```
xiaoxichen, line 103: Can you have an option of 
```
table ScrubResultEntry {
  shard_id:     uint64;
  blob_id:      uint64;
  scrub_result: ScrubResult;  // default = NONE → absent for healthy blobs (saves space)
  hash:         [ubyte];      // absent for shallow scrub or corrupted blobs
}


// use single scrub map for Blob/Shard/PG
table ScrubMap {
    reqID:         uint64;
    type:           ScrubType;
    scrub_result: [ScrubResultEntry]
}
```
For shard , the blobid field left empty
For PG, both shard_id/blob_id left empty.

docs/adr/scrub-flatbuffers-schema-design.md
```diff
@@ -0,0 +1,219 @@
...
+// blob_scrub_map.fbs
+table BlobScrubResultEntry {
+  shard_id:     uint64;
+  blob_id:      uint64;
+  scrub_result: ScrubResult;  // default = NONE → absent for healthy blobs (saves space)
+  hash:         [ubyte];      // absent for shallow scrub or corrupted blobs
+}
+table BlobScrubMap {
+  scrub_info:         ScrubInfo;
```
xiaoxichen: how is scrub_info useful in response? 
if only a few fields useful, can we just taking those fields?

docs/adr/scrub-flatbuffers-schema-design.md
```diff
@@ -0,0 +1,219 @@
...
+
+// scrub_req.fbs
+table ScrubReq { scrub_info: ScrubInfo; }
+
+// blob_scrub_map.fbs
+table BlobScrubResultEntry {
+  shard_id:     uint64;
+  blob_id:      uint64;
+  scrub_result: ScrubResult;  // default = NONE → absent for healthy blobs (saves space)
+  hash:         [ubyte];      // absent for shallow scrub or corrupted blobs
```
xiaoxichen: what is the hash algorithm we are targeting? 
IIRC I did a research and seems like `CRC64NVME` is the best candidate, combining both computational expense as well as strengthen.

If it is CRC32/CRC64NVME, then we can use uint64 for it, avoiding the byte arrary.

docs/adr/scrub-flatbuffers-schema-design.md
```diff
@@ -0,0 +1,219 @@
...
+
+Having separate `BlobScrubMap`, `ShardScrubMap`, and `PGMetaScrubMap` result tables (rather than a single unified result table or a union) achieves type safety at the granularity level that matters: the transport layer tags each result message with its type via a 1-byte `ScrubType` prefix sent alongside the FlatBuffer, and the receiver deserializes directly into the right table type. This avoids union overhead while keeping the number of tables bounded at 3 (not 10 as in Option A).
+
+### 5. Shared `ScrubInfo` is a single point of change for correlation fields
+
+`pg_id`, `task_id`, `req_id`, `scrub_lsn`, and `issuer_uuid` are correlation identifiers needed by every scrub message for deduplication and request-response matching (see `ScrubManager::PGScrubContext::add_scrub_map` → `bsm->match(current_req)`). Embedding `ScrubInfo` in every table means these fields are changed in exactly one place in the schema when the correlation protocol evolves.
+
+---
+
+## Consequences
```
xiaoxichen: *Schema evolution* how this related with the major discussion point of this ADR?  
*Absent-field risk* :  Can this be mitigated by setting the default value of last_obj_id to MAX_UINT64?

docs/adr/scrub-flatbuffers-schema-design.md
```diff
@@ -0,0 +1,219 @@
...
+
+table ScrubInfo {
+  pg_id:          uint16;
+  task_id:        uint64;
+  req_id:         uint64;
+  scrub_lsn:      int64;
+  scrub_type:     ScrubType;      // discriminator; default = PG_META (0)
+  start_shard_id: uint64;         // absent → 0 for non-shard types (harmless)
+  end_shard_id:   uint64;         // absent → 0 for non-shard types (harmless)
+  last_blob_id:   uint64;         // absent → 0 for non-blob types  (harmless)
```
xiaoxichen: depending on the discussion of previous design,  last_blob_id ? or <start_blob_id, end_blob_id> is subjuect to change

docs/adr/scrub-blob-range-coverage.md
JacksonYao287: the blob count of a shard is returned with the shard id in shard scrub result. if a shard is missing on one replica, it will not return the shard as well as the blob count of this shard. as a result , we don`t know the max_blob_count of this shard, since it is got by comparing the shard scrub result from the tree replicas.

another solution is we scrub missing shard in a batch , which contains only this single shard?

xiaoxichen: yes obviously we can choose to do one missing shard per batch.

docs/adr/scrub-blob-range-coverage.md
JacksonYao287: we can add a 20ms sleep after scrub a cumulated blob size of 16M to make sure not to take all the io resource, I will add this to the doc

xiaoxichen: reasoning is needed.  20ms is a magic number,   how long it will take to complete 16MB read,  with 20ms sleep that means what portion of IO given to scrubbing  ? and how long the impact window would be (with previous assumption, how long will it take to finish a 2GB shard),  then translate to what would be the propriate timeout value on leader side ? 

If we can scrub a shard that is full-chunk-size without slicing/paging,  then we can do the scrubbing on chunk level, which is much simpler than this approach. We are circling  here for several times.

docs/adr/scrub-flatbuffers-schema-design.md
JacksonYao287: last_blob_id.  we don`t need <start_blob_id, end_blob_iwd>.

when scrubbing blobs, assuming the shard batch is [start_shard_id, end_shard_id], 
 the range of index range_query will be [{start_shard_id, 0}, {end_shard_id, last_blob_id}]

docs/adr/scrub-flatbuffers-schema-design.md
JacksonYao287: for now, [the hash algorithm we are using is CRC32](https://github.com/eBay/HomeObject/blob/3fd3bdad151173a46e6fd9bec82b7afb8a499705/src/lib/homestore_backend/hs_blob_manager.cpp#L157)

however, we support multiple hash algorithms now, and [the blob_max_hash_len is 32,](https://github.com/eBay/HomeObject/blob/3fd3bdad151173a46e6fd9bec82b7afb8a499705/src/lib/homestore_backend/hs_homeobject.hpp#L436).

to keep compatible if algorithm changes, we have to use byte arrary, even if we know it is 32bit(4B) now

docs/adr/scrub-flatbuffers-schema-design.md
JacksonYao287: almost all the fileds are used to check if the BlobScrubMap is the one I want.  that is , if it matches the current req of leader 
https://github.com/eBay/HomeObject/pull/395/changes#diff-a2f2f7f16456c62c69e3a806813dbbe974551df7c32fdce849596b10dadd215cR143-R144

JacksonYao287: if sm restart, req_id will start from 0. then it probably received scrub map from a stale scrub req, which has the same req id.

docs/adr/scrub-flatbuffers-schema-design.md
JacksonYao287: for shard, I want  ScrubResultEntry contains shard_id and blob count, 
for pg, I want ScrubResultEntry contains the shard count and blob count of this pg.

so , they will not be empty.

```
table ScrubResultEntry {
  key_1:     uint64;
  key_2:      uint64;
  scrub_result: ScrubResult;  // default = NONE → absent for healthy blobs (saves space)
  hash:         [ubyte];      // absent for shallow scrub or corrupted blobs
}

table ScrubMap {
    scrub_info:         ScrubInfo;
    scrub_result: [ScrubResultEntry]
}
```

this is a more generic definition.
if scrub type is deep_blob/shallow blob,  key_1 is shard_id, key_2 is blob_i2
if scrub type is depp_shard/shallow_shard, key_1 is shard_id, key_2 is blob count
if scrub type is pg_meta, key_1 is total_shard_count, key_2 is total_blob_count.

 total_shard_count and  total_blob_count be useful to compare the replica at a hight level

xiaoxichen: How do you expect the blob_count on PG level be used? If we can keep one result per entry it is more straightforward.

Also, maybe we can change ScrubResult to int64 , use negative for error?  and positive for value?

docs/adr/scrub-flatbuffers-schema-design.md
xiaoxichen: I think your TO-DO clearly indicating the right path --- now a lot of fields used trying to make a request uniq, however, it can be achieved by  a randomized requestID,  uuid, etc.

docs/adr/scrub-flatbuffers-schema-design.md
xiaoxichen: the blob hash we stored on disk, doesnt necessary to match the hash we used for scrubbing,  for example if we decide to use SHA256 for scrubbing , we can do that by sending back SHA256 of each blob in BlobMap.  Or reverse, if  data stored locally use MD5, we can still use CRC32 for scrubbing. 

I think this isolation should be part of the scrubber design , for example if we change the blob hashing algorithm on PutBlob side, there will be a mixture of blobs with hash_algo1 and hash_algo2.  Even worse, due to rolling upgrade of different FD,  a single blob's checksum algorithm, might not be the same across replicas.

Shortcut can be taken if blob hashing algo match with scrubbing algo, but this should be an optimization , althouhg the optimization works for 99% cases.

docs/adr/scrub-blob-range-coverage.md
```diff
@@ -0,0 +1,223 @@
...
+- Resulting duty cycle: 40 / (40 + 240) ≈ **14.3%**, leaving **85.7%** of HDD time to foreground I/O. These values will be tuned based on production iostat data.
+
+**Impact window for a large shard:**
+
+| Shard size | Effective scrub rate | Estimated scrub time |
+|---|---|---|
+| 2 GB | ~14 MB/s | ~146 s |
+| 4 GB (full chunk) | ~14 MB/s | ~293 s |
+
+The leader-side RPC timeout for a single blob scrub batch must exceed the worst-case time for a full chunk-sized shard. With a 4 GB shard taking ~293 s, the recommended leader timeout floor is **≥ 600 s** (approximately 2× worst-case to account for I/O contention variability).
```
xiaoxichen: how will the 300s or 600s timeout impacting us?

JacksonYao287: the impact is the retry after timeout will be too late if timeout is too large.

for example, is timeout is 10s, if not receive any response, we consider timeout and time out retry immetiately.
if  600s, we only can know the timeout after 600s, retry and wait for another 600s. this will too time consuming.

taking small steps and moving fast is better and wait a result for too long .

I have update the doc with the IOPS Throttling analysis . let`s first discuss and try to finalize scrub-blob-range-converage adr.

docs/adr/scrub-flatbuffers-schema-design.md
JacksonYao287: I don`t have a clear for this so far. I think they will useful in some scenario in the future and it is transfered only once(pg_meta scrub is executed only once) and not expensive, so I put here.

>positive for value?
the value stands for what for a single blob?

docs/adr/scrub-flatbuffers-schema-design.md
JacksonYao287: requestID LGTM

docs/adr/scrub-flatbuffers-schema-design.md
JacksonYao287: LGTM，let`s use uint32 here for hash instead of  byte array。 if we change the hash algorithm for put_blob in the future, we can use a seperate PR to  add a crc32 as the hash algorithm dedicated for scrubbing.

docs/adr/scrub-flatbuffers-schema-design.md
xiaoxichen: If doesn't impact single blob
NONE: 0
ERROR: NEGTIVE 
POSITIVE: APP_defined.

docs/adr/scrub-flatbuffers-schema-design.md
xiaoxichen: It is a separate discussion/reasoning regarding what would be the best hashing algorithm for scrubbing purpose.

Suggeting uint64 based on my previous analyze as it suggest to CRC64NVM

docs/adr/scrub-blob-range-coverage.md
xiaoxichen: Yes agree on the timeout impact, then what is the solution

JacksonYao287: pls see the updated adr, no long timeout will be involved in the scrubbing process. we never scrub a large shard in a single rpc.

docs/adr/scrub-blob-range-coverage.md
```diff
@@ -0,0 +1,294 @@
...
+
+```cpp
+std::map<shard_id_t, uint64_t> shard_blob_count;
+```
+
+Two special cases apply:
+
+- **Empty shard** (`blob_count == 0`): use `max(1, blob_count) = 1`. This ensures empty shards still contribute to the batch size accumulator and each batch eventually seals, bounding batch count.
+
+- **Missing shard** (exists on some replicas but not all): set `blob_count = UINT64_MAX`. This forces the missing shard into its own isolated batch (see Step 2). Without a reliable blob count from cross-replica comparison, the only safe assumption is worst-case coverage.
```
xiaoxichen: the UINT64_MAX  makes the. `current_batch_count + effective_count > MAX_BLOB_COUNT:` overflow.

Suggest limiting the MAX_BLOB_COUNT range within uint32 (but underlying use uint64) which might avoid the overflow concerns.

docs/adr/scrub-blob-range-coverage.md
```diff
@@ -0,0 +1,294 @@
...
+2. Each batch covers `[{start_shard_id, 0}, {end_shard_id, last_blob_id}]`. The follower's index range query with a `shard_id` prefix ensures every blob in this range is visited.
+3. The streaming protocol continues until the follower returns empty — meaning the full range was exhausted.
+
+**Conclusion**: Every committed blob is covered.
+
+### Safety Under Leader Data Loss
+
+If the leader has lost blobs in shard S:
+- Its local blob count for S is too low.
+- But `shard_blob_count[S]` uses the **max across replicas**, so the batch range includes S with the correct (follower-derived) upper bound.
```
xiaoxichen, line 214: it has nothing to do with 

```
- Its local blob count for S is too low.
- But `shard_blob_count[S]` uses the **max across replicas**, so the batch range includes S with the correct (follower-derived) upper bound.
```

docs/adr/scrub-blob-range-coverage.md
```diff
@@ -0,0 +1,294 @@
...
+{ start_shard_id, end_shard_id, last_blob_id }
+```
+
+`last_blob_id` is read once at scrub start and used as the scan ceiling throughout. Raft guarantees a follower's committed log never exceeds the leader's, so no per-replica negotiation is needed.
+
+**Follower streaming protocol:**
+
+1. The follower executes an index range query over `[{start_shard_id, 0}, {end_shard_id, last_blob_id}]`, leveraging the `{shard_id, blob_id}` composite key for a direct seek to `start_shard_id`. Leader do it also in the backgroud when waiting for followers' responses. The follower streams results back to the leader in batches of at most `MAX_BLOB_COUNT_PER_RPC` blobs (configurable, reference: 10 blobs per batch).
+2. After reading every MAX_BLOB_COUNT_PER_RPC blobs (configurable), the follower returns the scrub results for those MAX_BLOB_COUNT_PER_RPC blobs to the leader. Suppose the last blob in this batch is `{shard_id_n, blob_id_n}`; the leader then issues the next sub-request starting at `[{shard_id_n, blob_id_n + 1}, {end_shard_id, last_blob_id}]`.
+3. The follower executes a fresh index range query from `{shard_id_n, blob_id_n + 1}` and returns the next batch of results.
```
xiaoxichen, line 150: Please distinguish ( and [ when indicating a range and thinking about duplication. 

`({start_shard, start_blob}` vs `[{start_shard, start_blob}`

docs/adr/scrub-blob-range-coverage.md
```diff
@@ -0,0 +1,294 @@
...
+---
+
+## Efficiency Analysis
+
+| Scenario | Option A (current) | Proposed design |
+|---|---|---|
+| Delete-intensive (blob_id sparse, few live blobs) | RTTs ≈ max\_blob\_id / 1000; can be 100× too many | RTTs ≈ Σ blob\_count / 1000; tracks actual data |
+| Many small shards (1–2 blobs each) | Full index scan per interval; no locality | Multiple shards per RPC; shard-level seek |
+| Single large shard (up to 524,288 blobs) | Full index scan across all shard boundaries | Solo batch; direct seek; streaming keeps per-RPC latency bounded |
+| Leader has lost blobs in a shard | Range too narrow; follower blobs silently skipped | Max-across-replicas count corrects the range |
```
xiaoxichen: again the range has nothing to do with count

docs/adr/scrub-blob-range-coverage.md
```diff
@@ -0,0 +1,294 @@
...
+| Missing shard | Leader's shard list missing the shard entirely | UINT64\_MAX count → solo batch → covered |
+
+---
+
+## Trade-offs and Limitations
+
+1. **shard_blob_count is a snapshot, not a live count.** New writes between shard scrub and blob scrub increase actual blob counts above the estimate. Those blobs have IDs above `last_blob_id` and are outside this round's scope — correctness is unaffected.
```
xiaoxichen: the upper limit of `last_blob_id` excludes those newly written blobs from the scrub

docs/adr/scrub-blob-range-coverage.md
```diff
@@ -0,0 +1,294 @@
...
+
+## Consequences
+
+**Benefits**
+
+- RTT count is proportional to actual blob count, not to `max_blob_id`. Delete-intensive workloads do not generate wasteful empty RTTs.
+- Index access uses `shard_id`-prefix seeks, eliminating full-tree scans for each interval.
+- Coverage is safe even when the leader has lost data — the max-across-replicas count drives the range.
+- Missing shards are covered conservatively without blocking scrub of healthy shards.
+- Per-RPC latency is bounded by the 1,000-blob streaming batch, independent of shard size.
```
xiaoxichen: use `MAX_BLOB_COUNT_PER_RPC` instead of magic number (1000), as in your previous section the default is `10`.

docs/adr/scrub-blob-range-coverage.md
```diff
@@ -0,0 +1,294 @@
...
+- Leader peak memory is bounded at ≈ 77.1 MB, regardless of PG size.
+
+**Costs**
+
+- Implementation is coupled to the shard scrub phase; `shard_blob_count` must be available before blob scrub begins.
+- Streaming sub-response protocol adds implementation complexity compared to a single bulk response per range.
+- The missing-shard worst-case path generates more RPCs than necessary if the missing shard's actual blob range is small. This is intentional: correctness takes priority over efficiency in a degraded state.
```
xiaoxichen: why? 
assuming the missing blob has only one blob with ID `last_blob_id`,  isnt we finish the scrubbing within one RPC?

xiaoxichen (commented):
skipped review for the flatbuffer review, only the blob-range as the later one depends on the first one.

docs/adr/scrub-blob-range-coverage.md
JacksonYao287: yes, I have changed this, MAX_BLOB_COUNT is enough

docs/adr/scrub-blob-range-coverage.md
JacksonYao287: removed, thanks

docs/adr/scrub-blob-range-coverage.md
JacksonYao287: thanks, I forgot upating this part

docs/adr/scrub-blob-range-coverage.md
JacksonYao287: you are right, I missed the update of this part.

docs/adr/scrub-blob-range-coverage.md
JacksonYao287: ack

docs/adr/scrub-flatbuffers-schema-design.md
JacksonYao287: sure

docs/adr/scrub-flatbuffers-schema-design.md
JacksonYao287: How do you expect APP_defined be used?  or just keep it for later extension?

xiaoxichen: For shard it is blob_count
for pg it is shard count (optionally if we really want blob count,  hash field can be used)
for blob,  reserve for later extension

docs/adr/scrub-flatbuffers-schema-design.md
xiaoxichen: seems we need start_blob_id now :)

JacksonYao287: let`s first discuss and try to finalize scrub-blob-range design adr, and then circle back on this. I believe we will be more clear what we exactly want then.

JacksonYao287: @xiaoxichen  pls review the flatbuffer adr. I have updated it according to our discussion yesterday

docs/adr/scrub-flatbuffers-schema-design.md
```diff
@@ -0,0 +1,283 @@
...
+///   DEEP_BLOB    → shard_id identifies the shard, value holds blob_id
+///
+/// scrub_result defaults to NONE (0) and is elided for healthy entries.
+/// hash is present only for DEEP_BLOB entries whose scrub_result == NONE;
+/// absent (→ 0) for META, SHALLOW_BLOB, and for any corrupted blob.
+table ScrubResultEntry {
+  shard_id:        uint64;
+  value:        uint64;
+  scrub_result: ScrubStatus;  // default NONE → elided for healthy entries (saves 1 B each)
+  hash:         uint64;       // CRC64; absent unless DEEP_BLOB and scrub_result == NONE
```
xiaoxichen: “present only when" is easier to read compared to “absent unless”

docs/adr/scrub-flatbuffers-schema-design.md
```diff
@@ -0,0 +1,283 @@
...
+///   META         → shard_id identifies the shard, value holds blob_count
+///   SHALLOW_BLOB → shard_id identifies the shard, value holds blob_id
+///   DEEP_BLOB    → shard_id identifies the shard, value holds blob_id
+///
+/// scrub_result defaults to NONE (0) and is elided for healthy entries.
+/// hash is present only for DEEP_BLOB entries whose scrub_result == NONE;
+/// absent (→ 0) for META, SHALLOW_BLOB, and for any corrupted blob.
+table ScrubResultEntry {
+  shard_id:        uint64;
+  value:        uint64;
```
xiaoxichen: as default field do not transfer, why dont we add blobk_id ?

docs/adr/scrub-flatbuffers-schema-design.md
```diff
@@ -0,0 +1,283 @@
...
+  scrub_results: [ScrubResultEntry];
+}
+root_type ScrubResult;
+```
+
+### shard_id / value / hash semantics by ScrubType
+
+| `scrub_type` | `shard_id` | `value` | `hash` |
+|---|---|---|---|
+| `META` | `shard_id` | `blob_count` | absent |
```
xiaoxichen: dont we read out the shard metablk and verify?

docs/adr/scrub-flatbuffers-schema-design.md
```diff
@@ -0,0 +1,283 @@
...
+| `SHALLOW_BLOB` | `shard_id` | `blob_id` | absent |
+| `DEEP_BLOB` | `shard_id` | `blob_id` | CRC64 (absent on error) |
+
+`META` produces one entry per shard in the requested range. The leader computes PG-level aggregates (total shard count, total blob count) by counting entries and summing `value` values — no separate summary entry is needed.
+
+### ScrubReq field applicability
+
+| `scrub_type` | `start_shard_id` | `start_blob_id` | `end_shard_id` | `last_blob_id` |
+|---|---|---|---|---|
+| `META` | set | absent → 0 | set | absent → 0 |
```
xiaoxichen: what is the decision regarding batch in META?

docs/adr/scrub-flatbuffers-schema-design.md
```diff
@@ -0,0 +1,283 @@
...
+
+### 5. `uint64` hash instead of `[ubyte]`
+
+Using a fixed-width `uint64` rather than `[ubyte]` removes the 4-byte vector-length prefix overhead and avoids a heap allocation on deserialization. CRC64/NVME fits in 8 bytes and offers stronger collision resistance than CRC32 for scrub purposes. If the hashing algorithm changes in the future, the field width can be revisited in a single schema change without altering the shape of `ScrubResultEntry`.
+
+### 6. `req_id` for epoch deduplication; `issuer_uuid` for sender identity
+
+The two fields serve distinct purposes and cannot replace each other:
+
+- **`req_id: uint64`** — a randomly generated per-sub-request identifier. The leader matches incoming `ScrubResult` messages to the outstanding request and discards any response whose `req_id` does not match, eliminating stale responses from a previous scrub epoch. It replaces the earlier variable-length `[ubyte]` UUID used solely for deduplication, at a fixed cost of 8 bytes with no heap allocation.
```
xiaoxichen: what is a `per-sub-request`?   a `ScrubReq` to a particular follower?

docs/adr/scrub-flatbuffers-schema-design.md
JacksonYao287: I don`t get this question. value is blob_id for blob scrub. 
yes, add blob_id is harmless. to make the fbs itself as simple as possible, the less Items the better.  
do you mean we can just rename value here to blob_id and we use it as blob count for shard scrub?

docs/adr/scrub-flatbuffers-schema-design.md
JacksonYao287: sure

docs/adr/scrub-flatbuffers-schema-design.md
JacksonYao287: yes, we will.  since shard meteblk resides on  nvme ssd, read them out will not take too much time.

but the verification is different from blob? we have no checksum for shard metablk, so I will read it into memory and compare with the in-memory shard meta. 

I am also thinking do we need a crc compare among the three replicas.  that is we compute the crc of each shard meta blk and see if it is the same among the three replicas.

docs/adr/scrub-flatbuffers-schema-design.md
JacksonYao287: it will be like blob scrub as we discussed. we set a large batch size (let`s say 10W) ,  then  in most cases , all the shard scrub results will be returned in a single scrub result response , and scans the pg index table only once.

docs/adr/scrub-flatbuffers-schema-design.md
JacksonYao287: sorry, it is typo . it should be per-request.

docs/adr/scrub-flatbuffers-schema-design.md
xiaoxichen: No I think we can add a blobID...using value for blobID is very tricky and hard to follow

docs/adr/scrub-flatbuffers-schema-design.md
xiaoxichen: yes this is exactly what I am asking,  I think we should use CRC to compare across replicas

docs/adr/scrub-flatbuffers-schema-design.md
xiaoxichen: then it is not absent, please also use "first batch " and "other batch " to describe. Also, if this is the case, do we want to set the default value of `end_shard_id` to max_uint64?

docs/adr/scrub-flatbuffers-schema-design.md
```diff
@@ -0,0 +1,283 @@
...
+
+### 4. Generic shard_id/value avoids multiple result entry types
+
+Two named entry structs (`MetaScrubResultEntry` and `BlobScrubResultEntry`) become one `ScrubResultEntry`. The semantic mapping (see table above) is documented once in the schema comments and once in the C++ dispatch code. When the leader receives a `ScrubResult`, it looks up the request context via `req_id` to obtain `scrub_type` and then interprets `shard_id`/`value` accordingly.
+
+### 5. `uint64` hash instead of `[ubyte]`
+
+Using a fixed-width `uint64` rather than `[ubyte]` removes the 4-byte vector-length prefix overhead and avoids a heap allocation on deserialization. CRC64/NVME fits in 8 bytes and offers stronger collision resistance than CRC32 for scrub purposes. If the hashing algorithm changes in the future, the field width can be revisited in a single schema change without altering the shape of `ScrubResultEntry`.
+
+### 6. `req_id` for epoch deduplication; `issuer_uuid` for sender identity
```
xiaoxichen: still, please provide a definition of a request --- even if we want to send identical req to 2 followers, we will set different req-id for them IIUC.

Also , do you thinking `sender_uuid` is better than `issuer_uuid`

xiaoxichen (commented):
lgtm, ideally if there is a sequence diagram can be generated by ai， will be easier to follower

docs/adr/scrub-flatbuffers-schema-design.md
JacksonYao287: compare shard meta blk crc is a little complicated, so I planned to do it but not yet in my code.

```
struct ShardInfo {
    static constexpr uint64_t meta_length = 1024 + 1;
    enum class State : uint8_t {
        OPEN = 0,
        SEALED = 1,
        DELETED = 2,
    };

    shard_id_t id;
    pg_id_t placement_group;
    State state;
    uint64_t lsn; // created_lsn
    uint64_t created_time;
    uint64_t last_modified_time;
    uint64_t available_capacity_bytes;
    uint64_t total_capacity_bytes;
    std::optional< peer_id_t > current_leader{std::nullopt};
    uint8_t meta[meta_length]{};

    auto operator<=>(ShardInfo const& rhs) const { return id <=> rhs.id; }
    auto operator==(ShardInfo const& rhs) const { return id == rhs.id; }
    bool is_open() const { return state == State::OPEN; }
};
```

pls see this, which exists in shard meta blk.  available_capacity_bytes and last_modified_time, although we don`t use them so far, but if some day we use them , they will probably be different across replicas since raft will change them all the time and thus lead to the crc mismatch. and if leader  changes , current_leader may be different , which will also make crc different.

if we really want to do crc , we can pick some key items , like shard_id, create_lsn, pg, state, to form a shard_crc_struct and compare the crc of shard_crc_struct.  we can do it in scrubber phase 2

the same as pg_info_blk

docs/adr/scrub-flatbuffers-schema-design.md
JacksonYao287: >we want to set the default value of end_shard_id to max_uint64?

yes., I am going to do like this.

> please also use "first batch " and "other batch " to describe
sure

docs/adr/scrub-flatbuffers-schema-design.md
JacksonYao287: >rename value here to blob_id
will this be better? for shard scrub, the blob_id means blob count.

docs/adr/scrub-flatbuffers-schema-design.md
xiaoxichen: yes this is correct, maybe we want to persist the `blob_count` we got from meta scrub as well that will create drift across replica.  The CRC should only check core fields.

Good to implement in phase 2, but better to standardise the protocol.. i.e even set a PLACE_HOLDER CRC into each shard as of now makes sense and wont break the logical.

docs/adr/scrub-flatbuffers-schema-design.md
JacksonYao287: >we will set different req-id for them IIUC.

no, we will send the same req-id to different follower.

req-id is used to identify a specifc scrub request.
issuer_uuid is used to identify where the scrub result from.

for leader, it will wait for the results with the same req(req-id) from different followers(issuer_id)  .

I just take the style from fetch_data proto. so I name it issuer_id.
https://github.com/eBay/HomeStore/blob/6e2fa5c576c4ee54f2fe8874627a747eca8d6d2d/src/lib/replication/fetch_data_rpc.fbs#L25

but sender_uuid seems better, let me change this

xiaoxichen: OK it is up to you , just please document to avoid ambiguous. 
Though I think the req across two replicas will drift once there is a blob diff --- the range will diff. So I would rather keep it per req (per client)

docs/adr/scrub-flatbuffers-schema-design.md
JacksonYao287: renamed value to blob_id, as a result, for meta scrub, blob_id holds the shard`s blob count

docs/adr/scrub-flatbuffers-schema-design.md
JacksonYao287: you are right, per req req_id is better and correct, I have add comments for this in the adr doc

docs/adr/scrub-blob-range-coverage.md
```diff
@@ -122,6 +122,8 @@ Two special cases apply:
 
 **Key safety property**: Because the count comes from the most-populated replica rather than the leader's own view, even if the leader has lost blobs in a shard, the coverage range for that shard is still driven by the follower's larger count.
 
+we only scan those shards whose `sealed_lsn ≤ scrub_lsn`, becasue the state of these shard will never change, so that we have a fixed set of shards. otherwise, if some shards are created after scrub_lsn, even if we wait for the commit_lsn >= scrub_lsn before scrubbing shard, we probably get different set of shards on different replicas, which will makes some fake missing shards. if we have a fixed set of shards and all the shards in this set are sealed(no more blobs will be put into these shard), last_blob_id is not necessary anymore.
```
JacksonYao287: @xiaoxichen  pls take a look at this and check if this makes sense to you.

since I want to scrub only those shards whose `sealed_lsn ≤ scrub_lsn`, last_blob_id in not necessary anymore, so I removed it from flatbuffer fbs file.

xiaoxichen: I dont get why a `last_shard_id`  doesnt solve the problem? ? This is a big changes, we need more reasoning and have this go throught with team.

docs/adr/scrub-flatbuffers-schema-design.md
xiaoxichen: have you address this?

docs/adr/scrub-flatbuffers-schema-design.md
xiaoxichen: where is the comments you added?

docs/adr/scrub-flatbuffers-schema-design.md
```diff
@@ -0,0 +1,286 @@
...
+  scrub_results: [ScrubResultEntry];
+}
+root_type ScrubResult;
+```
+
+### shard_id / blob_id / hash semantics by ScrubType
+
+| `scrub_type` | `shard_id` | `blob_id` | `hash` |
+|---|---|---|---|
+| `META` | `shard_id` | `blob_count` | hash of metablk |
```
JacksonYao287: updated

docs/adr/scrub-flatbuffers-schema-design.md
```diff
@@ -0,0 +1,286 @@
...
+
+// ─── Single request table ─────────────────────────────────────────────────────
+/// Covers all three scrub types.  The receiver inspects scrub_type to determine
+/// which range fields are meaningful.
+table ScrubReq {
+  pg_id:          uint16;
+  /// Randomised per-request identifier used by the receiver to match a response to
+  /// the outstanding request and to discard stale responses from a previous scrub
+  /// epoch.  The leader generates a distinct req_id for every (replica, request) pair:
+  /// two replicas receiving the same scrub round will each see a different req_id,
```
JacksonYao287, line 64: add comments for req_id

docs/adr/scrub-blob-range-coverage.md
```diff
@@ -54,9 +54,30 @@ All entries in the index table are keyed as `{shard_id, blob_id}`, sorted with `
 
 This design combines data produced during the shard scrub phase with a dynamic bin-packing algorithm and a streaming response protocol to satisfy all three constraints simultaneously.
 
-### Step 1: Build shard_blob_count from the Shard Scrub Phase
+### Step 1: META Scrub
```
JacksonYao287, line 57: @xiaoxichen  pls take a look , I have add the decription for meta scrub

xiaoxichen (commented):
lgtm.

A few other thinking point you might doc in here or other doc, or re-purpose the first doc into scrub flow etc.

1.  Step 1 when qeurying the PG btree, assuming all the shards are big shards, consider the lock intention with incoming IOs.
2.  How the shard metablk crc works
3.  When leader starts a scrubbing task,  it is not inside the state-machine which means the read to `lsn, last_shard_id, last_blobid` is not transactional.  Say LSN 101 is a create shard(0xAA) and LSN 102 is a put blob(0xBB), if we read <LSN=100, 0xAA, 0xBB> that will cause some false-positive of missing shard/missing PG. If we do spot check then it is not a big deal.  A 60% confidence thought is that we read the LSN at last which ensures the scrub-lsn covers last_shard_id and last_blob_id, make it <LSN=102, 0xAA, 0xBB>.  More carefully thinking to bump up the confidence level is needed.

JacksonYao287:
> 1 Step 1 when qeurying the PG btree, assuming all the shards are big shards, consider the lock intention with incoming IOs.

IIUC, when we do range query, let`s say the range covers  2 btree nodes. when we scan the second node, [we will not lock the first node] (https://github.com/eBay/HomeStore/blob/860330af43973ff99a862f6da17329de045dedc1/src/include/homestore/btree/detail/btree_query_impl.ipp#L33) ?.  
I need to confirm this by checking the code carefully. thanks for putting the point here ， I will check it later since LLM is not available now(504 error) , sigh...

> How the shard metablk crc works

the key point for pg/shard crc is which fields we will use to calculate the metablk crc, since some fields will probably be different among different replicas. we can discuss it later.
 
 >lsn, last_shard_id, last_blobid is not transactional.
 
 pls see 
 https://github.com/JacksonYao287/HomeObject/blob/40463ac34020dd3f06cf26c8bb80d2dbfe77e9d6/src/lib/homestore_backend/scrub_manager.cpp#L792
 
 and
 
 https://github.com/JacksonYao287/HomeObject/blob/40463ac34020dd3f06cf26c8bb80d2dbfe77e9d6/src/lib/homestore_backend/scrub_manager.cpp#L858

in my implementation, we get last_shard_id and last_blob_id before getting scrub_lsn. This can ensure that when committing to scrub_lsn, a replica should at least see last_shard_id and last_blob_id. this seems basically the same point as you put here.

xiaoxichen:
makes sense, just putting them into doc as well.

JacksonYao287:
sure , I have added it to the doc. 

by the way, I have check index btree with AI, when we do range query, we will not lock the entire btree, we only lock the btree node we are scanning. other nodes of this btree can be read and write in the middle of range query

JacksonYao287:
since we have the new target branch (stable/v4.x), I close this PR and use #410 to replace it

PR #408, opened 2026-04-14
SDSTOR-21848: Fix empty snapshot resync flow

yuwmao:
When a PG has no shards, write_snapshot_obj() called objId(get_sequence_num_from_shard_id(shard_list_end_marker), 0) which produced offset 0xFFFFFFFFFFFF8000 instead of LAST_OBJ_ID.
The leader decoded shard_seq_num=0xFFFFFFFFFFFF, found no matching shard, and logged "Invalid objId".

Adding UT in dynamic test(instead of `homestore_test_misc`) is to test the interaction between the follower and leader.

### DIFF
--- a/src/lib/homestore_backend/tests/test_homestore_backend_dynamic.cpp
+++ b/src/lib/homestore_backend/tests/test_homestore_backend_dynamic.cpp
@@ -956,6 +956,184 @@ TEST_F(HomeObjectFixture, RollbackReplaceMember) {
     g_helper->sync();
 }
 
+// Regression test: baseline resync for a PG with no shards must complete without getting stuck
+// in an infinite loop. Without the fix, write_snapshot_obj computes
+//   objId(get_sequence_num_from_shard_id(shard_list_end_marker), 0)
+// which gives shard_seq_num=0xFFFFFFFFFFFF, causing the leader to log "Invalid objId" and
+// reset its cursor forever. With the fix it correctly returns LAST_OBJ_ID.
+TEST_F(HomeObjectFixture, BaselineResyncEmptyPG) {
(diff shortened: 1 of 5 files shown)


src/lib/homestore_backend/tests/test_homestore_backend_dynamic.cpp
```diff
@@ -956,6 +956,184 @@ TEST_F(HomeObjectFixture, RollbackReplaceMember) {
...
+
+    // Step 2: Force a snapshot on the leader before replace_member so the spare
+    // can exercise baseline resync from the snapshot path instead of append_entries.
+    run_on_pg_leader(pg_id, [&]() {
+        auto* hs_pg = _obj_inst->get_hs_pg(pg_id);
+        ASSERT_NE(hs_pg, nullptr);
+        const auto compact_lsn = hs_pg->repl_dev_->get_last_commit_lsn();
+        ASSERT_GT(compact_lsn, 0);
+        LOGINFO("trigger manual snapshot before replace_member, pg={}, compact_lsn={}", pg_id, compact_lsn);
+        _obj_inst->trigger_snapshot_creation(pg_id, compact_lsn, true /* wait_for_commit */);
```
yuwmao, line 989: The reason why I use an explicit trigger instead of `snapshot_freq_distance/num_reserved_log_items` cfg is that the first log truncation triggered by snapshot will be skipped(truncation_upper_limit=0) and the log `truncation_upper_limit` is also clamped by `checkpoint_lsn`, so `trigger_snapshot_creation` is more deterministic.

xiaoxichen (commented):
lgtm

xiaoxichen (approved):

PR #409, opened 2026-05-06
SDSTOR-21989: check index_table existence before using it

JacksonYao287:
if homeobject crashes after index_table is destroyed in destroy_pg and before the pg is created in baseline resync, we can not find pg index table when restarts. so we need add a sanity check to see if index_table exists. if not, skip refreshing pg metrics

### DIFF
--- a/conanfile.py
+++ b/conanfile.py
@@ -10,7 +10,7 @@ required_conan_version = ">=1.60.0"
 
 class HomeObjectConan(ConanFile):
     name = "homeobject"
-    version = "4.1.11"
+    version = "4.1.12"
 
     homepage = "https://github.com/eBay/HomeObject"
     description = "Blob Store built on HomeStore"
--- a/src/lib/homestore_backend/hs_blob_manager.cpp
+++ b/src/lib/homestore_backend/hs_blob_manager.cpp
@@ -88,7 +88,7 @@ BlobManager::AsyncResult< blob_id_t > HSHomeObject::_put_blob(ShardInfo const& s
         return folly::makeUnexpected(BlobErrorCode::SHUTTING_DOWN);
     }
     incr_pending_request_num();
-        // check user key size
+    // check user key size
     if (blob.user_key.size() > BlobHeader::max_user_key_length) {
         BLOGE(tid, shard.id, 0, "input user key length > max_user_key_length {}", blob.user_key.size(),
               BlobHeader::max_user_key_length);
@@ -167,8 +167,7 @@ BlobManager::AsyncResult< blob_id_t > HSHomeObject::_put_blob(ShardInfo const& s
 
     // Set offset of actual data after the blob header and user key (rounded off)
     req->blob_header()->data_offset = req->blob_header_buf().size();
-    RELEASE_ASSERT(req->blob_header()->data_offset == _data_block_size,
-                       "blob header should equals _data_block_size");
+    RELEASE_ASSERT(req->blob_header()->data_offset == _data_block_size, "blob header should equals _data_block_size");
     // In case blob body is not aligned, create a new aligned buffer and copy the blob body.
     if (((r_cast< uintptr_t >(blob.body.cbytes()) % io_align) != 0) || ((blob_size % io_align) != 0)) {
         // If address or size is not aligned, create a separate aligned buffer and do expensive memcpy.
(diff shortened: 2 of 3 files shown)


xiaoxichen:
In general LGTM

1. During recovery, should we better handling destroyed PG? Currently we only partially handled the case that Index is destroyed on destroyed PG. (https://github.com/eBay/HomeObject/blob/77ba5c670f32ebe87c4ef95810126ce162a36cb5/src/lib/homestore_backend/hs_pg_manager.cpp#L951-L955)

I think we should redo the `HSHomeObject::pg_destroy` during recovery.

2.  the check index_table logic spread across the code,  is it better to add an accessor into HS_PG? which checks the existence of index and logging (if the PG is destroyed) or assert.

Besroy:
Agree, checking the PG state seems safer and cleaner. @yuwmao @koujl Do you recall the previous solution for handling crashes during destroy_pg? Perhaps we can integrate it with this fix as a unified solution

JacksonYao287:
agree, we need to redo pg_destroy to make sure all the pg resource are cleared before pg is recreated by BR or permanently destroyed.  cc @yuwmao @koujl  if you want to do this,  feel free to close this PR.

yuwmao:
No, I haven't met the destroy_pg crash issue before in my memory. But also agree on redo pg_destroy.
> Agree, checking the PG state seems safer and cleaner. @yuwmao @koujl Do you recall the previous solution for handling crashes during destroy_pg? Perhaps we can integrate it with this fix as a unified solution

JacksonYao287:
I have added an accessor to index_table.  redo pg destory will come in a later separate PR

src/lib/homestore_backend/gc_manager.cpp
```diff
@@ -541,10 +541,9 @@ bool GCManager::pdev_gc_actor::replace_blob_index(
     auto pg_id = move_from_vchunk->m_pg_id.value();
     auto hs_pg = m_hs_home_object->get_hs_pg(pg_id);
 
-    // TODO:: add logic to handle pg_index_table is nullptr if destroying pg happens when GC
     RELEASE_ASSERT(hs_pg, "Unknown PG for pg_id={}", pg_id);
-    auto pg_index_table = hs_pg->index_table_;
-    RELEASE_ASSERT(pg_index_table, "Index table not found for PG pg_id={}", pg_id);
+    auto pg_index_table = hs_pg->get_index_table();
+    if (!pg_index_table) { return false; }
```
Besroy: If the PG was destroyed, will GC/BR/put_blob/get_blob operations still occur after a restart? If not, can we limit the changes to `refresh_pg_statistics`? 
The reason is when I reviewing the replace_blob_index caller, I noticed that returning false causes a progress assertion failure, as do most functions that access it after retrieval.

JacksonYao287: sorry, I don`t fully get the point, can you pls put more details here?

actually, assertion failure is necessary. there are two case that false will be returned.

1  pg index table is not found.  this should not happen since if there is an on-going gc task,  pg can not be destroyed.  pls refer to drain_pg_pending_gc_task, which will wait until all the gc task of this pg is completed

2 failed to update pg index table. this means partial updated(some are updated successfully and others not) probably happens and it is very dangerous since data loss will happen if we don`t assert.

JacksonYao287: >If the PG was destroyed, will GC/BR/put_blob/get_blob operations still occur after a restart

pg is destroyed in 3 cases:  br (only at follower), destroy repl_dev(all member) and replace_member(only at out member)

if br case , pg will be recreated again, and thus GC/BR/put_blob/get_blob operations probably occurs after restart.

Besroy: What I mean is: keep the existing assert logic in those places unchanged, because:

  1. Correct me if I’m wrong, but GC / put_blob / get_blob are not expected to run on a destroyed PG. BR will destroy the PG again, so this case shouldn’t happen at that step either. This scenario would only show up during a restart when the PG state is inconsistent 
  2. Even if you get nullptr in those places, you should still keep the original asserts. Because even if you don’t assert there, we will still assert later based on the return value, or we’ll crash when dereferencing a null pointer anyway

JacksonYao287: @Besroy , done, ptal

xiaoxichen (approved):
lgtm considering this is a temporary fix

PR #411, opened 2026-05-06
SDSTOR-21407 Make per-service dev_type and size_pct configurable in i…

xiaoxichen:
…nit_homestore

Add HSDevTypeOverride enum and per-service {meta,log,index,replication}_{dev_type,size_pct} fields to HSBackendSettings. Defaults are AUTO/-1.0 so existing hybrid/single mode logic is fully preserved. When a config value is set it overrides the built-in default via resolve_dev_type() and resolve_size_pct() helpers.

Also tune standalone (single-device) defaults: meta/index 5%- >1%, replication 79%->87%.

### DIFF
--- a/docs/adr/blob-index-analyze.md
+++ b/docs/adr/blob-index-analyze.md
@@ -1,0 +1,311 @@
+# HomeObject Blob Index Analyze
+
+**Date:** 2026-05-06
+**Purpose:** Reference for blob index sizing and capacity planning
+
+---
+## Execution Summary
+
+### Scenario 1: 20TB Data with 2GB Memory Budget
+
+**Configuration:**
+- Total data: 20TB
+- `io_mem_size`: 2GB
+- **Fixed cache pool:** 2GB × 65% = **1.3 GB** (always allocated)
+
+| Blob Size | Blob Count | Index Disk | Hard Memory | Cache Pool | Cache Coverage |
+|-----------|------------|------------|-------------|------------|----------------|
+| **8 KB**  | 2.56B      | 60.1 GB    | 177 MB      | 1.3 GB     | **2.16%** ⚠️  |
+| **256 KB**| 80M        | 1.88 GB    | 5.51 MB     | 1.3 GB     | **69%** ✓      |
+
+
+---
+
+### Scenario 2: 128TB Data with 2GB Memory Budget
+
+**Configuration:**
+- Total data: 128TB
+- `io_mem_size`: 2GB
+- **Fixed cache pool:** 2GB × 65% = **1.3 GB** (always allocated)
+
+| Blob Size | Blob Count | Index Disk | Hard Memory | Cache Pool | Cache Coverage |
+|-----------|------------|------------|-------------|------------|----------------|
+| **8 KB**  | 16.4B      | 385 GB     | 1.13 GB     | 1.3 GB     | **0.34%** ⚠️⚠️ |
+| **256 KB**| 512M       | 12.0 GB    | 35.3 MB     | 1.3 GB     | **10.8%** ✓    |
+
+
+---
+
+## Action Items
+### Shrink down the size of Index_vdev
+#### [HDD SKU]
+Now we use 45% of META drive (200GB*45% =90GB) on HDD SKUs. The number is generally correct as worst case we can consume up to 60GB.
+Shrinking it down from 90GB to 60GB saves 60MB memory which is negligible.
+#### [QLC SKU]
+5% of QLC size makes it into 5721.96 GB,  as a result it consumes 18GB memory for allocator.
+Changing it to 0.8% based on below calculation, as a result, Index size would be ~1TB with Hard Memory ~3GB.
+
+```
+Worst case assuming blob size is 8KB, the Index size should be 0.61% of DataSize
+
+(DataSize/8K)*(4KB/167)*1.01 ==> DataSize *1.01/167 = DataSize * 0.61%.
+```
+### Configure mem_size for QLC nodes.
+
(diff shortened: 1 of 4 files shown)


src/lib/homestore_backend/hs_homeobject.cpp
```diff
@@ -204,36 +213,61 @@ void HSHomeObject::init_homestore() {
...
+                     .size_pct = resolve_size_pct(HS_BACKEND_DYNAMIC_CONFIG(replication_size_pct), 99.0f),
+                     .num_chunks = 0,
+                     .chunk_size = _hs_chunk_size,
+                     .block_size = _data_block_size,
+                     .alloc_type = blk_allocator_type_t::append,
+                     .chunk_sel_type = chunk_selector_type_t::CUSTOM}},
             });
         } else {
             auto run_on_type = has_fast_dev ? homestore::HSDevType::Fast : homestore::HSDevType::Data;
             LOGD("Running with Single mode, all service on {}", run_on_type);
```
JacksonYao287: nit: if we have different final HSDevType for different service, it is better to print them respectively.

docs/adr/blob-index-analyze.md
```diff
@@ -0,0 +1,311 @@
...
+
+---
+
+## Memory Overhead
+
+For index vdev size S:
+- **Allocator overhead:** ~0.2% of S
+    - Uses `FixedBlkAllocator` (4KB fixed block size)
+    - Maintains all free blocks in memory via `folly::MPMCQueue`
+    - Each slot: 12 bytes (4B blk_num_t + 8B atomic sequence)
```
JacksonYao287: sequence is atomic < uint32 >, so it is 4B, and each slot is 8B?

docs/adr/blob-index-analyze.md
```diff
@@ -0,0 +1,311 @@
...
+Now we use 45% of META drive (200GB*45% =90GB) on HDD SKUs. The number is generally correct as worst case we can consume up to 60GB.
+Shrinking it down from 90GB to 60GB saves 60MB memory which is negligible.
+#### [QLC SKU]
+5% of QLC size makes it into 5721.96 GB,  as a result it consumes 18GB memory for allocator.
+Changing it to 0.8% based on below calculation, as a result, Index size would be ~1TB with Hard Memory ~3GB.
+
+```
+Worst case assuming blob size is 8KB, the Index size should be 0.61% of DataSize
+
+(DataSize/8K)*(4KB/167)*1.01 ==> DataSize *1.01/167 = DataSize * 0.61%.
```
JacksonYao287: 8K/ 4K = 2？

it should be ( (DataSize/8K) / 167 ) *4k * 1.01  = DataSize * 0.3%.

docs/adr/blob-index-analyze.md
```diff
@@ -0,0 +1,311 @@
...
+
+---
+
+## Action Items
+### Shrink down the size of Index_vdev
+#### [HDD SKU]
+Now we use 45% of META drive (200GB*45% =90GB) on HDD SKUs. The number is generally correct as worst case we can consume up to 60GB.
+Shrinking it down from 90GB to 60GB saves 60MB memory which is negligible.
+#### [QLC SKU]
+5% of QLC size makes it into 5721.96 GB,  as a result it consumes 18GB memory for allocator.
```
JacksonYao287: can you also put the capacity of a QLS SSD here to make it more clear

docs/adr/blob-index-analyze.md
```diff
@@ -0,0 +1,311 @@
...
+```
+1. Blob count (N)       = Data size / Avg blob size
+
+2. Index disk space     ≈ (N / 167) × 4KB × 1.01
+                          (1.01 factor accounts for internal nodes)
+
+3. Hard memory          = Index disk / 4096 × 12 bytes
+   (Allocator)            (Always resident, non-negotiable)
+
+4. Soft memory          = min(Index disk × 10%, io_mem_size × 65%)
```
JacksonYao287: what does Index disk × 10%  mean?

docs/adr/blob-index-analyze.md
xiaoxichen: Thanks, nice catch, I copied this from the GDB result from Mehdi but actually he was in older version where BlkId (6B) was used for MPMCQ , 6+4=10 then padded to 12.

In 7.x we are using `blk_num_t` instead of `blk_idwhich ` is 4B , then no padding is needed

docs/adr/blob-index-analyze.md
xiaoxichen: trying to say keeping 10% of index in the memory for hot data (if any)

src/lib/homestore_backend/hs_homeobject.cpp
xiaoxichen: ack

JacksonYao287 (approved):
LGTM

Besroy:
LGTM, but I’m a bit confused about the use case for config dev_type configurable. What is its priority compared to the device type detected by HO? Since the dev info used by HS is detected by HO, but the vdev creation relies on the parameters specified in HO’s `hs_format_params`, does this mean the physical disk type remains as detected unless the override doesn’t change the hybrid/single mode, and HS’s vdev can be deployed on any disk type in this case? Also, is the cumulative percentage for the same type being kept under 100% ensured by HS’s checks?

xiaoxichen:
@Besroy
1.  the configuration takes priority ,  if configuration is wrong then the disk init would fail.
2.  if the config is not set,  auto-detect would be used.
3. I dont have clear use case for over-ridding the dev-type for specific vdev as of now... Honestly i even dont think we should tune the ratio once we have a good number for LOG, but makes them into config makes tuning easier in future.
4. the <100% is ensured by HS.

PR #413, opened 2026-05-07
SDSTOR-21465: scrubber phase 1

JacksonYao287:


src/lib/homestore_backend/scrub_manager.cpp
```diff
@@ -0,0 +1,2123 @@
...
+                                return;
+                            }
+
+                            const auto& exists = result.value();
+                            if (!exists) {
+                                SCRUBLOGD(pg_id, task_id,
+                                          "reconcile check: shard {} confirmed absent on peer {}, removing from "
+                                          "existence-tracking set",
+                                          shard_id, peer_id);
+                                scrub_report->remove_shard_existence_from_peer(shard_id, peer_id);
```
Hooper9973, line 1354: I think there may be an iterator invalidation risk here. `missing_shards` is returned by const reference, so this loop is walking the original map. If `remove_shard_existence_from_peer()` erases the current `shard_id` from that same map, the current range-for element / iterator could be invalidated. 

Might be safer to iterate over a snapshot copy, or apply removals after the traversal.

JacksonYao287: nice finding, but after checking the code, seems it is safe.

 line 1439,  use copy constructor to get a new  instance `missing_shards` (this is just like the snapshot you metioned).
```
const auto missing_shards = scrub_report->get_missing_shard_ids();
```
so, actually , the behavior is we iterate on a copied new instance of missing_shards( which is never changed), and the removal happens in the `source missing_shards`( inside scrub_report).  since they are totally different missing_shards, so it`s totally safe.

pls correct me if I am wrong

src/lib/homestore_backend/scrub_manager.cpp
```diff
@@ -0,0 +1,2002 @@
...
+    }
+    if (m_scrub_req_executor) {
+        m_scrub_req_executor->stop();
+        m_scrub_req_executor.reset();
+    }
+    LOGINFOMOD(scrubmgr, "scrub manager stopped!");
+}
+
+void ScrubManager::add_scrub_req(std::shared_ptr< scrub_req > req) {
+    m_scrub_req_executor->add([this, req = std::move(req)]() { handle_scrub_req(req); });
```
xiaoxichen: m_scrub_req_executor can be null

xiaoxichen (changes requested):
The NPE bug should be fixed before merging, other part LGTM.

The NPE can be triggered if  some SM enabled scrubbing during config change/upgrade, then causing cluster wide crashing.

src/lib/homestore_backend/scrub_manager.cpp
JacksonYao287: done , PTAL

xiaoxichen (dismissed):

xiaoxichen:
the ut failure is in scubber test, please take a look

JacksonYao287:
I triggered this UT again and CI passed.  Moreover, I run this UT locally for several times , but can not reproduce this failure.

I think it is a flaky case

JacksonYao287:
I tried several times again, but still can not reproduce this case.

I suspect the root cause is the unexpected leader switch during UT, just like the flaky stuck case we can sometimes see in  homestore_test_pg/shard/blob. 

if unexpected leader switch happens during UT, 
1 for scrubber test, the scrub task will be cancelled and leader will get an incorrect scrub report , which will result the UT failure.

2 for homestore_test_pg/shard/blob, follower will wait for something to happen,  and leader think it is not leader any more( because of leader switch) and do not schedule some op, then all the member will sync and wait at some point, and thus the UT is stuck.

so, I think thing we need try to handle the leader switch case  in raft test framework.

xiaoxichen:
OK, lets wait a few days for other team members to review. If no further comment lets merge as is.

xiaoxichen (approved):

PR #414, opened 2026-05-13
SDSTOR-21408 : refine index_kv constructor

JacksonYao287:
root cause is not figured out. but let`s make this change to refine index_kv constructor and see if this issue will occur again

### DIFF
--- a/conanfile.py
+++ b/conanfile.py
@@ -10,7 +10,7 @@ required_conan_version = ">=1.60.0"
 
 class HomeObjectConan(ConanFile):
     name = "homeobject"
-    version = "4.1.12"
+    version = "4.1.13"
 
     homepage = "https://github.com/eBay/HomeObject"
     description = "Blob Store built on HomeStore"
--- a/src/lib/homestore_backend/index_kv.hpp
+++ b/src/lib/homestore_backend/index_kv.hpp
@@ -16,7 +16,7 @@ private:
 public:
     BlobRouteKey() = default;
     BlobRouteKey(const BlobRoute key) : key_(key) {}
-    BlobRouteKey(const BlobRouteKey& other) : BlobRouteKey(other.serialize(), true) {}
+    BlobRouteKey(const BlobRouteKey& other) : homestore::BtreeKey(), key_(other.key_) {}
     BlobRouteKey(const homestore::BtreeKey& other) : BlobRouteKey(other.serialize(), true) {}
     BlobRouteKey(const sisl::blob& b, bool copy) :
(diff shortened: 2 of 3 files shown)


src/lib/homestore_backend/index_kv.hpp
```diff
@@ -16,7 +16,7 @@ class BlobRouteKey : public homestore::BtreeKey {
 public:
     BlobRouteKey() = default;
     BlobRouteKey(const BlobRoute key) : key_(key) {}
-    BlobRouteKey(const BlobRouteKey& other) : BlobRouteKey(other.serialize(), true) {}
+    BlobRouteKey(const BlobRouteKey& other) : homestore::BtreeKey(), key_(other.key_) {}
```
Besroy, line 19: `homestore::BtreeKey()` works now because it has no non-static data members, so the default copy constructor does nothing. However, if a data member is added in the future, these constructors may skip copying it without any compiler warning.

conanfile.py
```diff
@@ -10,7 +10,7 @@
 
 class HomeObjectConan(ConanFile):
     name = "homeobject"
-    version = "4.1.11"
+    version = "4.1.13"
```
Besroy, line 13: Version jumps from 4.1.11 to 4.1.13, skipping 4.1.12

JacksonYao287: this PR #409 takes 4.1.12, so in this PR I use 4.13.  if you got cycle, can you first review #409 ? it`s also a small change

xiaoxichen:
what is the RCA of the issue?

src/lib/homestore_backend/index_kv.hpp
JacksonYao287: if we add a new member , we should decide how to initialize it and add the corresponding code. 

BtreeKey is designed to be a pure virtual base class with no data member. if some day we have to add a data member to it,  we have to choices:
1 change all the BtreeKey derived class.  this will involve a lot of changes, not only homeobject, but also nublox and other upper layers, so I prefer not to do this.

2 add an implementation to  homestore::BtreeKey() where we can initialize the newly added member. this change will be unseeable  to the upper layer.

 compiler is not responsible for raising warnings for those newly added items unless it has compiling error.  it`s our responsibility to make every newly added member well-initialized, because only we can decide how to initialize it(explicitly or default)

JacksonYao287: what`s more, in the constructor of a derived class, the constructor of base class will be always called. which means, even if we do not put homestore::BtreeKey() here, it is actually called.

Besroy: Make sense, thanks for the clarification

PR #418, opened 2026-05-27
SDSTOR-22399: Add aggressive check in gc

JacksonYao287:


src/lib/homestore_backend/gc_manager.cpp
```diff
@@ -829,6 +838,21 @@ bool GCManager::pdev_gc_actor::copy_valid_data(
                         return folly::makeFuture< bool >(true);
                     }
 
+                    // check if all the pbas in the valid_blob_indexes are in move_from_chunk, if not, it means the
```
xiaoxichen: The `valid_blob_indexes`,` move_from_chunk` and `move_to_chunk` are both ready much earlier in the flow,  should we move all the checks including L833-840 up  to where the Index was queried.?

src/lib/homestore_backend/gc_manager.cpp
```diff
@@ -966,6 +990,10 @@ bool GCManager::pdev_gc_actor::copy_valid_data(
                             }
 
                             // write shard footer, which occupies one blk
+
+                            // in fact , we have a dangling reference of out_blkids here, but it does not matter since
+                            // we do not use it after this and the async_alloc_write will complete soon and the memory
+                            // of out_blkids will be released after that.
```
xiaoxichen, line 1006: better to fix this into `async_alloc_write` which allows `out_blkids` to be nullptr meaning I dont care.

src/lib/homestore_backend/gc_manager.cpp
```diff
@@ -1117,7 +1146,7 @@ bool GCManager::pdev_gc_actor::purge_reserved_chunk(chunk_id_t chunk, const uint
     return true;
 }
 
-bool GCManager::pdev_gc_actor::compare_blob_indexes(
+bool GCManager::pdev_gc_actor::double_check_blobs(
```
xiaoxichen: try to get a better naming from AI, maybe check_all_blobs_copied?

src/lib/homestore_backend/gc_manager.cpp
```diff
@@ -976,13 +1004,14 @@ bool GCManager::pdev_gc_actor::copy_valid_data(
...
+                                //  1 any blob copy fails, then err is operation_cancelled
                                 //  2 write footer fails， then err is the error code of write footer
-                                GCLOGE(task_id, pg_id, shard_id,
-                                       "Failed to copy some blos or failed to write shard footer for move_to_chunk={}, "
-                                       "err={}, error_category={}, error_message={}, pls check the log for more "
-                                       "detailed info",
-                                       move_to_chunk, err.value(), err.category().name(), err.message());
+                                GCLOGE(
+                                    task_id, pg_id, shard_id,
+                                    "Failed to copy some blobs or failed to write shard footer for move_to_chunk={}, "
```
xiaoxichen, line 1021: Can we use error code de distinguish ?

src/lib/homestore_backend/gc_manager.cpp
```diff
@@ -577,13 +577,22 @@ bool GCManager::pdev_gc_actor::replace_blob_index(
...
+                        // in recovery
+                        RELEASE_ASSERT(existing_pbas.chunk_num() == move_to_chunk,
+                                       "existing pbas chunk={} should be equal to move_to_chunk={}, blob_id={}, "
+                                       "move_from_chunk={}, existing_pbas={}, new_pbas={}",
+                                       existing_pbas.chunk_num(), move_to_chunk, blob, move_from_chunk,
+                                       existing_pbas.to_string(), new_pbas.to_string());
+                        return homestore::put_filter_decision::keep;
+                    }
+
+                    RELEASE_ASSERT(
```
xiaoxichen: why we want to assert?  If none of the checks in earlier stages catches the inconsistency,  when reaching here, meaning we had copied all blobs correctly, regardless where the original PBA, to the move-to chunk. Then why we dont want to add them into index (which fixed the divergence) ? 

How assert in this case helps?

xiaoxichen (commented):
Commit message need specific what are the "aggressive. checks" added?

src/lib/homestore_backend/gc_manager.cpp
JacksonYao287: sure, will do this. but we need to keep L833-840 here, the aim here is that we need only write a shard hear for a empty open shard in emergent gc caes, which means the emergent gc is triggered by the first put_blob of this shard.

src/lib/homestore_backend/gc_manager.cpp
JacksonYao287: yes, this change should be first made in hs sides. after homestore change is made, we can come back and fix this

src/lib/homestore_backend/gc_manager.cpp
JacksonYao287: the type of err here is std::error_code, which has node error code.

err.value(), err.category().name(), err.message() is enough ti distigguish them

src/lib/homestore_backend/gc_manager.cpp
JacksonYao287: done. rename it to check_blobs_consistency

src/lib/homestore_backend/gc_manager.cpp
JacksonYao287: the intention here is to crash sm and give us a chance to see why divergency occurs. it is ok to let it go, but we lose the change to figure out the root cause why divergency comes out.

I agree to change it to a warn log and let it go ATM. let`s change it to release assert for the sake of debugging in the future if necessary.

src/lib/homestore_backend/gc_manager.cpp
xiaoxichen: can we at least distingush between footer error  vs blob error?

src/lib/homestore_backend/gc_manager.cpp
xiaoxichen: sure, lets do it

src/lib/homestore_backend/gc_manager.cpp
```diff
@@ -577,13 +582,21 @@ bool GCManager::pdev_gc_actor::replace_blob_index(
...
-                        new_pbas.to_string());
-                    return homestore::put_filter_decision::keep;
+                    if (!task_id) {
+                        // in recovery
+                        RELEASE_ASSERT(existing_pbas.chunk_num() == move_to_chunk,
+                                       "existing pbas chunk={} should be equal to move_to_chunk={}, blob_id={}, "
+                                       "move_from_chunk={}, existing_pbas={}, new_pbas={}",
+                                       existing_pbas.chunk_num(), move_to_chunk, blob, move_from_chunk,
+                                       existing_pbas.to_string(), new_pbas.to_string());
+                        return homestore::put_filter_decision::keep;
```
xiaoxichen: I think  the logic should be 
```
if (in_recovery) && (exising_pba.chunk_num() == move_to_chunk)  {
     return keep;
} else {
     // LOG WARN: DIRVERGEN CHUNK
     // DEBUG ASSERT IS FINE
}
return replace
```

src/lib/homestore_backend/gc_manager.cpp
JacksonYao287: //  1 any blob copy fails, then err is operation_canceled
                                //  2 write footer fails， then err is the error code of write footer

the comments says:  for blob, the err is operation_canceled. for footer, it depends on the write error from os.

src/lib/homestore_backend/gc_manager.cpp
JacksonYao287: the logic is the same as what it is here , right? 

if existing_pbas.chunk_num() != move_from_chunk and !task_id(recovery), it will go down and hit homestore::put_filter_decision::replace


if we don`t want to assert, actually we can always do replace, even if the existing value is correct.  we can use a pba to replace the same pba。

JacksonYao287: I got your point， you mean we do not crash at line 587?

xiaoxichen: yes , recovery doesn't necessary to be more strict.

JacksonYao287: done, ptal

src/lib/homestore_backend/gc_manager.cpp
```diff
@@ -778,6 +797,21 @@ bool GCManager::pdev_gc_actor::copy_valid_data(
                    move_from_chunk);
         }
 
+        // check if all the pbas in the valid_blob_indexes are in move_from_chunk, if not, it means the
```
xiaoxichen, line 800: I am not sure this check makes sense after second thought.
If we hit here, the data is already spread across 2 chunks, meaning, we had hit a bug.
Stopping GC in this case makes some sense to avoid more damage. 

```
if not, it means the shard is being modified during gc, 
```
But your the reasoning is not correct,  it doesnt means the shard was modified during GC as of now.

To best-effort-detect the GC racing with put-blob etc,  the right check should be fetching the blob list N times with K seconds interval , and ensure the blob count is consistent...

xiaoxichen (approved):
lgtm

src/lib/homestore_backend/gc_manager.cpp
JacksonYao287: let`s rethink this  again thoroughly , and use another separate PR to do this.

PR #430, opened 2026-06-08
SDSTOR-22636: support yield leader to a specific candidate

Besroy:
also return more shard details on get_shard/dump_chunk/dump_shard API


xiaoxichen (approved):
lgtm

PR #435, opened 2026-06-16
SDSTOR-22424: redo destroy pg

JacksonYao287:
Fix crash-recovery gap in PG destroy, handle stale destroyed PGs on restart

  Extracted PG resource cleanup into a reusable destroy_pg_resource() helper and fixed two crash-recovery bugs:

  Problem 1 — Stale destroyed PGs after crash:
  When a repl_dev is marked destroyed (via leave()) but a crash occurs before pg_destroy cleans up the PG superblk, subsequent recovery finds a PG superblk with no corresponding repl_dev.
  Previously this triggered a fatal error log and left the PG resources dangling.

  Fix: on_pg_meta_blk_found now detects this case and tracks the pg_id in destoryed_stale_pgs_. On on_replica_restart, destroy_pg_resource is called for each stale PG to reclaim its chunks,
  index table, and superblk.

  Problem 2 — Crash mid-destroy during baseline resync:
  If a crash occurs after pg_destroy marks the PG destroyed but before the superblk is removed, log replay on recovery would attempt to write to the destroyed index table. This is safe
  because m_rd_sb->last_snapshot_lsn is persisted before pg_destroy is triggered, so all logs at or before that LSN are skipped on recovery (see raft_repl_dev::need_skip_processing). Added a
  guard in on_log_replay_done to return early when the PG is in DESTROYED state with a live repl_dev, letting the next snapshot message re-trigger pg_destroy.

  Additional fixes:
  - Simplified _exit_pg to rely on permanent_destroy → on_destroy → pg_destroy instead of manually finding and destroying the PG
  - Made repl_dev_ checks conditional in destroy_pg_superblk to handle null repl_dev
  - Renamed can_chunks_in_pg_be_gc → is_pg_alive for clarity
  - Fixed PGMetrics to use data_service().get_blk_size() instead of dereferencing potentially-null repl_dev_
  - Reset latch LSN and no-space-left error state when destroying a PG during baseline resync

### DIFF
--- a/src/lib/homestore_backend/hs_pg_manager.cpp
+++ b/src/lib/homestore_backend/hs_pg_manager.cpp
@@ -688,18 +688,7 @@ std::optional< pg_id_t > HSHomeObject::get_pg_id_with_group_id(group_id_t group_
 
 void HSHomeObject::_destroy_pg(pg_id_t pg_id) { pg_destroy(pg_id); }
 
-bool HSHomeObject::pg_destroy(pg_id_t pg_id, bool need_to_pause_pg_state_machine) {
-    if (need_to_pause_pg_state_machine && !pause_pg_state_machine(pg_id)) {
-        LOGI("Failed to pause pg state machine, pg_id={}", pg_id);
-        return false;
-    }
-    LOGI("Destroying pg={}", pg_id);
-    mark_pg_destroyed(pg_id);
-
-    // we have the assumption that after pg is marked as destroyed, it will not be marked as alive again.
-    // TODO:: if this assumption is broken, we need to handle it.
-    gc_mgr_->drain_pg_pending_gc_task(pg_id);
-
+void HSHomeObject::destroy_pg_resource(pg_id_t pg_id) {
     destroy_shards(pg_id);
     destroy_hs_resources(pg_id);
     destroy_pg_index_table(pg_id);
@@ -709,8 +698,31 @@ bool HSHomeObject::pg_destroy(pg_id_t pg_id, bool need_to_pause_pg_state_machine
     // which must be done after destroying pg super blk to avoid multiple pg use same chunks
     bool res = chunk_selector_->return_pg_chunks_to_dev_heap(pg_id);
     RELEASE_ASSERT(res, "Failed to return pg={} chunks to dev_heap", pg_id);
+    LOGI("resource of pg={} is destroyed", pg_id);
+}
 
-    LOGI("pg={} is destroyed", pg_id);
+bool HSHomeObject::pg_destroy(pg_id_t pg_id, bool need_to_pause_pg_state_machine) {
+    // Baseline resync concern: if pg_destroy partially completes before a crash (e.g., the index
+    // table is destroyed but the PG superblk is not), log replay on recovery might attempt to write
+    // to the destroyed index table and fail.
+    //
+    // This is not an issue. Before pg_destroy is triggered during baseline resync,
+    // m_rd_sb->last_snapshot_lsn is persisted to snapshot.get_last_log_idx(). Any log at or
+    // before that LSN is skipped on recovery and never replayed.
+    //
+    // See raft_repl_dev::need_skip_processing for details.
+    if (need_to_pause_pg_state_machine && !pause_pg_state_machine(pg_id)) {
+        LOGI("Failed to pause pg state machine, pg_id={}", pg_id);
+        return false;
+    }
+
+    LOGI("Destroying pg={}", pg_id);
+    mark_pg_destroyed(pg_id);
+
+    // we have the assumption that after pg is marked as destroyed, it will not be marked as alive again.
+    // TODO:: if this assumption is broken, we need to handle it.
+    gc_mgr_->drain_pg_pending_gc_task(pg_id);
+    destroy_pg_resource(pg_id);
     return true;
 }
 
@@ -724,29 +736,8 @@ PGManager::NullResult HSHomeObject::_exit_pg(uuid_t group_id, peer_id_t peer_id,
         LOGI("group_id is nil, nothing to exit, trace_id={}", tid);
         return folly::makeUnexpected(PGError::INVALID_ARG);
     }
-    pg_id_t pg_id{0};
-    {
-        auto lg = std::shared_lock(_pg_lock);
-        auto iter = std::find_if(_pg_map.begin(), _pg_map.end(), [group_id](const auto& entry) {
-            return pg_repl_dev(*entry.second).group_id() == group_id;
-        });
-        if (iter != _pg_map.end()) {
-            pg_id = iter->first;
-        } else {
-            // There is a known case during adding member: the new member may think itself already in group but actually
-            // not, so the pg is not created yet.
-            LOGI("no pg found, group_id={}, trace_id={}", group_id, tid);
-        }
-    }
-    if (pg_id != 0 && !pg_destroy(pg_id)) {
-        // don't need to pause state machine here, this api is called during member leaving or the member is not in the
-        // cluster actually.
-        LOGE("failed to destroy pg={}, group_id={}, trace_id={}", pg_id, group_id, tid);
-        return folly::makeUnexpected(PGError::UNKNOWN);
-    }
-    LOGI("pg is cleaned, going to destroy repl_dev, group_id={}, trace_id={}", group_id, tid);
-    // TODO pass peer_id into destroy_repl_dev for peer validation
-    // destroy_repl_dev will leave raft group
+
+    // mark pg as destoryed and then permanent_destroy will call destory_pg to reclaim pg resource.
     auto ret = hs_repl_service().destroy_repl_dev(group_id);
     if (ret == ReplServiceError::SERVER_NOT_FOUND) {
         LOGW("repl dev not found, ignore, group_id={}, trace_id={}", group_id, tid);
@@ -756,6 +747,7 @@ PGManager::NullResult HSHomeObject::_exit_pg(uuid_t group_id, peer_id_t peer_id,
         LOGE("Failed to destroy repl dev for group_id={}, error={}, trace_id={}", group_id, ret, tid);
         return folly::makeUnexpected(toPgError(ret));
     }
+
     return folly::Unit();
 }
 
@@ -800,7 +792,7 @@ void HSHomeObject::mark_pg_destroyed(pg_id_t pg_id) {
     LOGD("pg={} is marked as destroyed", pg_id);
 }
 
-bool HSHomeObject::can_chunks_in_pg_be_gc(pg_id_t pg_id) const {
+bool HSHomeObject::is_pg_alive(pg_id_t pg_id) const {
     auto lg = std::scoped_lock(_pg_lock);
     auto hs_pg = const_cast< HS_PG* >(_get_hs_pg_unlocked(pg_id));
     if (hs_pg == nullptr) {
@@ -859,7 +851,11 @@ void HSHomeObject::destroy_pg_superblk(pg_id_t pg_id) {
         }
 
         hs_pg->pg_sb_.destroy();
-        destroy_snapshot_sb(hs_pg->repl_dev_->group_id());
+
+        // FIXME:: if repl_dev does not exist, how to get group_id to destory snapshot sb? we should store group_id in
+        // pg superblk to avoid this issue.
+        if (hs_pg->repl_dev_) { destroy_snapshot_sb(hs_pg->repl_dev_->group_id()); }
+
         hs_pg->snp_rcvr_info_sb_.destroy();
         hs_pg->snp_rcvr_shard_list_sb_.destroy();
 
@@ -871,9 +867,12 @@ void HSHomeObject::destroy_pg_superblk(pg_id_t pg_id) {
 }
 
 void HSHomeObject::add_pg_to_map(unique< HS_PG > hs_pg) {
-    RELEASE_ASSERT(hs_pg->pg_info_.replica_set_uuid == hs_pg->repl_dev_->group_id(),
-                   "PGInfo replica set uuid mismatch with ReplDev instance for {}",
-                   boost::uuids::to_string(hs_pg->pg_info_.replica_set_uuid));
+    if (hs_pg->repl_dev_) {
+        RELEASE_ASSERT(hs_pg->pg_info_.replica_set_uuid == hs_pg->repl_dev_->group_id(),
+                       "PGInfo replica set uuid mismatch with ReplDev instance for {}",
+                       boost::uuids::to_string(hs_pg->pg_info_.replica_set_uuid));
+    }
+
     auto lg = std::scoped_lock(_pg_lock);
     auto id = hs_pg->pg_info_.id;
     auto [it1, _] = _pg_map.try_emplace(id, std::move(hs_pg));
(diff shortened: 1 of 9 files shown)


src/lib/homestore_backend/hs_pg_manager.cpp
```diff
@@ -709,8 +698,35 @@ bool HSHomeObject::pg_destroy(pg_id_t pg_id, bool need_to_pause_pg_state_machine
...
+bool HSHomeObject::pg_destroy(pg_id_t pg_id) {
+    auto hs_pg = const_cast< HS_PG* >(get_hs_pg(pg_id));
+    RELEASE_ASSERT(hs_pg, "pg={} is null", pg_id);
+    auto repl_dev = hs_pg ? hs_pg->repl_dev_ : nullptr;
+    RELEASE_ASSERT(repl_dev, "repl_dev for pg={} is null", pg_id);
+
+    // when reaching here, we will not receive any new log, both for BR or leave_group case. we wait for all logs to be
+    // committed and trigger a cp, so that when restart, no log will be replayed since cp_lsn is equal to
+    // last_append_lsn.
+    while (true) {
```
xiaoxichen: I feel this while loop is probably wrong.
if this member receive a leave_group, who will tell it to commit all appened log?
similar in BR, if there are some appended but not yet committed log in logstore, but leader decide to send us a BR,  probably leader wont tell us to commit

xiaoxichen: what is the intent here? 
avoid the ongoing commit thread racing with destroying? 
crash recovery?

src/lib/homestore_backend/replication_state_machine.cpp
```diff
@@ -449,6 +449,10 @@ void ReplicationStateMachine::write_snapshot_obj(std::shared_ptr< homestore::sna
         set_snapshot_context(context); // Update the snapshot context in case apply_snapshot is not called
         auto hs_pg = home_object_->get_hs_pg(m_snp_rcv_handler->get_context_pg_id());
         hs_pg->pg_state_.clear_state(PGStateMask::BASELINE_RESYNC);
+        // we only reset this if destroying pg happens in BR case. for other cases (on_destroy and _exit_pg),
```
xiaoxichen, line 452: any drawback if we clean up in both cases?

It is unlikely but what if we move PG X from SM A->SM B ->SM A, will the leftovers bite us?

xiaoxichen: also even we want to do so, why we do that in is_last_obj instead of first_obj?

src/lib/homestore_backend/replication_state_machine.cpp
```diff
@@ -499,7 +503,7 @@ void ReplicationStateMachine::write_snapshot_obj(std::shared_ptr< homestore::sna
         if (home_object_->pg_exists(pg_data->pg_id())) {
             LOGI("pg already exists, clean pg resources before snapshot, pg={} {}", pg_data->pg_id(), log_suffix);
             // Need to pause state machine before destroying the PG, if fail, let raft retry.
```
xiaoxichen, line 505: comments out of date, as well as we dont have  a branch that returns false as of now.

src/lib/homestore_backend/hs_pg_manager.cpp
```diff
@@ -709,8 +698,35 @@ bool HSHomeObject::pg_destroy(pg_id_t pg_id, bool need_to_pause_pg_state_machine
     // which must be done after destroying pg super blk to avoid multiple pg use same chunks
     bool res = chunk_selector_->return_pg_chunks_to_dev_heap(pg_id);
     RELEASE_ASSERT(res, "Failed to return pg={} chunks to dev_heap", pg_id);
+    LOGI("resource of pg={} is destroyed", pg_id);
+}
+
+bool HSHomeObject::pg_destroy(pg_id_t pg_id) {
+    auto hs_pg = const_cast< HS_PG* >(get_hs_pg(pg_id));
+    RELEASE_ASSERT(hs_pg, "pg={} is null", pg_id);
+    auto repl_dev = hs_pg ? hs_pg->repl_dev_ : nullptr;
```
xiaoxichen: the assert ensures hs_pg not null?

src/lib/homestore_backend/replication_state_machine.cpp
```diff
@@ -1030,7 +1034,33 @@ void ReplicationStateMachine::on_log_replay_done(const homestore::group_id_t& gr
...
+        // RaftReplDev::leave() will be called in the following scenarios:
+        //    a. destroy_repl_dev: commit log entry of journal_type_t::HS_CTRL_DESTROY
+        //    b. removed from raft group: the move_out member in replace member.
+
+        // for baseline resync, no need to redo destroy pg here since the first snapshot message will be received again
+        // and trigger pg_destroy again if the pg already exists. but for other cases, we need to redo destory pg to
+        // clean up the stale pg resources since no message will be received to trigger pg_destroy again.
+
+        // so, we call it here for all the above cases to make sure the stale pg resources are cleaned up.
+        home_object_->destroy_pg_resource(pg_id);
```
xiaoxichen: do we need cleaning GC resources? like any pending gc metablk?  gc index table? no_space_left_marker? latch_lsn?

src/lib/homestore_backend/hs_pg_manager.cpp
JacksonYao287: > what is the intent here?
avoid the ongoing commit thread racing with destroying?
crash recovery?

both. 
1 currently , we pause statemachine here to avoid the ongoing commit thread racing with destroying pg.  here,  I try to achieve this by waiting for all the appended log to be committed, which means no more log will not be commit anymore . as a result, there is  no racing between committing log and destroying pg.

2 crash recovery. after all the appended log are committed and a cp is triggered, when recovery, no log replay will happen(cp_lsn == last_append_lsn), so that we can avoid log replay to access any stale pg resource. for example, we have a put blob log, whose lsn is larger than cp_lsn but smaller than dc_lsn.  if destroy_pg is called and the pg_index_Table is destroyed but before pg_super_blk is destoryed, crash occures, Then, when recovery and replay this put blob log, it will try to access the destroyed pg_index_table.  other pg resources also has this potential issue(shard or pg chunk)


>if this member receive a leave_group, who will tell it to commit all appened log?
similar in BR, if there are some appended but not yet committed log in logstore, but leader decide to send us a BR, probably leader wont tell us to commit

let`s say, leader starts from lsn 30(lsn 1  - lsn 29 are truncated), and the last append log in follower is lsn 20 and last commit log in follower is 10. there is an important fact that **any truncated log must be already committed before it is truncated**.  so the commit_lsn of the whole raft group is at least lsn 29.

leader will start sending snapshot to follower only if it sends an append_entry req to follower and the next_lsn (in follower resp) follower wants is smaller than the smallest lsn leader currently has. note that, in the append_entry req, follower can see the latest commit_lsn in the raft group(at least lsn 29), which must be greater than the follower`s last_append_lsn（lsn 20）.

to answer your question, the append_entry req , which will trigger install_snapshot, will tell the laggy follower what is the latest commit_lsn currently. and the latest commit_lsn must be greater than the last_append_lsn in follower , which can make sure all the appended logs to be committed.

src/lib/homestore_backend/replication_state_machine.cpp
JacksonYao287: >any drawback if we clean up in both cases? 

no specific drawback. in my mind, no more log is expected to be received for on_destroy case, so no necessity 

>It is unlikely but what if we move PG X from SM A->SM B ->SM A, will the leftovers bite us?
I thinks no. 

if A is the move_out member ( SM A->SM B), the repl_dev, as well as the statemachine attached to this repl_dev, will be destroyed in on_destroy(remove_member).
 
 when we move from SM B ->SM A again, new repl_dev (and statemachine)will be created for the pg, since SM A is now a new member joining this raft group.
 
 >also even we want to do so, why we do that in is_last_obj instead of first_obj?

 actually I do it at first_obj in my first version of this PR. since when snapshot is on going , no new log will be received,  there is no specific difference between do this in is_last_obj and first_obj.  
 
 I put it here means before this member is ready to accept new logs, clear all the stale state for it.

src/lib/homestore_backend/replication_state_machine.cpp
JacksonYao287: let`s remove this out-of-date comments after addressing other comments for this PR

src/lib/homestore_backend/hs_pg_manager.cpp
JacksonYao287: yes. let`s see if there is any case we will destroy an unknown pg

src/lib/homestore_backend/replication_state_machine.cpp
JacksonYao287: gc index table is pdev related , not pg. no_space_left_marker and  latch_lsn has already done in this PR, see reset_latch_lsn, reset_no_space_left_inf.

the gc metablk is a nice catch. there is indeed a case that a gc metablk left when crash. let`s do it in handle_recovered_gc_task.  if we find the pg state is destroyed, skip this task and destroy this gc meta blk.

src/lib/homestore_backend/hs_pg_manager.cpp
xiaoxichen: >let`s say, leader starts from lsn 30(lsn 1 - lsn 29 are truncated), and the last append log in follower is lsn 20 and last commit log in follower is 10. there is an important fact that any truncated log must be already committed before it is truncated. so the commit_lsn of the whole raft group is at least lsn 29.

This is not necessary true, we keep maximum `raft_logstore_reserve_threshold` logs, 
```
    repl_lsn_t raft_logstore_reserve_threshold = HS_DYNAMIC_CONFIG(resource_limits.raft_logstore_reserve_threshold);
    repl_lsn_t truncation_upper_limit = std::max(leader_commit_idx - raft_logstore_reserve_threshold, minimum_repl_idx);
 
```
----------------

>to answer your question, the append_entry req , which will trigger install_snapshot, will tell the laggy follower what is the latest commit_lsn currently. and the latest commit_lsn must be greater than the last_append_lsn in follower , which can make sure all the appended logs to be committed.


This part  is wrong as well,  say if a member has log up to 150, commited to 100, however it gets a install_snapshot_req, it wont know the commit lsn of leader(say 50000) , and it wont commit locally to 150.

src/lib/homestore_backend/hs_pg_manager.cpp
xiaoxichen: no I mean , we alredy assert the hs_pg is not null in L706, then in L707 we dont need checking the hs_pg == null

xiaoxichen: it is a nit, resolving it for now to keep the discussion clean

src/lib/homestore_backend/replication_state_machine.cpp
JacksonYao287: after checking , no need to handle gc metablk.

if there is any gc metablk, it means before crash, drain_pending_gc_task is not returned yet(and thus no pg resource is destroyed). when recovery , these gc metablk will be handle by handle_recovered_gc_task before starting log replay.

src/lib/homestore_backend/hs_pg_manager.cpp
JacksonYao287: raft_logstore_reserve_threshold does not matter , only last_append_lsn and last_commit_lsn matters.

>however it gets a install_snapshot_req

in what case it will get a install_snapshot_req?  as I said , the only case is that leader sends an append_entry req to follower and the next_lsn in follower resp, which indicates the next log follower wants,  is smaller than the smallest lsn leader currently has.  does this make sense?

the append_entry will have a commit_index , which indicates the lsn that the follower can commit to. commit lsn must be larger than followers last_append_lsn.

if install_snapshot is triggered, then
leader_commit_lsn >= leader_smallest_lsn > follower_last_append_lsn >= follower_commit_lsn

xiaoxichen: IIRC , a rejected `append_log_request` would not goes to the bottom and update the target_commit_index of the follower.

src/lib/homestore_backend/replication_state_machine.cpp
```diff
@@ -1030,7 +1034,42 @@ void ReplicationStateMachine::on_log_replay_done(const homestore::group_id_t& gr
...
+        // here for all the above cases to make sure the stale pg resources are cleaned up.
+
+        // there is also a concern that in baseline resync case, if some resource is destroyed in pg_destroy but not all
+        // the resources are destroyed before crash, then when recovery, log replay will hit those destroyed resources
+        // and cause error. for example, we have a put_blob log after cp_lsn and before dc_lsn, and the pg_index_table
+        // is destory but before pg_super_blk is destroyed, crash happens. when recovery, log replay will hit the
+        // put_blob log and try to write to the index table, but since the index table is destroyed, it will cause
+        // error. but actually, this is not a problem. since before we starting pg_destroy in baseline resync,
+        // m_rd_sb->last_snapshot_lsn will be persisted upto the snapshot.get_last_log_idx(). then all the log less than
+        // or equal to m_rd_sb->last_snapshot_lsn will not be replayed or committed after recovery. so, the concern is
```
JacksonYao287: @xiaoxichen in baseline resync case , before we start destroying pg, we will m_rd_sb->last_snapshot_lsn upto snapshot.get_last_log_idx(). then raft_repl_dev#need_skip_processing will help us skipping replaying all the logs in recovery path(so that we will not hit those destroyed resources , like pg_index_table, etc.). so we don`t need wait for all the appended log to be committed in pg_destroy for BR case

xiaoxichen: so basically you reverted your changes

xiaoxichen: I dont get the point of this comments, you said for br no need to redo destroy PG but we do it anyway.  

The concern of log replay vs destroy is not valid here ... if we reach here the log replay had been done...If we want to record the thinking why waiting for log commit is not needed, better rephrase this paragraph and move it to `destroy_pg `


Similar for L1043-1052,  those lines explains the situations that a PG can be destroy,  better to move to `destroy_pg` rather than here, especially we use same action in recovery path , for all source

xiaoxichen (dismissed):
LGTM aside from inline comments cleanup.

Try to use LLM to polish the language a bit.

JacksonYao287:
I need to rethink the implementation here again , seems I miss some repl_dev details.

JacksonYao287:
let`s me  explain redo destroy pg with more details. Bassically,  the destoryed pg can be ‌categorized into two case:

**1 RaftReplDev::leave() is not called and thus m_rd_sb->destroy_pending = 0x0:** only BR belongs to this case. in this case repl_dev will be recovered, log will be replayed.  For this case , we need do nothing for redoing destroy pg, no matter is pg state is alive or destroyed. the reason is that if the pg super blk exists, it means the first snapshot message (obj_id.shard_seq_num == 0) is not successfully handled and crash happens before pg super blk is destroyed . Then, when recovery, leader will resend the first snapshot message and follower will handle this message again and call pg_destroy to redo pg destory.

**2 RaftReplDev::leave() is called and thus m_rd_sb->destroy_pending = 0x1**: remove_member, destroy_repl_dev and  destroy_raft_group all belong to this case. if m_rd_sb->destroy_pending = 0x1, [this repl_dev will not be loaded and no log replay.](https://github.com/eBay/HomeStore/blob/27aed5f47ba5ff4636309fdced2fd1aa67823f97/src/lib/replication/service/raft_repl_service.cpp#L477-L495). so permenant_destroy will not see this repl_dev,  but the pg resource are probably still there if crash happens. so  we need to redo pg destory to carefully reclaim them.

@xiaoxichen ptal

xiaoxichen:
> **1 RaftReplDev::leave() is not called and thus m_rd_sb->destroy_pending = 0x0:** only BR belongs to this case. in this case repl_dev will be recovered, log will be replayed. For this case , we need do nothing for redoing destroy pg, no matter is pg state is alive or destroyed. the reason is that if the pg super blk exists, it means the first snapshot message (obj_id.shard_seq_num == 0) is not successfully handled and crash happens before pg super blk is destroyed . Then, when recovery, leader will resend the first snapshot message and follower will handle this message again and call pg_destroy to redo pg destory.

any possibility it is case 2 but crashed just before setting the destroy_pending = 0x1?

JacksonYao287:
> any possibility it is case 2 but crashed just before setting the destroy_pending = 0x1?

yes , of course, but it does not matter, since pg_destory is always called after RaftReplDev::leave() is successfully called (it is called in [permenant_destroy](https://github.com/eBay/HomeStore/blob/ef9ab9d0c5a8061b59719f7cfbcf69ae0a64f60a/src/lib/replication/repl_dev/raft_repl_dev.cpp#L2317-L2329)  in gc_repl_devs()#leave_group) and [RaftReplDev::leave()](https://github.com/eBay/HomeStore/blob/ef9ab9d0c5a8061b59719f7cfbcf69ae0a64f60a/src/lib/replication/repl_dev/raft_repl_dev.cpp#L2331-L2356) is the only way to change destroy_pending to 0x1.


let`s discuss case by case:

1 destroy_raft_group: will propose a log with type of HS_CTRL_DESTROY, and `leave()` will be called in `on_commit` of this log. if crash happens before changing destroy_pending to 0x1, when recovery, repl_dev will be recovered(destroy_pending is still 0x0) and this log will be commit again, and `leave()` will be called again and then permenant_destroy will call pg_destory to reclaim pg resource.

2 destroy_repl_dev(exit_pg): [will directly call leave()](https://github.com/eBay/HomeStore/blob/ef9ab9d0c5a8061b59719f7cfbcf69ae0a64f60a/src/lib/replication/service/raft_repl_service.cpp#L612). so if crash happens before changing destroy_pending to 0x1, nothing happens since now in exit_pg , what we do is only changing destroy_pending to 0x1 , no pg resource will be destory there.

3 remove_from_cluster:  [will call leave() when receiving ](https://github.com/eBay/nuraft_mesg/blob/251ecb4dce1864091d098739c7b67a5b2dd6a0da/src/lib/manager_impl.cpp#L145-L149) a RemovedFromCluster. so if  crash happens before changing destroy_pending to 0x1, then after recovery, [it will received RemovedFromCluster again](https://github.com/eBay/HomeStore/blob/ef9ab9d0c5a8061b59719f7cfbcf69ae0a64f60a/src/lib/replication/repl_dev/raft_repl_dev.cpp#L2334-L2339).

as a summary,  if it is case 2 but crashed just before setting the destroy_pending = 0x1, pg resource will not be touched and everything will go well

xiaoxichen (approved):
lgtm

PR #436, opened 2026-06-17
SDSTOR-22783: priority scheduling with dual watermarks and cross-scan quota

xiaoxichen:
- Sort eligible chunks by garbage ratio (desc) before submission so the most garbage-heavy chunks are always GC'd first
- Add gc_garbage_rate_threshold_low (default 30%) as a low watermark; chunks between the two watermarks consume at most half the quota
- Track m_pending_normal_gc_task_count in pdev_gc_actor to reflect tasks queued or running in m_gc_executor across scan cycles; previous code only capped submissions per scan, allowing unbounded queue growth
- scan_chunks_for_gc now skips a pdev entirely when already at quota, and derives low_tier_cap proportionally from remaining_capacity
- Add ADR docs/adr/gc-priority-scheduling.md

### DIFF
--- a/src/lib/homestore_backend/gc_manager.cpp
+++ b/src/lib/homestore_backend/gc_manager.cpp
@@ -1,3 +1,6 @@
+#include <algorithm>
+#include <queue>
+
 #include <homestore/btree/btree_req.hpp>
 #include <homestore/btree/btree_kv.hpp>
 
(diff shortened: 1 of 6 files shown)


src/lib/homestore_backend/gc_manager.cpp
```diff
@@ -1225,6 +1308,10 @@ void GCManager::pdev_gc_actor::on_gc_task_completed(uint8_t priority, pg_id_t pg
     const auto final_state =
         priority == static_cast< uint8_t >(task_priority::normal) ? ChunkState::AVAILABLE : ChunkState::INUSE;
 
+    if (priority == static_cast< uint8_t >(task_priority::normal)) {
```
JacksonYao287: should not put this here, should put it into the destructor of gc_task_guard

on_gc_task_completed will also be called in handle_recoverd_gc_task , where we should not decrease any counter.

pls refer to where decr_pg_pending_gc_task is called.

src/lib/homestore_backend/gc_manager.cpp
```diff
@@ -246,38 +248,118 @@ bool GCManager::is_eligible_for_gc(chunk_id_t chunk_id) {
...
+        // Collect all chunks eligible for GC (above the low watermark) along with their garbage ratios.
+        // Chunks below gc_thresh_low are not worth scheduling given the available quota.
+        struct ChunkGCInfo {
+            chunk_id_t chunk_id;
+            // integer percentage [0,100]; computed as (100*defrag_blks)/total_blks
+            uint32_t garbage_ratio_pct;
+        };
+        std::vector< ChunkGCInfo > eligible;
         for (const auto& chunk_id : chunks) {
-            if (is_eligible_for_gc(chunk_id)) {
```
JacksonYao287, line 258: > is_eligible_for_gc

will this be used again? if not, we can remove it

src/lib/homestore_backend/gc_manager.cpp
```diff
@@ -246,38 +248,118 @@ bool GCManager::is_eligible_for_gc(chunk_id_t chunk_id) {
...
+            const uint32_t ratio_pct = static_cast< uint32_t >((100 * defrag_blk_num) / total_blk_num);
+            LOGDEBUGMOD(gcmgr,
+                        "gc scan chunk_id={}, use_blks={}, available_blks={}, total_blks={}, defrag_blks={}, "
+                        "garbage_ratio_pct={}",
+                        chunk_id, chunk->get_used_blks(), chunk->available_blks(), total_blk_num, defrag_blk_num,
+                        ratio_pct);
+            eligible.push_back({chunk_id, ratio_pct});
+        }
+
+        // Sort eligible chunks by garbage ratio descending so the most garbage-heavy chunks are GC'd first.
```
JacksonYao287: this is a topK problem. 
suggest to use a heap with a capacity of remaining_capacity (priority_queue), so that we can hold at most remaining_capacity ChunkGCInfo  in memory. and for any new ChunkGCInfo, we only need to compare it with the top of the heap.

src/lib/homestore_backend/gc_manager.cpp
```diff
@@ -246,38 +248,118 @@ bool GCManager::is_eligible_for_gc(chunk_id_t chunk_id) {
...
+        LOGERRORMOD(gcmgr,
+                    "gc_garbage_rate_threshold_low={} exceeds gc_garbage_rate_threshold={}, "
+                    "auto-correcting low to {}",
+                    gc_thresh_low, gc_thresh_high, gc_thresh_high / 2);
+        gc_thresh_low = gc_thresh_high / 2;
+    }
 
     for (const auto& [pdev_id, chunks] : m_chunk_selector->get_pdev_chunks()) {
-        auto max_task_num = 2 * (reserved_chunk_num_per_pdev - reserved_chunk_num_per_pdev_for_egc);
+        const uint32_t max_task_num = 2 * (reserved_chunk_num_per_pdev - reserved_chunk_num_per_pdev_for_egc);
```
JacksonYao287, line 272: here is a fact that :

**the higher the garbage ratio of a chunk is , the less blobs will be copied in the gc process of this chunk, and thus will bring less io pressure to hardware.**

so , in my mind, assuming we limit the io pressure of gc in a certain period，we can submit more gc tasks if the candidate chunks has higher  ratio. I don`t have a clear idea, but just think can max_task_num be optimized to a more flexible  value?

xiaoxichen: The maximum conccurrent is limit by reserved_chunk_num_per_pdev.    It is more on how frequent we do the scanning and submits GC task.  Utimiately assuming this is a busy loop then we always keep the Q at 2 * reserved_chunks.

However i think the tuning target of GC (in most of the time) is not maximum GC speed,  it is enough to catch up with data rotation (or maybe 2x of data rotation rate).

src/lib/homestore_backend/gc_manager.cpp
xiaoxichen: IMO It doesnt worth those lines of code ,   considering the maximum chunk is 32K.

src/lib/homestore_backend/gc_manager.cpp
xiaoxichen: good call, should we move can_chunks_in_pg_be_gc into `int is_eligible_for_gc`  (returns the garbage_ratio_ptc) ?

xiaoxichen: address

src/lib/homestore_backend/gc_manager.cpp
xiaoxichen: addressed

src/lib/homestore_backend/gc_manager.cpp
JacksonYao287, line 1364: also add m_gc_actor->m_pending_normal_gc_task_count.fetch_sub here

src/lib/homestore_backend/gc_manager.cpp
```diff
@@ -388,6 +462,7 @@ folly::SemiFuture< bool > GCManager::pdev_gc_actor::add_gc_task(uint8_t priority
                 LOGDEBUGMOD(gcmgr, "start gc task : move_from_chunk_id={}, priority={}", move_from_chunk, priority);
                 process_gc_task(move_from_chunk, priority, std::move(promise), gc_task_id);
             });
+            m_pending_normal_gc_task_count.fetch_add(1, std::memory_order_relaxed);
```
JacksonYao287: it will be better if m_pending_normal_gc_task_count.fetch_add is called before  m_gc_executor->add.

for example :

T1: m_pending_normal_gc_task_count == 0, m_gc_executor->add

T2: the added task is immediately completed before  m_pending_normal_gc_task_count.fetch_add, so m_pending_normal_gc_task_count will integer underflow。

this is one line change, we`d better not let underflow happen

src/lib/homestore_backend/gc_manager.cpp
JacksonYao287: ok, let`s keep it like this, and revisit here if we really need it in the future

src/lib/homestore_backend/gc_manager.cpp
JacksonYao287: better to have it  since it is not complicated and will also make code more simple, for example, the loop on the entire chunk collection will not be involved.

xiaoxichen:
@JacksonYao287 addressed

JacksonYao287 (dismissed):

JacksonYao287 (approved):

PR #438, opened 2026-06-23
SDSTOR-22421: shard superblk backward compat for log-only shard refactor

Hooper9973:
- Migrate v2 shard superblk (sb_version=0x02) to v3 on recovery; unify decode logic into decode_shard_sb() shared by raft-log path and on_shard_meta_blk_found

- Decode old shard raft-log messages (v2 header_extn) on new code path; extend ResyncShardMetaData schema with sealed_lsn at end for old-buffer compat

- Override get_blk_alloc_hint for CREATE/SEAL_SHARD to return committed-blk hint, skipping data-block allocation for log-only shard messages

### DIFF
--- a/src/include/homeobject/shard_manager.hpp
+++ b/src/include/homeobject/shard_manager.hpp
@@ -35,14 +35,14 @@ struct ShardInfo {
     shard_id_t id;
     pg_id_t placement_group;
     State state;
-    uint64_t lsn;                   // created_lsn
-    uint64_t sealed_lsn{INT64_MAX}; // sealed_lsn
+    uint64_t lsn; // created_lsn
     uint64_t created_time;
     uint64_t last_modified_time;
     uint64_t available_capacity_bytes;
     uint64_t total_capacity_bytes;
     std::optional< peer_id_t > current_leader{std::nullopt};
     uint8_t meta[meta_length]{};
+    uint64_t sealed_lsn{INT64_MAX}; // appended after all v2 fields for forward compat
 
     auto operator<=>(ShardInfo const& rhs) const { return id <=> rhs.id; }
     auto operator==(ShardInfo const& rhs) const { return id == rhs.id; }
@@ -58,7 +58,8 @@ public:
 
     virtual AsyncResult< ShardInfo > get_shard(shard_id_t id, trace_id_t tid = 0) const = 0;
     virtual AsyncResult< InfoList > list_shards(pg_id_t id, trace_id_t tid = 0) const = 0;
-    virtual AsyncResult< ShardInfo > create_shard(pg_id_t pg_owner, uint64_t size_bytes, std::string meta, trace_id_t tid = 0) = 0;
+    virtual AsyncResult< ShardInfo > create_shard(pg_id_t pg_owner, uint64_t size_bytes, std::string meta,
+                                                  trace_id_t tid = 0) = 0;
     virtual AsyncResult< ShardInfo > seal_shard(shard_id_t id, trace_id_t tid = 0) = 0;
 };
 
--- a/src/include/homeobject/shard_manager.hpp
+++ b/src/include/homeobject/shard_manager.hpp
@@ -35,7 +35,7 @@ struct ShardInfo {
     shard_id_t id;
     pg_id_t placement_group;
     State state;
-    uint64_t lsn; // created_lsn
+    uint64_t create_lsn;
     uint64_t created_time;
     uint64_t last_modified_time;
     uint64_t available_capacity_bytes;
--- a/src/lib/homestore_backend/hs_shard_manager.cpp
+++ b/src/lib/homestore_backend/hs_shard_manager.cpp
@@ -31,7 +31,7 @@ static ShardInfo shardinfo_from_v2(const HSHomeObject::v2_ShardInfo& v2) {
     info.id = v2.id;
     info.placement_group = v2.placement_group;
     info.state = v2.state;
-    info.lsn = v2.lsn;
+    info.create_lsn = v2.lsn;
     info.sealed_lsn = INT64_MAX;
     info.created_time = v2.created_time;
     info.last_modified_time = v2.last_modified_time;
@@ -46,7 +46,7 @@ static ShardInfo shardinfo_from_v2(const HSHomeObject::v2_ShardInfo& v2) {
 // Handles all supported on-disk versions. Returns nullopt for unknown versions.
 static std::optional< std::tuple< ShardInfo, homestore::chunk_num_t, homestore::chunk_num_t > >
 decode_shard_sb(const HSHomeObject::shard_info_superblk* sb) {
-    if (sb->sb_version >= HSHomeObject::shard_info_superblk::shard_sb_version) {
+    if (sb->sb_version == HSHomeObject::shard_info_superblk::shard_sb_version) {
         return std::make_tuple(sb->info, sb->p_chunk_id, sb->v_chunk_id);
     } else if (sb->sb_version == 0x02) {
         const auto* v2sb = r_cast< const HSHomeObject::v2_shard_info_superblk* >(sb);
@@ -122,7 +122,7 @@ std::string HSHomeObject::serialize_shard_info(const ShardInfo& info) {
     j["shard_info"]["shard_id_t"] = info.id;
     j["shard_info"]["pg_id_t"] = info.placement_group;
     j["shard_info"]["state"] = info.state;
-    j["shard_info"]["lsn"] = info.lsn;
+    j["shard_info"]["lsn"] = info.create_lsn;
     j["shard_info"]["sealed_lsn"] = info.sealed_lsn;
     j["shard_info"]["created_time"] = info.created_time;
     j["shard_info"]["modified_time"] = info.last_modified_time;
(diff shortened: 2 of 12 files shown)


src/lib/homestore_backend/hs_shard_manager.cpp
```diff
@@ -26,6 +26,37 @@ SISL_LOGGING_DECL(shardmgr)
...
+    info.current_leader = v2.current_leader;
+    std::memcpy(info.meta, v2.meta, ShardInfo::meta_length);
+    return info;
+}
+
+// Core decode: given a typed shard_info_superblk pointer, extract ShardInfo + chunk IDs.
+// Handles all supported on-disk versions. Returns nullopt for unknown versions.
+static std::optional< std::tuple< ShardInfo, homestore::chunk_num_t, homestore::chunk_num_t > >
+decode_shard_sb(const HSHomeObject::shard_info_superblk* sb) {
+    if (sb->sb_version >= HSHomeObject::shard_info_superblk::shard_sb_version) {
```
xiaoxichen: what would be the situation of `sb->sb_version > HSHomeObject::shard_info_superblk::shard_sb_version` ? older version read data written by newer version?

src/include/homeobject/shard_manager.hpp
```diff
@@ -35,14 +35,14 @@ struct ShardInfo {
     shard_id_t id;
     pg_id_t placement_group;
     State state;
-    uint64_t lsn;                   // created_lsn
-    uint64_t sealed_lsn{INT64_MAX}; // sealed_lsn
+    uint64_t lsn; // created_lsn
```
xiaoxichen: maybe rename it as create_lsn?

xiaoxichen (commented):
lgtm in general

src/lib/homestore_backend/replication_state_machine.cpp
```diff
@@ -213,13 +212,21 @@ ReplicationStateMachine::get_blk_alloc_hints(sisl::blob const& header, uint32_t
...
+        if (data_size == 0) {
+            // New log-only format: should not reach here, but handle gracefully.
+            LOGW("get_blk_alloc_hints called for log-only shard message type={}, shard={}, pg={}", msg_header->msg_type,
+                 msg_header->shard_id, msg_header->pg_id);
+            return homestore::blk_alloc_hints{};
+        } else {
+            // Old format: shard message carried shard header/footer data blocks. Return committed_blk_id
+            // so HomeStore skips block allocation and data write. Shard on_commit will ignore pbas.
+            homestore::blk_alloc_hints hints;
+            hints.committed_blk_id = homestore::MultiBlkId{homestore::BlkId{0, 1, 0xFFFF}};
```
xiaoxichen: make this as a global constant (like tombstone_pba) so other pieces of code is easier to check/assert based on this value.

src/lib/homestore_backend/tests/hs_shard_tests.cpp
```diff
@@ -370,35 +372,37 @@ TEST_F(HomeObjectFixture, ShardVersionMigrationRecovery) {
             old_sb->info.available_capacity_bytes = orig_data.info.available_capacity_bytes;
             old_sb->info.total_capacity_bytes = orig_data.info.total_capacity_bytes;
             old_sb->info.current_leader = orig_data.info.current_leader;
-            // Note: v1 doesn't have meta field, so we don't copy it
+            std::memcpy(old_sb->info.meta, orig_data.info.meta, ShardInfo::meta_length);
+            // Note: sealed_lsn is not present in v2, so we don't set it
             old_sb->p_chunk_id = orig_data.p_chunk_id;
             old_sb->v_chunk_id = orig_data.v_chunk_id;
             old_sb.write();
+            LOGINFO("Created v2 shard {}", shard_id);
```
JacksonYao287: remove this log, since line 363 has already loged

src/lib/homestore_backend/replication_state_machine.cpp
```diff
@@ -213,13 +212,21 @@ ReplicationStateMachine::get_blk_alloc_hints(sisl::blob const& header, uint32_t
     switch (msg_header->msg_type) {
     case ReplicationMessageType::CREATE_SHARD_MSG:
     case ReplicationMessageType::SEAL_SHARD_MSG: {
```
JacksonYao287, line 214: suggest adding a release_assert(msg_header->msg_type != ReplicationMessageType::CREATE_SHARD_MSG && msg_header->msg_type != ReplicationMessageType::SEAL_SHARD_MSG)

IIUC， it is not expected that new version HO receives a old-version log?

xiaoxichen: it can happen... for example one member is upgraded and leader(not upgraded) sends log to it.

JacksonYao287: in my mind, we will upgrade sm to this new version subcluster by subcluster, and this is a non-rolling upgrade. 

In my mind, for each subcluster :
1 stop writing to this subcluster.
2 check if all the sms in this cluster have committed to the last log in log store.
3 check if all the last commit log of sms in the same pg are identical. 

the aim of above steps is to guarantee that no new logs will be committed and no new logs will be sent from leader to follower.

now, we can safely do the upgrade in this subcluster.

xiaoxichen: as we cannot do the CR,  the step above is very tricky to be executed.  
we can disable this SC from DS so that no more create_shard will happen on this SC.
we can even disable DM, it is a config map update, which stops the create shard/seal shard completely.
but we cannot stop write easily as gateway can still writes to open shard... we cannot stop delete-blob.

with that , no way to guarantee no writes to this SC. your step 2/3 is tricky to execute.

xiaoxichen: it is not 2/3 is impossible,  just to minimize the upgrade effort, if some cases can be handled in code without too much corner cases.

src/lib/homestore_backend/tests/hs_shard_tests.cpp
Hooper9973: since 363 is removed, this line will be retained.

src/lib/homestore_backend/hs_shard_manager.cpp
Hooper9973: `sb_version > shard_sb_version` happens when older SM receive new SM's msg (rolling upgrade). With >=, old code silently parses unknown future struct layout — dangerous if new version changed field offsets or types. Returning nullopt here causes RELEASE_ASSERT at callers, which is more safe than slient unexpected error. 
Changed to ==.

src/lib/homestore_backend/hs_shard_manager.cpp
```diff
@@ -584,56 +623,39 @@ void HSHomeObject::on_shard_message_commit(int64_t lsn, sisl::blob const& h, sha
 }
 
 void HSHomeObject::on_shard_meta_blk_found(homestore::meta_blk* mblk, sisl::byte_view buf) {
-    // First peek at the version
     auto* header = reinterpret_cast< const DataHeader* >(buf.bytes());
+    RELEASE_ASSERT(header->version == DataHeader::data_header_version, "Unknown shard superblock DataHeader version {}",
```
Besroy: Could you use `valid()` here, or add a clear marker so this if can be removed when we upgrade the data version later?  This assert will likely fail during future data version updates, and the original context may be lost by then.

src/lib/homestore_backend/replication_state_machine.cpp
JacksonYao287: ok.   I want to do this because assert can helping us identify the unexpected case after upgrade is done. 

we can change this part in the next release after upgrade is done.

src/lib/homestore_backend/replication_state_machine.cpp
Hooper9973: changed it to tombstone_pba directly.
Although chunk_id seems hardly to be the 0xFFFF, but tombstone_pba is more safe.
And when rollback mesg, homestore will free the blk, and tombstone_pba's blk_count is 0 which is more safe.

src/lib/homestore_backend/hs_shard_manager.cpp
Hooper9973: Good!

src/include/homeobject/shard_manager.hpp
Hooper9973: Done

src/lib/homestore_backend/tests/hs_shard_tests.cpp
```diff
@@ -321,84 +321,87 @@ TEST_F(HomeObjectFixture, ShardVersionMigrationRecovery) {
...
-            old_sb->version = 0x01; // Old version
-            // Simulate the actual bug: first v1 shard has incorrect type field set to BLOB_INFO
+            old_sb->version = HSHomeObject::DataHeader::data_header_version;
+            old_sb->type = HSHomeObject::DataHeader::data_type_t::SHARD_INFO;
+            old_sb->sb_version = 0x02;
+            // Simulate the actual bug: first v2 shard has incorrect type field set to BLOB_INFO
             if (shard_id == shard_ids[0]) {
                 old_sb->type = HSHomeObject::DataHeader::data_type_t::BLOB_INFO; // Bug scenario
-                LOGINFO("Created v1 shard {} with BLOB_INFO type (bug scenario)", shard_id);
+                LOGINFO("Created v2 shard {} with BLOB_INFO type (bug scenario)", shard_id);
```
koujl: The bug is already fixed, let's remove this part.

Hooper9973: done

JacksonYao287 (approved):
LGTM

PR #440, opened 2026-06-24
Remove DEBUG_ASSERT for gc_thresh_low <= gc_thresh_high

xiaoxichen:
As we have auto adjust, it is unnecessary to have the debug assert


JacksonYao287 (approved):
LGTM

PR #441, opened 2026-06-24
Fix garbage_ratio_pct truncation for chunks with very few defrag blocks

xiaoxichen:
Use float instead of uint32_t for garbage_ratio_pct so that chunks with a small defrag_blks count (e.g. 1 block out of 200) are not silently rounded to 0 and excluded from GC even when gc_thresh is set to 0.


JacksonYao287 (approved):
LGTM

PR #444, opened 2026-06-25
improve raft test framework

JacksonYao287:
occasionally，we can see CI is stuck at homestore_test_pg/shard/blob。 the root cause is unexpected leader switch.
follower will wait for something to happen, but leader think it is not leader any more( because of leader switch) and do not schedule some op, then all the member will sync and wait at some point, and thus the UT is stuck.

This PR try to add more retry and avoid this case

### DIFF
--- a/conanfile.py
+++ b/conanfile.py
@@ -10,7 +10,7 @@ required_conan_version = ">=1.60.0"
 
 class HomeObjectConan(ConanFile):
     name = "homeobject"
-    version = "4.1.22"
+    version = "4.1.23"
 
     homepage = "https://github.com/eBay/HomeObject"
     description = "Blob Store built on HomeStore"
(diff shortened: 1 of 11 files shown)


xiaoxichen:
TL, DR:
any possibility to use the ho api to reconcile leadership?

JacksonYao287:
it`s not related with reconciling leadership. what I want in this PR is that the op(for example, put_blob) is eventually be executed by a replica and will not be missed even if unexpected leader switch happens.

From another perspective, if three replicas all think they are not leader and stuck at waiting for some op(for example, blob_exist), who should schedule reconciling leadership and how to reconcile leadership(replicas are all stuck now).

xiaoxichen (dismissed):

xiaoxichen (approved):

PR #446, opened 2026-06-30
SDSTOR-22799 clean up: remove GC shard header/footer I/O, blob commit validation

Hooper9973:
- GC no longer writes physical shard header/footer blocks

- Blob commit validates chunk state matches shard metadata

### DIFF
--- a/src/lib/homestore_backend/CMakeLists.txt
+++ b/src/lib/homestore_backend/CMakeLists.txt
@@ -177,3 +177,17 @@ add_test(NAME HomestoreTestGC_ShardRaceTests
         --override_config hs_backend_config.gc_garbage_rate_threshold=0
         --override_config hs_backend_config.gc_scan_interval_sec=5
         --gtest_filter=HomeObjectFixture.StalePChunkRouteAfterGC:HomeObjectFixture.StaleBlobRouteAfterSealAndGC)
+
+# All tests share disk, port, and shared-memory resources, so run them serially to avoid parallel conflicts.
+set_tests_properties(
+    HomestoreTestBlob
+    HomestoreTestMisc
+    HomestoreTestReplaceMember
+    HomestoreTestReplaceMemberWithBaselineResync
+    HomestoreTestBaselineResync
+    HomestoreResyncTestWithFollowerRestart
+    HomestoreResyncTestWithLeaderRestart
+    FetchDataWithOriginatorGC
+    HomestoreTestGC
+    HomestoreTestGC_ShardRaceTests
+    PROPERTIES RUN_SERIAL TRUE)
--- a/src/lib/homestore_backend/gc_manager.cpp
+++ b/src/lib/homestore_backend/gc_manager.cpp
@@ -650,27 +650,6 @@ bool GCManager::pdev_gc_actor::replace_blob_index(
     return true;
 }
 
-sisl::sg_list GCManager::pdev_gc_actor::generate_shard_super_blk_sg_list(shard_id_t shard_id) {
-    // TODO: do the buffer check before using it.
-    auto raw_shard_sb = m_hs_home_object->_get_hs_shard(shard_id);
-    RELEASE_ASSERT(raw_shard_sb, "can not find shard super blk for shard_id=0x{:x} !!!", shard_id);
-
-    const auto shard_sb =
-        const_cast< HSHomeObject::HS_Shard* >(d_cast< const HSHomeObject::HS_Shard* >(raw_shard_sb))->sb_.get();
-
-    auto blk_size = homestore::data_service().get_blk_size();
-    auto shard_sb_size = sizeof(HSHomeObject::shard_info_superblk);
-    auto total_size = sisl::round_up(shard_sb_size, blk_size);
-    auto shard_sb_buf = iomanager.iobuf_alloc(blk_size, total_size);
-
-    std::memcpy(shard_sb_buf, shard_sb, shard_sb_size);
-
-    sisl::sg_list shard_sb_sgs;
-    shard_sb_sgs.size = total_size;
-    shard_sb_sgs.iovs.emplace_back(iovec{.iov_base = shard_sb_buf, .iov_len = total_size});
-    return shard_sb_sgs;
-}
-
 // note that, when we copy data, there is not create shard or put blob in this chunk, only delete blob might happen.
 bool GCManager::pdev_gc_actor::copy_valid_data(
     chunk_id_t move_from_chunk, chunk_id_t move_to_chunk,
@@ -723,13 +702,11 @@ bool GCManager::pdev_gc_actor::copy_valid_data(
 
     homestore::blk_alloc_hints hints;
     hints.chunk_id_hint = move_to_chunk;
-    homestore::MultiBlkId out_blkids;
 
     auto pg_index_table = m_hs_home_object->get_hs_pg(pg_id)->index_table_;
     const auto blk_size = data_service.get_blk_size();
 
     for (const auto& shard_id : shards) {
-        bool is_last_shard = (shard_id == last_shard_id);
         std::vector< std::pair< BlobRouteKey, BlobRouteValue > > valid_blob_indexes;
         auto start_key = BlobRouteKey{BlobRoute{shard_id, std::numeric_limits< uint64_t >::min()}};
         auto end_key = BlobRouteKey{BlobRoute{shard_id, std::numeric_limits< uint64_t >::max()}};
@@ -786,12 +763,8 @@ bool GCManager::pdev_gc_actor::copy_valid_data(
         }
 
         if (valid_blob_indexes.empty()) {
-            GCLOGD(task_id, pg_id, shard_id, "empty shard found in move_from_chunk={}", move_from_chunk);
-            // TODO::send a delete shard request to raft channel. there is a case that when we are doing gc, the
-            // shard becomes empty, need to handle this case
-
-            // we should always write a shard header for the last shard if the state of it is open.
-            if (last_shard_state != ShardInfo::State::OPEN || !is_last_shard) continue;
+            GCLOGD(task_id, pg_id, shard_id, "empty shard found in move_from_chunk={}, skip", move_from_chunk);
+            continue;
         } else {
             GCLOGD(task_id, pg_id, shard_id, "{} valid blobs found in move_from_chunk={}", valid_blob_indexes.size(),
                    move_from_chunk);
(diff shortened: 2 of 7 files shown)


src/lib/homestore_backend/gc_manager.hpp
```diff
@@ -292,9 +292,6 @@ class GCManager {
         pdev_gc_metrics& metrics() { return metrics_; }
 
     private:
```
JacksonYao287: remove this private

src/lib/homestore_backend/heap_chunk_selector.cpp
```diff
@@ -347,10 +348,27 @@ void HeapChunkSelector::update_vchunk_info_after_gc(const chunk_num_t move_from_
...
+        }
+    } else {
+        // Normal GC: move_to becomes AVAILABLE for future shards in this PG and must have no owner.
+        move_to_vchunk->m_owner_shard_id = std::nullopt;
+    }
 
     // 2 update the chunk state of move_from_chunk, now it is a reserved chunk
     move_from_vchunk->m_pg_id = std::nullopt;
     move_from_vchunk->m_v_chunk_id = std::nullopt;
+    move_from_vchunk->m_owner_shard_id = std::nullopt;
```
JacksonYao287: this line will revert the change in line 361

src/lib/homestore_backend/hs_shard_manager.cpp
```diff
@@ -662,21 +662,23 @@ void HSHomeObject::on_shard_meta_blk_found(homestore::meta_blk* mblk, sisl::byte
...
+        inuse_chunks.reserve(pair.second->shards_.size());
         for (auto& shard : pair.second->shards_) {
             if (shard->info.state == ShardInfo::State::OPEN) {
-                excluding_chunks.emplace(d_cast< HS_Shard* >(shard.get())->sb_->v_chunk_id);
+                inuse_chunks.emplace(d_cast< HS_Shard* >(shard.get())->sb_->v_chunk_id, shard->info.id);
             }
         }
-        bool res = chunk_selector_->recover_pg_chunks_states(pair.first, excluding_chunks);
+        // After restart, each chunk retains at most one shard log, because every
+        // seal-shard commit forces a cp flush.
```
JacksonYao287: no, we only flush dc_lsn, not trigger a cp flush as of now

src/lib/homestore_backend/hs_blob_manager.cpp
```diff
@@ -229,6 +229,53 @@ bool HSHomeObject::local_add_blob_info(pg_id_t const pg_id, BlobInfo const& blob
...
+                  "sb_pchunk={}",
+                  pg_id, sb_vchunk_id, blob_info.blob_id, ext_chunk->m_state, ext_chunk->m_owner_shard_id.value_or(0),
+                  ext_chunk->get_chunk_id(), sb_pchunk_id);
+            RELEASE_ASSERT(
+                false,
+                "traceID={}, commit-time chunk state invalid: pg={} v_chunk={} blob={} state={} owner=0x{:x} "
+                "pchunk={} sb_pchunk={}",
+                tid, pg_id, sb_vchunk_id, blob_info.blob_id, ext_chunk->m_state,
+                ext_chunk->m_owner_shard_id.value_or(0), ext_chunk->get_chunk_id(), sb_pchunk_id);
+        }
```
JacksonYao287: better consolidate BLOGE and RELEASE_ASSERT into only one RELEASE_ASSERT for all the logs here

src/lib/homestore_backend/heap_chunk_selector.cpp
Hooper9973: this line(371) is to clear move_from_vchunk.
line 361 is to set move_to_vchunk.

src/lib/homestore_backend/CMakeLists.txt
```diff
@@ -177,3 +177,17 @@ add_test(NAME HomestoreTestGC_ShardRaceTests
         --override_config hs_backend_config.gc_garbage_rate_threshold=0
         --override_config hs_backend_config.gc_scan_interval_sec=5
         --gtest_filter=HomeObjectFixture.StalePChunkRouteAfterGC:HomeObjectFixture.StaleBlobRouteAfterSealAndGC)
+
+# All tests share disk, port, and shared-memory resources, so run them serially to avoid parallel conflicts.
```
JacksonYao287, line 181: any issue without this change?

src/lib/homestore_backend/gc_manager.cpp
```diff
@@ -488,6 +488,33 @@ void GCManager::pdev_gc_actor::handle_recovered_gc_task(
...
+                open_shard = last_shard_id;
+                break;
+            }
+        }
+        if (open_shard.has_value()) {
+            move_from_vchunk->m_owner_shard_id = open_shard;
+            LOGDEBUGMOD(gcmgr, "EGC recovery: set move_from_chunk={} owner to shard=0x{:x} for replay",
+                        move_from_chunk, open_shard.value());
+        } else {
+            LOGWARNMOD(gcmgr, "EGC recovery: no open shard found in move_from_chunk={} or move_to_chunk={}",
```
JacksonYao287: should we assert here?  this is an unexpected case

src/lib/homestore_backend/CMakeLists.txt
Hooper9973: image

I met one time on my env, two case runs in parallel.

src/lib/homestore_backend/heap_chunk_selector.cpp
```diff
@@ -487,14 +495,8 @@ bool HeapChunkSelector::recover_pg_chunks_states(pg_id_t pg_id,
...
         pg_chunk_collection->m_total_blks += chunk->get_total_blks();
-        if (excluding_v_chunk_ids.find(v_chunk_id) == excluding_v_chunk_ids.end()) {
-            chunk->m_state = ChunkState::AVAILABLE;
-            ++pg_chunk_collection->available_num_chunks;
-            pg_chunk_collection->available_blk_count += chunk->available_blks();
-
-        } else {
-            chunk->m_state = ChunkState::INUSE;
-        }
+        ++pg_chunk_collection->available_num_chunks;
```
Besroy: Only increase when chunk available?

src/lib/homestore_backend/heap_chunk_selector.h
```diff
@@ -32,12 +32,25 @@ class HeapChunkSelector : public homestore::ChunkSelector {
...
-                VChunk(chunk), m_state(ChunkState::AVAILABLE), m_pg_id(), m_v_chunk_id() {}
+                VChunk(chunk), m_state(ChunkState::AVAILABLE), m_pg_id(), m_v_chunk_id(), m_owner_shard_id() {}
         ~ExtendedVChunk() = default;
         ChunkState m_state;
         std::optional< pg_id_t > m_pg_id;
         std::optional< chunk_num_t > m_v_chunk_id;
-        bool available() const { return m_state == ChunkState::AVAILABLE; }
+        std::optional< shard_id_t > m_owner_shard_id; // runtime only, not persisted
+        bool available() const { return valid() && m_state == ChunkState::AVAILABLE; }
+        bool valid() const {
```
Besroy: Could you please explain the purpose of this function in more detail? I noticed assert valid is used in a lot of places, and I’m a bit concerned that GC or some other mechanism might expose a transient inconsistent state and cause an assertion failure, even though I don’t have evidence for that yet

src/lib/homestore_backend/heap_chunk_selector.h
```diff
@@ -73,9 +86,10 @@ class HeapChunkSelector : public homestore::ChunkSelector {
 
     csharedChunk select_chunk([[maybe_unused]] homestore::blk_count_t nblks, const homestore::blk_alloc_hints& hints);
 
-    // this function will be used by create shard or recovery flow to mark one specific chunk to be busy, caller should
-    // be responsible to use release_chunk() interface to release it when no longer to use the chunk anymore.
-    homestore::cshared< ExtendedVChunk > select_specific_chunk(const pg_id_t pg_id, const chunk_num_t v_chunk_id);
+    // State machine: GC→wait+retry, AVAILABLE→INUSE(owner=shard), INUSE+same_owner→success, INUSE+other_owner→fail.
+    // Only leader will call this function twice during shard creation, follower will only call once.
```
Besroy: leader call once? (pick_most_available_blk_chunk + select_specifi_chunk)

src/lib/homestore_backend/heap_chunk_selector.cpp
```diff
@@ -487,14 +495,8 @@ bool HeapChunkSelector::recover_pg_chunks_states(pg_id_t pg_id,
     for (size_t v_chunk_id = 0; v_chunk_id < pg_chunks.size(); ++v_chunk_id) {
         auto chunk = pg_chunks[v_chunk_id];
         pg_chunk_collection->m_total_blks += chunk->get_total_blks();
-        if (excluding_v_chunk_ids.find(v_chunk_id) == excluding_v_chunk_ids.end()) {
```
Besroy, line 490: Not sure  if the "DISK_DOWN" state will be affected

src/lib/homestore_backend/hs_blob_manager.cpp
```diff
@@ -229,6 +229,54 @@ bool HSHomeObject::local_add_blob_info(pg_id_t const pg_id, BlobInfo const& blob
         return false;
     }
     if (!exist_already) {
+        // Validate shard vchunk/pchunk mapping for this blob and verify chunk state at commit time.
+        // During log replay, GC triggers a checkpoint flush, so logs after cp_lsn should not reference GC-moved blobs.
```
Besroy: What corner case does GC refer to here? Seems a bit complex. Could we move the check to the allocation stage instead? That way, we might avoid the more complicated validation logic later.

src/lib/homestore_backend/hs_blob_manager.cpp
```diff
@@ -229,6 +229,54 @@ bool HSHomeObject::local_add_blob_info(pg_id_t const pg_id, BlobInfo const& blob
...
+        // the chunk is in INUSE state
+        const auto ext_chunk = chunk_selector()->get_pg_vchunk(pg_id, sb_vchunk_id);
+        if (ext_chunk == nullptr) {
+            BLOGE(tid, blob_info.shard_id, blob_info.blob_id,
+                  "chunk not found at commit time: pg={} v_chunk={} shard=0x{:x} blob={}", pg_id, sb_vchunk_id,
+                  blob_info.shard_id, blob_info.blob_id);
+            RELEASE_ASSERT(false, "traceID={}, chunk not found at commit time: pg={} v_chunk={} shard=0x{:x} blob={}",
+                           tid, pg_id, sb_vchunk_id, blob_info.shard_id, blob_info.blob_id);
+        } else if (!ext_chunk->valid() ||
+                   (ext_chunk->m_state == ChunkState::INUSE && ext_chunk->m_owner_shard_id != blob_info.shard_id) ||
```
Besroy: IIUC, after this PR the shard state is either loaded when the shard SB is found, or updated by that shard’s log replay. Also, only the chunk for the open shard should be inuse (on_log_replay_done selects the specific chunk). But could blob commit hit a case where the blob’s shard has already been sealed, and the same chunk has since been taken over by a new shard, causing the shard-id check here to fail?

src/lib/homestore_backend/heap_chunk_selector.h
Hooper9973: you r right, I will change the comment. The purpose is to explain leader will want to use the specific chunk twice : `pick_most_available_blk_chunk` + `select_specific_chunk`

src/lib/homestore_backend/heap_chunk_selector.h
Hooper9973: The purpose is to keep the chunk consistency. Which can help identify the issue earlier.
GC should keep the consistency.

src/lib/homestore_backend/hs_blob_manager.cpp
Hooper9973: Here checks three things:
1. shard pchunk, vchunk consistency
2. blob pba should match pchunk
3. chunk state check.

If moves to `blob_put_get_blk_alloc_hints`, here are two concerns :
1. we cannot check the 2 item, which has shown up in gc issue.
2. It's hard to define what to do if the blob is trying to alloc on a sealed shard( seal shard+put blob racing). Because a sealed shard may in GC state, which is a intermediate state. Although we can reject the request, set error to the context, it will trigger a emergency on the specific chunk. And the concern 1 cannot be resolved.


The comment means :
Not this pr involved. When I implement the PR, I found the line223:`add_to_index_table` , this function's behavior is:
if put a exist key, the value(pba) should be same, otherwise it will not return `btree_status_t::success`. 
So I'm confused, what will happen, if we replay a sealed shard 's put blob log (which chunk has finished gc), the blob's currenty pbs value is different with the pbs in logs.
But GC will triggered CP flush, so this scenario won't show up.

src/lib/homestore_backend/hs_blob_manager.cpp
Hooper9973: The major concern is to adapt log replay. 

Currently all chunks are initalized after recovering meta(pg, shard)
And will select all open shard `on log replay done`
Here are all the logs between cp_lsn & dc_lsn:

LSN 11: create shard 1 on chunk x
LSN 12: put blob b1_1
LSN 13: seal shard1 on chunk y
LSN 14: create shard2 on chunk x
LSN 15: put blob b2_1

Current on commit logic is:
- create shard, won't select a existing shard.
- seal shard, seal all shard.

The recovery will like, the chunk's always be available with no owner_shard_id,  before `on_log_replay_done`.
So I only check the owner shard when the chunk is in use.

src/lib/homestore_backend/heap_chunk_selector.cpp
Hooper9973: https://github.com/eBay/HomeObject/pull/446#discussion_r3511249369

The implement is to let all chunks to be available (initialized state), and let log replay and on log replay done to recovery the final state.

src/lib/homestore_backend/heap_chunk_selector.h
xiaoxichen: i think the concern is valid,  the state and m_owner_shard_id update is not transactional.  How you deal with the situation the chunkState has been changed but m_owner_shard_id hasnt? 

for example https://github.com/eBay/HomeObject/pull/446/changes#diff-8353a9872ea900abf04e7b517e286ae1d7007ad19bb04bef3b689d4153ac8026R180-R181
if there is an `is_valid` call between L180/181, what should we do

src/lib/homestore_backend/hs_blob_manager.cpp
xiaoxichen: I am not convinced of putting such complicated check on each blob commit... 
it just as wired as you have a blood check at the end of each breath

xiaoxichen (commented):
can we target the ownership tracking as best-effort? 
for example if the ownership tracking is established in-memory upon `on_log_replay_done`,  will it simplify some of the corner cases?

The ownership tracking is something that helps us to reduce complexity and detecting issues. if we need a lot of fancy stuff to make the tracking right,  even potentially more bugs and fixes in the tracking code, it doesnt make sense

Hooper9973:
https://github.com/eBay/HomeObject/pull/446#pullrequestreview-4619535261

> can we target the ownership tracking as best-effort?
> for example if the ownership tracking is established in-memory upon on_log_replay_done, will it simplify some of the corner cases?
> 
> The ownership tracking is something that helps us to reduce complexity and detecting issues. if we need a lot of fancy stuff to make the tracking right, even potentially more bugs and fixes in the tracking code, it doesnt make sense

https://github.com/eBay/HomeObject/pull/446#discussion_r3514730656

> i think the concern is valid, the state and m_owner_shard_id update is not transactional. How you deal with the situation the chunkState has been changed but m_owner_shard_id hasnt?
> 
> for example https://github.com/eBay/HomeObject/pull/446/changes#diff-8353a9872ea900abf04e7b517e286ae1d7007ad19bb04bef3b689d4153ac8026R180-R181
> if there is an is_valid call between L180/181, what should we do
> 



@Besroy @xiaoxichen  @JacksonYao287 I'd like to answer the two question together. In my opinion, it's a trade-off, for ownership tracking there are two options:
1. **best-effort**, only treat ownership as debug info, **doesn't check the ownership** when select chunk or release chunk, which can support more infos for RCA issues.
2. **Hard validation**, this should make the chunk always keep consistency, which can block issues earlier. But this will carry some effort. 
	1. How to make the chunk changes transactional? We can make items private (m_state, m_owner_shard_id), HeapChunkSelector funcs should keep the consistency, ensure consistency before and after the call. 
	2. To be honest, it does bring some mental burden: 
		1. GC: recovery GC is a different way from the normal path,  which will break runtime assumption and transactional: https://github.com/eBay/HomeObject/pull/446/changes/102268efdd779b5d6fe264eadce6cba7cd8a3e09#diff-2867727478be7bf0adc37898831711b64337151e2ebe9799d30258f2fe6c0eb8R498
		2. replay log after restart. https://github.com/eBay/HomeObject/pull/446#discussion_r3511249369 
		
		
Do you have any suggestions? Or option 3?

src/lib/homestore_backend/gc_manager.hpp
Hooper9973: removed

src/lib/homestore_backend/gc_manager.cpp
Hooper9973: done

src/lib/homestore_backend/heap_chunk_selector.h
Hooper9973: remove valid()

src/lib/homestore_backend/hs_blob_manager.cpp
Hooper9973: only check the consistency between blob pba and shard's pchunkid

src/lib/homestore_backend/hs_blob_manager.cpp
Hooper9973: removed

src/lib/homestore_backend/hs_blob_manager.cpp
Hooper9973: done

src/lib/homestore_backend/hs_blob_manager.cpp
Besroy: I cannot get why cannot check the 2 item, could you explain more? --- BTW checking at alloc also works for 'replay a sealed shard 's put blob log '

src/lib/homestore_backend/hs_blob_manager.cpp
Besroy: Yes I understand this case, but will this case happen and failed check: the blob’s shard has already been sealed, and the same chunk has since been taken over by a new shard, then ext_chunk->m_owner_shard_id != blob_info.shard_id

Hooper9973: That case will be intercepted before being inserted into the index table by comparing with sealed_lsn :
https://github.com/Hooper9973/HomeObject/blob/ee5081f4965743898d6cddf88a14f8ada87d17d3/src/lib/homestore_backend/hs_blob_manager.cpp#L306

src/lib/homestore_backend/hs_blob_manager.cpp
Hooper9973: Here are two time spot:

1. get_blob_alloc_hint
2. on_commit

`get_blob_alloc_hint` will be racing with two parallel operation : 1. seal shard on commit 2.GC
Whenever you check, you can only get the temporary state.  
For an example, the **GC issue2**: put blob to a sealed shard.

T1: alloc pba, if check at this time, all things looks good.
T2: seal shard
T3: GC, since the blob is not committed, the data will be lost.

Although this case will be fixed, we can know only the on commit is the final state, we check the consistency after inserting the blob index, if the shard chunk id match the blob's pba, the blob is good.

src/lib/homestore_backend/gc_manager.cpp
```diff
@@ -488,6 +488,31 @@ void GCManager::pdev_gc_actor::handle_recovered_gc_task(
...
+            const auto* shard = m_hs_home_object->_get_hs_shard(last_shard_id);
+            if (shard && shard->info.state == ShardInfo::State::OPEN) {
+                open_shard = last_shard_id;
+                break;
+            }
+        }
+        RELEASE_ASSERT(open_shard.has_value(),
+                       "EGC recovery: no open shard found in move_from_chunk={} or move_to_chunk={}", move_from_chunk,
+                       move_to_chunk);
+        move_from_vchunk->m_owner_shard_id = open_shard;
```
JacksonYao287: move_from_vchunk->m_owner_shard_id = open_shard.value()?

src/lib/homestore_backend/CMakeLists.txt
```diff
@@ -177,3 +177,17 @@ add_test(NAME HomestoreTestGC_ShardRaceTests
         --override_config hs_backend_config.gc_garbage_rate_threshold=0
         --override_config hs_backend_config.gc_scan_interval_sec=5
         --gtest_filter=HomeObjectFixture.StalePChunkRouteAfterGC:HomeObjectFixture.StaleBlobRouteAfterSealAndGC)
```
JacksonYao287, line 179: suggest combine these two test into only one HomestoreTestGC

src/lib/homestore_backend/heap_chunk_selector.cpp
```diff
@@ -125,36 +101,34 @@ HeapChunkSelector::select_specific_chunk(const pg_id_t pg_id, const chunk_num_t
...
+                --pg_chunk_collection->available_num_chunks;
+                pg_chunk_collection->available_blk_count -= chunk->available_blks();
                 break;
+            } else {
+                // INUSE: same owner → idempotent; different owner → error.
+                if (chunk->m_owner_shard_id == owner_shard_id) { break; }
+                LOGWARNMOD(homeobject,
+                           "v_chunk_id={} for pg={} is INUSE by shard=0x{:x}, cannot select for shard=0x{:x}",
+                           v_chunk_id, pg_id, chunk->m_owner_shard_id.value_or(0), owner_shard_id);
+                return nullptr;
```
JacksonYao287: I am not sure it is safe to return a nullptr, should we continue waiting in the loop until the chunk to be available

src/lib/homestore_backend/snapshot_receive_handler.cpp
```diff
@@ -301,7 +301,7 @@ int HSHomeObject::SnapshotReceiveHandler::process_blobs_snapshot_data(ResyncBlob
             state = (*iter->second)->info.state;
         }
         if (state == ShardInfo::State::SEALED) {
-            home_obj_.chunk_selector()->release_chunk(ctx_->pg_id, v_chunk_id.value());
+            home_obj_.chunk_selector()->release_chunk(ctx_->pg_id, v_chunk_id.value(), ctx_->shard_cursor);
```
JacksonYao287: add an assert to check if the return value of release_chunk is true

src/lib/homestore_backend/hs_blob_manager.cpp
```diff
@@ -229,6 +229,22 @@ bool HSHomeObject::local_add_blob_info(pg_id_t const pg_id, BlobInfo const& blob
         return false;
     }
     if (!exist_already) {
+        // Validate the blob was written to the pchunk the shard records.
+        // Catches GC-related inconsistencies (EGC moved chunk while blob was in-flight).
```
JacksonYao287: egc will start only when all the appended log are committed, so I think if egc is working, there is no in-flight put_blob

src/lib/homestore_backend/heap_chunk_selector.cpp
```diff
@@ -180,8 +154,17 @@ bool HeapChunkSelector::release_chunk(const pg_id_t pg_id, const chunk_num_t v_c
     }
     std::scoped_lock lock(pg_chunk_collection->mtx);
     auto chunk = pg_chunks[v_chunk_id];
+    if (chunk->m_state == ChunkState::AVAILABLE) {
```
JacksonYao287: suggest add release_assert(chunk->m_state != ChunkState::GC)

xiaoxichen:
@Hooper9973  based on the review discussion,   it seems like we should first merge those cleanup codes into 0.5.6
and leave the chunk ownership tracking into 0.5.7 , what do you think?

cc @JacksonYao287 @Besroy

src/lib/homestore_backend/hs_blob_manager.cpp
Besroy: Will this happen in replay senario:
T1: put blob on shard A with LSN 100
T2: seal shard A with LSN 101, shard sb: sealed_lsn=101
T3: create shard B with LSN 102, chunk owner = B
T4: crash
After crash, assume that lsn100 should replay and the blob idx not persited, then assert at chunk owner_shard_id  check

JacksonYao287:
> leave the chunk ownership tracking into 0.5.7

if the committed two prs are sufficient to solve the create shard/ GC issues, I am +1 with this proposal

Besroy:
> @Hooper9973 based on the review discussion, it seems like we should first merge those cleanup codes into 0.5.6 and leave the chunk ownership tracking into 0.5.7 , what do you think?
> 
> cc @JacksonYao287 @Besroy

lgtm, we can merge thoese cleanup codes into 0.5.6, build a new image and test on SH as soon as possible

Hooper9973:
> @Hooper9973 based on the review discussion, it seems like we should first merge those cleanup codes into 0.5.6
> and leave the chunk ownership tracking into 0.5.7 , what do you think?

Sounds good, I agree with that split. I’ll prepare a new version

src/lib/homestore_backend/CMakeLists.txt
Hooper9973: ShardRaceTests need additional parameter `--chunks_per_pg 1`, that's the reason I split them

src/lib/homestore_backend/gc_manager.cpp
```diff
@@ -812,220 +785,120 @@ bool GCManager::pdev_gc_actor::copy_valid_data(
...
                                     GCLOGE(task_id, pg_id, shard_id,
-                                           "Failed to copy blob for move_to_chunk={}, will cancel this task",
-                                           move_to_chunk);
-                                    // if any op fails, we drop this gc task.
-                                    return folly::makeFuture< std::error_code >(
-                                        std::make_error_code(std::errc::operation_canceled));
+                                           "Failed to insert new key to gc index table for "
+                                           "move_to_chunk={}, blob_id={}, err={}",
+                                           move_to_chunk, blob_id, status);
+                                    return false;
```
JacksonYao287, line 870: NIT：we can have an improvment here. 

not return false or true for blob copy,  return the error directly, so that we can see why blob copy failed.

JacksonYao287 (approved):
LGTM. only a minor suggestion

Besroy (approved):

xiaoxichen (approved):
lgtm

PR #451, opened 2026-07-09
SDSTOR-23112: fix snapshot shard state when sealed_lsn exceeds snapshot LSN cutoff

JacksonYao287:
## Problem

During snapshot generation, PGBlobIterator builds a shard list from the leader's in-memory PG state filtered by `upto_lsn` (the snapshot LSN cutoff). Before this fix, a shard whose `create_lsn <= upto_lsn` was always included with its current state — even if its `sealed_lsn` was beyond `upto_lsn` (i.e., the seal happened *after* the snapshot point).

When a follower received such a snapshot it saw the shard as SEALED and released the backing chunk. Log replay then tries to replay the `seal_shard` entry for that LSN, but the chunk has already been reclaimed or relocated, causing the commit to fail with stale physical-chunk references.

## Root Cause

The race window is:

1. Leader creates shard S at LSN L1.
2. Leader creates shard S2 at LSN L2 (L2 > L1).
3. Leader seals shard S at LSN L3 (L3 > L2).
4. Snapshot is taken with `upto_lsn = L2`.

In step 4, S has `create_lsn = L1 <= L2` so it passes the LSN filter, but `sealed_lsn = L3 > L2`. Old code included S with state SEALED. The follower then applied the snapshot, treated S as SEALED, released its chunk, and later failed when replaying the seal_shard log at L3.

## Fix

In `PGBlobIterator::PGBlobIterator`, after the `create_lsn` filter, add a second check:

    if (shard_info.state == SEALED && shard_info.sealed_lsn > upto_lsn) {
        shard_info.state = OPEN;
        shard_info.sealed_lsn = INT64_MAX;
    }

This converts the shard's snapshot state back to OPEN so the follower does not release the chunk prematurely. The follower will replay the `seal_shard` log entry when log replay catches up, at which point it correctly transitions the shard to SEALED with the right `sealed_lsn`.

Other changes in this commit:
- Remove dead commented-out LOGDEBUG in heap_chunk_selector.cpp.
- Bump version to 4.2.5 in conanfile.py.

## Tests

- **PGBlobIteratorSealedLsnCutoff (new)**: directly exercises both branches of the fix:
  - Case A — `snp_lsn` between `create_lsn` and `sealed_lsn`: asserts the shard appears as OPEN with `sealed_lsn == INT64_MAX`.
  - Case B — `snp_lsn >= sealed_lsn`: asserts the shard remains SEALED with its original `sealed_lsn`.
- **PGBlobIterator (updated)**: the existing test seals a shard after the snapshot LSN, so its state/sealed_lsn assertions are updated to expect the downgraded OPEN state produced by the fix.

### DIFF
--- a/src/lib/homestore_backend/pg_blob_iterator.cpp
+++ b/src/lib/homestore_backend/pg_blob_iterator.cpp
@@ -24,14 +24,37 @@ HSHomeObject::PGBlobIterator::PGBlobIterator(HSHomeObject& home_obj, homestore::
     if (upto_lsn != 0) {
         // Iterate all shards and its blobs which have lsn <= upto_lsn
         for (auto& shard : pg->shards_) {
-            if (shard->info.create_lsn <= upto_lsn) {
-                auto v_chunk_id = home_obj_.get_shard_v_chunk_id(shard->info.id);
-                shard_list_.emplace_back(shard->info, v_chunk_id.value());
+            auto shard_info = shard->info;
+            if (shard_info.create_lsn <= upto_lsn) {
+                auto v_chunk_id = home_obj_.get_shard_v_chunk_id(shard_info.id);
+                RELEASE_ASSERT(v_chunk_id.has_value(), "v_chunk_id not found for shard_id={}", shard_info.id);
+
+                /*
+                 * Snapshot metadata may reflect the leader's latest shard state, even when that
+                 * state transition happened after the snapshot LSN cutoff. If such a shard is
+                 * treated as sealed during snapshot apply, the receiver may release the backing
+                 * chunk too early. That chunk can then be reclaimed or relocated before log replay
+                 * catches up, leaving later writes to use stale physical-chunk information and
+                 * eventually fail during commit.
+                 *
+                 * To keep snapshot state consistent with the requested cutoff, convert any shard
+                 * whose sealed_lsn is beyond upto_lsn back to OPEN for iteration purposes and follower will received
+                 * and commit the seal_shard log for this shard after the snapshot apply is completed, where the shard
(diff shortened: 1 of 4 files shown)


src/lib/homestore_backend/pg_blob_iterator.cpp
```diff
@@ -24,14 +24,37 @@ HSHomeObject::PGBlobIterator::PGBlobIterator(HSHomeObject& home_obj, homestore::
...
+                 * eventually fail during commit.
+                 *
+                 * To keep snapshot state consistent with the requested cutoff, convert any shard
+                 * whose sealed_lsn is beyond upto_lsn back to OPEN for iteration purposes and follower will received
+                 * and commit the seal_shard log for this shard after the snapshot apply is completed, where the shard
+                 * state and sealed_lsn will be updated to the correct value.
+                 */
+
+                if (shard_info.state == ShardInfo::State::SEALED && shard_info.sealed_lsn > upto_lsn) {
+                    shard_info.state = ShardInfo::State::OPEN;
```
yuwmao, line 47: Do we need to make a copy of shard?

JacksonYao287: see line 27 

```
auto shard_info = shard->info;
```

shard_info is already a copy of shard->info

koujl (approved):
lgtm

src/lib/homestore_backend/pg_blob_iterator.cpp
yuwmao: OK, thanks!

Hooper9973 (approved):
LGTM

yuwmao (approved):
lgtm

PR #452, opened 2026-07-14
Fix data corruption on CP persistence inconsistency

koujl:
https://jirap.corp.ebay.com/browse/SDSTOR-23023

During BR the IndexSvc journal and the AppendBlkAllocator CP superblock are flushed in parallel with no atomicity guarantee. If the process crashes after the index entry is durable but before the allocator superblock is written, the receiver resumes with the blk_id visible in the index but the allocator watermarks reset. The dedup skip path then bypasses commit_blk entirely, leaving the blocks silently available for reallocation by subsequent blob writes and corrupting the data on disk.

The fix re-commits the blk_id recovered from the index entry so the allocator watermarks are restored before new allocations proceed.

Related PR: https://github.com/eBay/HomeStore/pull/896

---

Also fixes a bug from #435 (https://jirap.corp.ebay.com/browse/SDSTOR-23162)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@
 .cache
 .clangd
 .claude/*
+.opencode/*
 
 build
 test_package/build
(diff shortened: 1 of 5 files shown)


JacksonYao287 (approved):
LGTM

xiaoxichen (approved):
lgtm

PR #454, opened 2026-07-29
Fix batch resend bug caused by prefetch limit

koujl:
Retained look-ahead data could exhaust the donor prefetch budget while rebuilding a resent batch, preventing a required earlier blob from being loaded and causing each retry to fail at the same cursor, sticking the resync progress at the same point.

Allow blobs preceding the retained prefetch frontier to bypass the bounded look-ahead budget, and cover the scenario with a focused regression test.

Context: https://jirap.corp.ebay.com/browse/SDSTOR-24121

### DIFF
--- a/conanfile.py
+++ b/conanfile.py
@@ -10,7 +10,7 @@ required_conan_version = ">=1.60.0"
 
 class HomeObjectConan(ConanFile):
     name = "homeobject"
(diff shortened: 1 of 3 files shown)


JacksonYao287 (approved):

Besroy (approved):
LGTM

PR #455, opened 2026-07-31
Improve baseline resync logging

koujl:
Eliminate inconsistent terminology, context, and severity levels across existing BR read/write logs. Add structured lifecycle, payload, retry, and failure records with consistent PG, shard, batch, blob, and block identifiers. Keep key object and payload transitions at INFO while moving per-blob details to TRACE to reduce routine log noise.

### DIFF
--- a/src/lib/homestore_backend/snapshot_receive_handler.cpp
+++ b/src/lib/homestore_backend/snapshot_receive_handler.cpp
@@ -13,8 +13,6 @@ HSHomeObject::SnapshotReceiveHandler::SnapshotReceiveHandler(HSHomeObject& home_
         home_obj_(home_obj), repl_dev_(std::move(repl_dev)), cp_fut(folly::makeFuture< bool >(true)) {}
 
 int HSHomeObject::SnapshotReceiveHandler::process_pg_snapshot_data(ResyncPGMetaData const& pg_meta) {
-    LOGI("process_pg_snapshot_data pg={}", pg_meta.pg_id());
-
     // Init shard list
     ctx_->shard_list.clear();
     const auto ids = pg_meta.shard_ids();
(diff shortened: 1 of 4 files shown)


JacksonYao287:
only log changes in this PR, no logic changes, right?

src/lib/homestore_backend/snapshot_receive_handler.cpp
```diff
@@ -271,29 +277,34 @@ int HSHomeObject::SnapshotReceiveHandler::process_blobs_snapshot_data(ResyncBlob
...
                     // Add local blob info to index & PG
                     bool success =
                         home_obj_.local_add_blob_info(ctx_->pg_id, BlobInfo{ctx_->shard_cursor, blob_id, blk_id});
                     if (!success) {
-                        LOGE("Failed to add blob info for blob_id={}", blob_id);
+                        LOGE("Failed to index resync blob: pg={}, shard_id=0x{:x}, blob={}, blkid={}", ctx_->pg_id,
+                             ctx_->shard_cursor, blob_id, blk_id.to_string());
                         homestore::data_service().async_free_blk(blk_id).get();
-                        return err;
+                        return std::make_error_code(std::errc::io_error);
```
koujl, line 301: @JacksonYao287 All the changes are solely log optimizations, except a minor bugfix here detected during the log optimization, regarding the async return value.

JacksonYao287 (approved):

Besroy:
LGTM. After this PR is merged, what log mode do you expect? IIRC production `homeobject` currently uses `info`, while SH tests use `debug`.

koujl:
> LGTM. After this PR is merged, what log mode do you expect? IIRC production `homeobject` currently uses `info`, while SH tests use `debug`.

Let's keep the production log level at 'info' as the current SM sherlock collection is already overwhelmed. For SH we can elevate the level to `trace` manually only when we're reproducing bugs involving blob-level behaviors.

Besroy:
> > LGTM. After this PR is merged, what log mode do you expect? IIRC production `homeobject` currently uses `info`, while SH tests use `debug`.
> 
> Let's keep the production log level at 'info' as the current SM sherlock collection is already overwhelmed. For SH we can elevate the level to `trace` manually only when we're reproducing bugs involving blob-level behaviors.

OK, SH will stay in debug mode unless issue reproduction is needed.

PR #462, opened 2026-08-28
gc: expose per-pdev backlog and pressure metrics

xiaoxichen:
Add five per-pdev gauges to pdev_gc_metrics for operator visibility into the GC backlog:

  - pending_gc_bytes            reclaimable bytes in PG-owned chunks
  - eligible_gc_bytes           bytes GC would pick up under current policy
  - eligible_gc_chunk_count     count for the above
  - pending_normal_gc_task_count normal-priority queue depth
  - pending_gc_chunks_ratio     10-series labeled gauge over (0,10]% .. (90,100]%

The bucket metric emits as a single Prometheus family with a 'bucket' label, differentiated by direct m_impl_ptr->register_gauge() calls (the compile-time-name macro cannot express label-differentiated series). Descriptions embed the bucket label so the sisl JSON dump does not collapse the 10 entries to one.

Values are refreshed once per gc_scan_interval by scan_chunks_for_gc, using a new private GCManager::get_chunk_gc_snapshot helper that folds the four ExtendedVChunk field reads into a single lookup per chunk; get_chunk_gc_ratio is retained as a thin wrapper for backward compatibility.

### DIFF
--- a/src/lib/homestore_backend/gc_manager.cpp
+++ b/src/lib/homestore_backend/gc_manager.cpp
@@ -223,32 +223,52 @@ std::shared_ptr< GCManager::pdev_gc_actor > GCManager::get_pdev_gc_actor(uint32_
     return it->second;
 }
 
-float GCManager::get_chunk_gc_ratio(chunk_id_t chunk_id) {
+GCManager::ChunkGCSnapshot GCManager::get_chunk_gc_snapshot(chunk_id_t chunk_id, uint8_t gc_thresh_low) {
+    ChunkGCSnapshot snap;
     auto chunk = m_chunk_selector->get_extend_vchunk(chunk_id);
 
-    // Only AVAILABLE chunks are eligible: INUSE means an open shard owns it, GC means already being processed.
-    if (chunk->m_state != ChunkState::AVAILABLE) { return 0.0f; }
+    // Populate raw fields regardless of eligibility. Metrics callers include chunks that are
+    // currently INUSE, being GCed, or owned by a not-yet-alive PG in the "pending" backlog, so
+    // we must not early-return here.
+    snap.defrag_blks = chunk->get_defrag_nblks();
+    snap.total_blks = chunk->get_total_blks();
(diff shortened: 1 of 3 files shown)


src/lib/homestore_backend/gc_manager.cpp
JacksonYao287: when reaching here, it means some gc tasks are backloged. it is valuable for us to report eligible_gc_bytes/eligible_gc_chunk_count/pending_gc_chunks_ratio , etc. to metrics now to show us the gc pressure. 

if we just continue and skip publish_scan_snapshot, we can not get those information timely.

JacksonYao287 (commented):
generally , LGTM. better to wait for @Besroy  and @yuwmao  to also take a look.

src/lib/homestore_backend/gc_manager.cpp
xiaoxichen: makes sense, we should remove this checks to ensure we refreshed the metrics

Besroy:
Generally LGTM.  Just curious why we want to add these metrics --- eligible makes sense to me, but could you share more context on why we track pending cases where there is only a small amount of garbage and the threshold has not been reached yet?

BTW from my experience during triage, metrics like “running GC task count” or “GC task cost time” may be more helpful for identifying stuck GC tasks.

xiaoxichen:
> Generally LGTM. Just curious why we want to add these metrics --- eligible makes sense to me, but could you share more context on why we track pending cases where there is only a small amount of garbage and the threshold has not been reached yet?
> 
> BTW from my experience during triage, metrics like “running GC task count” or “GC task cost time” may be more helpful for identifying stuck GC tasks.

we are seeing a significant amount of capacity drift between "used capacity report by SM-> CM->DM" vs "user consumed capacity reported by usage exporter" so the added metrics is to capture how much space we can reclaim in theory vs how much spac our configuration allow us to reclaim.

Besroy (approved):

